PR #7, opened 2018-05-15
Fix parsing

jdlara-berkeley:
This enables parsing of PSSE files and Matpower using PowerModels parser. Also includes the updates to the test. The code for parsing CSV is not tested.


test/parsestandard.jl
```diff
@@ -0,0 +1,43 @@
...
+        print("Parsing $f ...\n")
+        ParseStandardFiles(f)
+        println("Successfully parsed $f")
+    catch
+        warn("Error while parsing $f")
+        catch_stacktrace()
+    end
+end
+
+cd(string(homedir(),"/.julia/v0.6/PowerSystems/data/psse_raw"))
```
kdheepak: Avoid `cd` to prevent change of directory when not expected.

PR #9, opened 2018-05-18
Provide support v7

jdlara-berkeley:
Added some conditionals to load both versions of the code. The tests are likely to fail because the test data still uses the @NT sintax


kdheepak:
Why do we have to have separate folders for v0.6 and v0.7?

jdlara-berkeley:
I copied the code for v7 and added a conditional on ```PowerSystems.jl``` on which folder to load depending on the version. It seemed a reasonable way to go.

kdheepak:
We will use a separate branch for v0.7 instead.

PR #31, opened 2018-07-03
Feature parsing 

### LINKED ISSUE #13: Enable forecasting parsing and types
Just for the record

### LINKED ISSUE #24: add thermal limit (or max rate) for all power lines.
When building a line or a transformer we need to make sure the rate is defined and below the SIL value. 

I am removing the ```Union{Nothing, Float64}``` in the rate field at branch types. 

see [cd32958](https://github.com/NREL/PowerSystems.jl/commit/cd329580a300d8686e28d63364585d4d086d0c22)

### LINKED ISSUE #29: Remove data and add build.jl
Create a deps file with all the data and remove data folder from the repository.

sourabhdalvi:
Ability parse data from PowerModels to PowerSystems, fixes #13 
Removed data files and added build.jl to download data files from release based on tag,  fixes #29 
Add SIL limit check for lines , fixes #24 
Now only prints PowerModels errors while parsing, no more warnings


jdlara-berkeley:
@sourabhdalvi, some of the testings, is failing, can you check where, please?

deps/build.jl
```diff
@@ -0,0 +1,72 @@
+using InfoZIP
```
jdlara-berkeley: This line shuld be removed and use `InfoZip.X` instead.

src/models/topological_elements.jl
```diff
@@ -19,7 +19,17 @@ Bus(;   number = 0,
         basevoltage = nothing
     ) = Bus(number, name, bustype, angle, voltage, PowerSystems.orderedlimits(voltagelimits, "Voltage"), basevoltage)
 
-struct LoadZones
+struct LoadZones  <: PowerSystemDevice
     number::Int
     name::String
+    buses::Array{PowerSystems.Bus,1}
+    maxrealpower::Float64
```
jdlara-berkeley: Is this a maxrealpower transfer? @claytonpbarrows what is the common use of load zones?

src/models/forecasts.jl
```diff
@@ -3,11 +3,12 @@ abstract type
 end
 
 struct Deterministic <: Forecast
+    device::PowerSystemDevice
```
jdlara-berkeley: @claytonpbarrows  should we add a field that specifies the forecast use? For instance Day Ahead vs Real Time?

src/models/topological_elements.jl
claytonpbarrows: no, I think it's peak load

src/models/forecasts.jl
claytonpbarrows: I don't think so. I think that should be driven by some combination of the existing fields (which aren't completely clear to me). Basically we need to know:
 - What the forecast is for (device)
 - When the forecast is available (initial time?)
 - The horizon of the forecast (e.g. the timestamps for the values in the forecast timeseries). (interval, resolution, and horizon?)

PR #47, opened 2018-07-19
Dev line checks

sourabhdalvi:
Updated Thermal limit checks for lines 
Updated Angle limit checks for lines

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -50,12 +50,8 @@ struct PowerSystem{L <: ElectricLoad,
 
         sources = genclassifier(generators);
         time_length = timeseriescheckload(loads)
-<<<<<<< HEAD
         !isa(sources.renewable, Nothing) ? timeserieschecksources(sources.renewable, time_length) : true
         !isa(sources.hydro, Nothing) ? timeserieschecksources(sources.hydro, time_length): true
-=======
-        timeserieschecksources(sources.renewable, time_length)
->>>>>>> 9a89aa68c2f128d7bd1ab748a2a8b6059fbd3b0f
         calculate_thermal_limits!(branches,basepower)
         check_angle_limits!(branches)
         #timeserieschecksources(sources.hydro, time_length)
--- a/src/base.jl
+++ b/src/base.jl
@@ -52,6 +52,9 @@ struct PowerSystem{L <: ElectricLoad,
         time_length = timeseriescheckload(loads)
         !isa(sources.renewable, Nothing) ? timeserieschecksources(sources.renewable, time_length) : true
         !isa(sources.hydro, Nothing) ? timeserieschecksources(sources.hydro, time_length): true
+        calculate_thermal_limits!(branches,basepower)
+        check_angle_limits!(branches)
+        #timeserieschecksources(sources.hydro, time_length)
 
         new{L, B, Nothing}(buses,
                 sources,
@@ -99,6 +102,8 @@ struct PowerSystem{L <: ElectricLoad,
         slackbuscheck(buses)
         buscheck(buses)
         pvbuscheck(buses, generators)
+        calculate_thermal_limits!(branches,basepower)
+        check_angle_limits!(branches)
 
 
         sources = genclassifier(generators);
--- a/src/models/branches/lines.jl
+++ b/src/models/branches/lines.jl
@@ -9,12 +9,12 @@ struct Line <: Branch
     anglelimits::@NT(max::Float64, min::Float64) #Degrees
 
     function Line(name::String, available::Bool, connectionpoints::@NT(from::Bus, to::Bus), r::Float64, x::Float64, b::@NT(from::Float64, to::Float64), rate, anglelimits)
-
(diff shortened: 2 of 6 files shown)


test/parsestandard.jl
```diff
@@ -15,7 +15,7 @@ for f in files
         println("Successfully parsed $f to PowerSystems dict")
         Buses, Generators, Storage, Branches, Loads, LoadZones ,Shunts = PowerSystems.ps_dict2ps_struct(ps_dict)
         println("Successfully parsed $f to PowerSystems devices")
-        sys_test = PowerSystem(Buses, Generators,Loads,Branches,Storage,230.0,1000.0) # TODO: Add DClines, Shunts , LoadZones
+        sys_test = PowerSystem(Buses, Generators,Loads,Branches,Storage,230.0,ps_dict["baseMVA"]) # TODO: Add DClines, Shunts , LoadZones
```
jd-lara: Read the based voltage from the Dictionary if provided,

src/utils/system_checks.jl
```diff
@@ -66,43 +66,58 @@ function pvbuscheck(buses::Array{Bus}, generators::Array{T}) where {T<:Generator
...
-    if isa(fr_vmax,Nothing) || isa(to_vmax,Nothing)
-        fr_vmax = 1.0
-        to_vmax = 0.9
-        diff_angle = abs(connectionpoints.from.angle -connectionpoints.to.angle)
-        new_rate = y_mag*fr_vmax*to_vmax*cos(theta_max)
-    else
-    m_vmax = max(fr_vmax, to_vmax)
-    c_max = sqrt(fr_vmax^2 + to_vmax^2 - 2*fr_vmax*to_vmax*cos(theta_max))
-    new_rate = y_mag*m_vmax*c_max
+function calculate_thermal_limits!(Branches::Array{<:Branch,1},baseMVA::Float64)
```
jd-lara: follow stylistic standard, `branches` and `basemva`. Capital letter are reserved for struct definitions.

src/utils/system_checks.jl
```diff
@@ -66,43 +66,58 @@ function pvbuscheck(buses::Array{Bus}, generators::Array{T}) where {T<:Generator
...
+                Branches[ix] = Line(deepcopy(l.name),deepcopy(l.available),
+                                    deepcopy(l.connectionpoints),deepcopy(l.r),
+                                    deepcopy(l.x),deepcopy(l.b),deepcopy(l.rate),
+                                    newanglelimits)
+            end
+        end
+    end
 end
 
-function calculate_thermal_limits(r::Float64, x::Float64, rate::@NT(from_to::Float64, to_from::Float64), anglelimits::@NT(max::Float64, min::Float64), connectionpoints::@NT(from::Bus, to::Bus))
```
jd-lara: remove underscores from function names. As per stylistic guides

src/utils/system_checks.jl
```diff
@@ -66,43 +66,58 @@ function pvbuscheck(buses::Array{Bus}, generators::Array{T}) where {T<:Generator
     end
 end
 
-function check_angle_limits(anglelimits::@NT(max::Float64, min::Float64))
```
jd-lara: remove underscore from function names as per stylistic guidelines

src/models/branches/lines.jl
```diff
@@ -9,12 +9,12 @@ struct Line <: Branch
     anglelimits::@NT(max::Float64, min::Float64) #Degrees
 
     function Line(name::String, available::Bool, connectionpoints::@NT(from::Bus, to::Bus), r::Float64, x::Float64, b::@NT(from::Float64, to::Float64), rate, anglelimits)
```
jd-lara: we can remove the inner constructor completely.

src/base.jl
```diff
@@ -52,6 +52,9 @@ struct PowerSystem{L <: ElectricLoad,
         time_length = timeseriescheckload(loads)
         !isa(sources.renewable, Nothing) ? timeserieschecksources(sources.renewable, time_length) : true
         !isa(sources.hydro, Nothing) ? timeserieschecksources(sources.hydro, time_length): true
+        calculate_thermal_limits!(branches,basepower)
```
jd-lara: there is no need to run the thermal limit check twice.

src/base.jl
```diff
@@ -52,6 +52,9 @@ struct PowerSystem{L <: ElectricLoad,
...
+        check_angle_limits!(branches)
```
jd-lara: there is no need to run the angle check twice.

sourabhdalvi: I dont get it, it should be in each constructor that takes a Array of Branches, just like the timeserieschecksources check

src/models/branches/lines.jl
sourabhdalvi: I have left it there because in most cases just one rate is passed and not a NT(from_to, to_from), so it checks if the rate is in correct format.

jd-lara: That is a good idea, but then we should move this to an outer constructor.

PR #55, opened 2018-07-31
Simplify test dependencies

kdheepak:


deps/build.jl
```diff
@@ -1,75 +1,27 @@
-Pkg.add("InfoZIP")
```
jd-lara: If we do the import and the user doesn't have InfoZip installed the build will fail. 

I saw you added it to the requirements.

PR #61, opened 2018-08-06
Ramp Checks

### LINKED ISSUE #50: Time step consistency between ramping rates and time series

sourabhdalvi:
Added Ramp Checks 
Fixed polynomial cost functions
Added default Ramp rates in constructors
Added New scalingfactor with more than one points to find time-delta for ramp-check
Fixes #50


deps/build.jl
```diff
@@ -14,9 +14,10 @@ const DATA_FOLDER = Pkg.dir("PowerSystems/data")
 function download_data()
 
     if !isdir(DATA_FOLDER)
-        mkdir(DATA_FOLDER)
-        temp_folder = mktempdir()
-        temp_data_zip = joinpath(temp_folder, "data.zip")
+        mkpath(DATA_FOLDER)
```
jd-lara: This is making the Windows check fail, why did you change it back? Please remove this commit.

src/PowerSystems.jl
```diff
@@ -71,6 +71,7 @@ using PowerModels
 using DataFrames
 using CSV
 using Memento
+using StatsBase
```
jd-lara: What are you using StatsBase for?

src/models/generation/hydro_generation.jl
```diff
@@ -23,7 +23,7 @@ TechHydro(;installedcapacity = 0.0,
           realpowerlimits = @NT(min = 0.0, max = 0.0),
           reactivepower = nothing,
           reactivepowerlimits = nothing,
-          ramplimits = nothing,
+          ramplimits = @NT(up=0.0, down=0.0),
```
jd-lara: Why are you adding default 0.0 ramps?

src/models/generation/thermal_generation.jl
```diff
@@ -139,6 +139,6 @@ end
 ThermalGenSeason(; name = "init",
                 status = false,
                 bus = Bus(),
-                tech = nothing,
-                econ = nothing,
-                scalingfactor = TimeSeries.TimeArray(today(), [1.0])) = ThermalGenSeason(name, status, bus, tech, econ, scalingfactor)
+                tech = TechThermal(),
+                econ = EconThermal(),
+                scalingfactor = TimeArray(collect(DateTime(today()):Hour(1):DateTime(today()+Day(1))), ones(25))) = ThermalGenSeason(name, status, bus, tech, econ, scalingfactor)
```
jd-lara: This default scaling factor creates 24 steps by default, there is ni need for that.

sourabhdalvi: I needed more than 1 step to measure the time delta, I can change it to have just 2 steps if you want

deps/build.jl
sourabhdalvi: I didn't make that change

src/PowerSystems.jl
sourabhdalvi: Not me again

src/models/generation/hydro_generation.jl
sourabhdalvi: I can change to 10.0 , it just needs to be something instead of a Nothing

src/models/generation/thermal_generation.jl
```diff
@@ -56,7 +56,7 @@ TechThermal(; realpower = 0.0,
           realpowerlimits = @NT(min = 0.0, max = 0.0),
           reactivepower = nothing,
           reactivepowerlimits = nothing,
-          ramplimits = @NT(up=0.0, down=0.0),
+          ramplimits = @NT(up=10.0, down=10.0),
```
jd-lara: The default should be nothing, I still don't get why are you adding values here. The whole point is to make it work when there are not ramping limits as a nothing.

src/models/generation/renewable_generation.jl
```diff
@@ -54,7 +54,7 @@ RenewableCurtailment(; name = "init",
                 bus= Bus(),
                 installedcapacity = 0.0,
                 econ = EconRenewable(),
-                scalingfactor = TimeArray(collect(DateTime(today()):Hour(1):DateTime(today()+Day(1))), ones(25))) = RenewableCurtailment(name, status, bus, installedcapacity, econ, scalingfactor)
+                scalingfactor =  TimeArray([round(now(),Hour(1)),round(now()+Hour(1),Hour(1))],ones(2))) = RenewableCurtailment(name, status, bus, installedcapacity, econ, scalingfactor)
```
jd-lara: Also, still don't get why you are making the time series in this format.

PR #67, opened 2018-08-16
Make PowerSystems v0.7 compatible

kdheepak:


jd-lara:
I am cleaning up all the upgrade branches into master and see where we are. We had three.

PR #69, opened 2018-08-25
Dcline modeling

### LINKED ISSUE #52: Create proper DC line structure
Related to #36

jd-lara:
Added Dc line support for two types of models and consistent Abstract hierarchy. 

Also improved the testing code in general. Closes #52

### DIFF
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -12,7 +12,8 @@ export Branch
 export Network
 export Line
 export DCLine
-export SimpleDCLine
+export HVDCLine
+export VSCDCLine
 export Transformer2W
 export TapTransformer
 export PhaseShiftingTransformer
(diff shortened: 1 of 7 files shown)


kdheepak:
Why are you using `try ... finally ... end` in tests?

jd-lara:
So it is possible to test each individual test independently and get a report at the end of which in particular failed. Check the whole test code, I updated several parts

ccoffrin:
Instead of using try finally, have you tried just wrapping your imports in another `@testset` block?  This is how I accomplish this in PowerModels.

jd-lara:
@ccoffrin What I am trying to do here is to create a `@testset` for all the constructors. With this syntax, each constructor is treated as an individual test such that we run the test we can independently know how many constructors work and how many don't. 

https://github.com/NREL/PowerSystems.jl/blob/c5e6d234c60457a736901d42ecee1e3f01bef25f/test/runtests.jl#L7-L17

The first test set targets loading the `*.jl` files and the second targets each constructor individually. 

https://github.com/NREL/PowerSystems.jl/blob/master/test/constructors.jl

Issue #64, opened 2018-08-13
update parsing code to v0.7 


jd-lara:
Please consider that `Pkg.dir()` doesn't work anymore. Check the code in `readnetworkdara.jl` as template on how to navigate directories.

jd-lara:
This issue might require some changes in the working branch of PowerModels.

jd-lara:
Made some updates to the testing code to improve tests of parsing.

jd-lara:
Remove Memento warnings, otherwise, it is impossible to follow the tests results.

jd-lara:
closed with #74

PR #74, opened 2018-08-29
Fix parsing

claytonpbarrows:
addresses #64 
depends on a [PowerModels#moi-julia-v0.7 PR](https://github.com/lanl-ansi/PowerModels.jl/commit/b21a25dd698868d536bd1dd498504ef92b3094b7) to fix the dclines parsing

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -131,7 +131,7 @@ version = "0.19.0"
 
 [[JuMP]]
 deps = ["Calculus", "Compat", "DataStructures", "ForwardDiff", "MathOptInterface", "NaNMath"]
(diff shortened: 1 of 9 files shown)


claytonpbarrows:
it was coming in as:
```julia
julia> typeof(pv_dict["name"])
SubString{String}
```

jd-lara:
We need to find a better way to do that, the conversion can make the translation from Dict -> Struct incredibly slow.

jd-lara:
@claytonpbarrows Can you add a comment with the packages that you added using the development branches? So we can keep track of them and check when the new release comes out.

PR #91, opened 2018-10-19
Manifest update

jd-lara:
Check that this manifest works with your workflow.


jd-lara:
@claytonpbarrows this dependency on Pycall and python seems to be problematic. We should revert back to CSV, if we are going to use smaller files for testing then the one large file that is failing shouldn't become a problem.

jd-lara:
@claytonpbarrows and @kdheepak there is something really wrong with the data files. I tested CSV parsing from the RTS GMLC data reading directly from the original data in the RTS repo and all of them work perfectly fine. We should revert to use CSV and avoid using Pandas and any other Python dependency.

jd-lara:
@claytonpbarrows and @kdheepak there is something really wrong with the data files. I tested CSV parsing from the RTS GMLC data reading directly from the original data in the RTS repo and all of them work perfectly fine. We should revert to use CSV and avoid using Pandas and any other Python dependency.

PR #96, opened 2018-10-24
Add Windows Travis support

kdheepak:


kdheepak:
Travis does not support Julia on Windows at the moment. Closing for now, we can revisit later.

PR #97, opened 2018-10-24
Fix v1.0 issues

kdheepak:


kdheepak:
Tests pass locally for me in v1.0.1. 

screen shot 2018-10-24 at 3 55 48 pm


I can merge once tests on Travis pass.

kdheepak:
Tests pass on Travis. This PR is ready to be merged.

kdheepak:
Once @jd-lara says these changes are fine I can submit a PR to PowerModels with something similar.

src/parsers/dict_to_struct.jl
```diff
@@ -188,8 +188,8 @@ function gen_dict_parser(dict::Dict{String,Any})
     for (gen_type_key,gen_type_dict) in dict
         if gen_type_key =="Thermal"
             for (thermal_key,thermal_dict) in gen_type_dict
-                push!(Generators,ThermalDispatch(convert(String,thermal_dict["name"]),
```
jd-lara: Do we need a conversion here? The name should be a String. Many conversions might slow down the process.

kdheepak: It is possible that these are not required. I'll run the tests again without some of these conversions, and if they pass I'll commit the changes and merge this.

kdheepak:
I think there's issues with not using the `String` function. The dict doesn't seem to always contain a `String` in the name, it sometimes contains a `Int64` or a `Float64` (?).

```
  Got exception outside of a @test
  MethodError: Cannot `convert` an object of type Float64 to an object of type TechRenewable
  Closest candidates are:
    convert(::Type{S}, !Matched::T<:(Union{CategoricalString{R}, CategoricalValue{T,R} where T} where R)) where {S, T<:(Union{CategoricalString{R}, CategoricalValue{T,R} where T} where R)} at /Users/$USER/.julia/packages/CategoricalArrays/rQrLR/src/value.jl:86
    convert(::Type{T}, !Matched::T) where T at essentials.jl:154
    TechRenewable(::Float64, !Matched::Union{Nothing, NamedTuple{(:min, :max),Tuple{Float64,Float64}}}, !Matched::Union{Nothing, Float64}) at /Users/$USER/GitRepos/PowerSystems.jl/src/models/generation/renewable_generation.jl:6
    ...
  Stacktrace:
 ```

I'm going to merge this for now and we can figure out this issue separately.

Issue #109, opened 2018-10-31
Check consistency of the data when making a PowerSystem


jd-lara:
We have mixed units in the PowerSystem. There has to be consistency when creating a system and make sure all the data is in P.U.

jd-lara:
- [ ] Check the relationship of baseMVA w.rt. MW in data for limits
- [ ] Check the relationship of baseMVA w.rt. MW in data for peak load.
- [ ] Check the values of the provided angles w.r.t 3.14

claytonpbarrows:
I agree. However, this is similar to #98 and #99 where I did the opposite implementation and made the parser conform to the mixed-units.
This should also help address numerical scaling issues when we encounter them.

jd-lara:
@claytonpbarrows I am reviewing all the code that makes the checks as per the check-list above such that we always know the data is in p.u. and in radians when building a PowerSystem.

jd-lara:
It seems that there is no good way to find out if the data is in P.U. or in MW. The only device where this is possible is the branches. We need to figure out a way to guarantee that the data in a PowerSystem is all in PU. Otherwise we will run into problems in PowerSimulations.

claytonpbarrows:
Lets start with a short document describing the calculations of P.U. for each type of device. Short of extensive data checking and validation, flexible implementation (user defines if data is in P.U.) and good documentation is probably the best substitute.

ccoffrin:
If you want to be serious about units, you might also consider using a package like https://github.com/ajkeller34/Unitful.jl/ to be very explicit about it in the implementation layer.

PR #110, opened 2018-11-01
Pu review

### LINKED ISSUE #84: Fix angle limits checks and represent in radians

jd-lara:
closes #84 #109 

This PR implements a consistent way of checking and calculating the rate of the line. Also creates a new type for lines with transmission limits. Three test files are failing in the parsing stage but there seems to be something wrong with the Logging since it doesn't show which file is failing. 

Using a print statement I got that the failing parsing files are `case_ACTIVSg2000.m`, `case_ACTIVSg10k.m` and `case30.m`

After many attempts, I can't reproduce the error. In a jupyter notebook it parses ok. If I comment this line 
`sys_test = PowerSystem(Buses, Generators,Loads,Branches,Storages,float(ps_dict["baseMVA"]))` it works ok in the tests, but can't point exactly to where is failing. 

This code runs perfectly 

```Julia
f = "ACTIVSg2000.m"
@info "Parsing $f ..."
pm_dict = PowerSystems.parse_file(abspath(joinpath(dirname(Base.find_package("PowerSystems")), "../data/matpower",f)))
 @info "Successfully parsed $f to PowerModels dict"
 PowerSystems.make_mixed_units(pm_dict)
 @info "Successfully converted $f to mixed_units"
 ps_dict = PowerSystems.pm2ps_dict(pm_dict)
 @info "Successfully parsed $f to PowerSystems dict"
 Buses, Generators, Storages, Branches, Loads, LoadZones ,Shunts = PowerSystems.ps_dict2ps_struct(ps_dict)
 @info "Successfully parsed $f to PowerSystems devices"
 sys_test = PowerSystem(Buses, Generators,Loads,Branches,Storages,float(ps_dict["baseMVA"]))
 @info "Successfully parsed $f to PowerSystem struct"
```


src/models/branches/lines.jl
```diff
@@ -3,38 +3,52 @@ const From_To_Float = NamedTuple{(:from, :to),Tuple{Float64,Float64}}
...
     r::Float64 #[pu]
     x::Float64 #[pu]
     b::From_To_Float # [pu]
-    rate::FromTo_ToFrom_Float #MW
+    rate::Float64
     anglelimits::Min_Max #Degrees
 end
 
 function Line(name::String, available::Bool, connectionpoints::From_To_Bus,
-              r::Float64, x::Float64, b::From_To_Float, rate::FromTo_ToFrom_Float, anglelimits::Float64)
```
claytonpbarrows: Do we need to remove the directional flow limit definition? If so, fine... If not, I like having this feature.

src/models/branches/lines.jl
```diff
@@ -3,38 +3,52 @@ const From_To_Float = NamedTuple{(:from, :to),Tuple{Float64,Float64}}
...
+    connectionpoints::From_To_Bus
+    r::Float64 #[pu]
+    x::Float64 #[pu]
+    b::From_To_Float # [pu]
+    flowlimits::FromTo_ToFrom_Float #MW
+    rate::Float64
+    anglelimits::Min_Max #Degrees
 end
 
-function Line(name::String, available::Bool, connectionpoints::From_To_Bus,
```
claytonpbarrows: I like the addition of the `ConstrainedLine`. Is the constraint only with respect to the flow limit? Or is this also to w.r.t other constraints? If it's just flow limits, we should call it `MonitoredLine` as that's more in line with the industry terminology. Also, we should create an `UnMonitoredLine`.

src/models/generation.jl
```diff
@@ -2,17 +2,12 @@ abstract type
     Generator <: PowerSystemDevice
 end
 
+include("generation/tech_common.jl")
+include("generation/econ_common.jl")
 include("generation/renewable_generation.jl")
 include("generation/thermal_generation.jl")
 include("generation/hydro_generation.jl")
 
-#=
```
claytonpbarrows: I think I like this, but it will break PowerSimulations.

src/models/branches/lines.jl
jd-lara: The constraint is only w.r.t to flow in this form. It can be extended to other limits but I think those can be generalized to flow in most cases. I will change the name to `Monitored` but I don't see why we would need the `Unmonitored`

src/models/generation.jl
jd-lara: I shouldn't because I didn't generalize this yet. Just separated it, for now, to get PowerSimulations working I don't want to touch it.

jd-lara:
@claytonpbarrows any ideas about why the parsing might be failing here?

src/models/branches/lines.jl
jd-lara: The idea is to separate the notion of the rate which the inherent physical limitations of the line from the flow_limits from other constraints. This will remove the definition conflict with PowerModels.

src/models/branches/lines.jl
claytonpbarrows: You can disregard the `unmonitored` part of my comment. On further thought, there are better ways to filter the enforcement of flow constraints

claytonpbarrows:
I don't know right now. I can try to check on it tomorrow.

jd-lara:
The merge looks ok to me after resolving the conflicts

PR #111, opened 2018-11-01
Ts scaling

claytonpbarrows:
this PR:
 - adds a shortcut PowerSystem() constructor that takes paths to a load flow case, and a time series directory to create a PowerSystem
 - adds a `getresolution()` function that returns the time resolution of a `TimeArray`

### DIFF
--- a/src/parsers/dict_to_struct.jl
+++ b/src/parsers/dict_to_struct.jl
@@ -140,7 +140,13 @@ function add_time_series(Device_dict::Dict{String,Any}, df::DataFrames.DataFrame
     for (device_key,device) in Device_dict
         if device_key in map(String, names(df))
             ts_raw = df[Symbol(device_key)]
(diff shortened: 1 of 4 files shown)


src/parsers/dict_to_struct.jl
```diff
@@ -140,7 +140,13 @@ function add_time_series(Device_dict::Dict{String,Any}, df::DataFrames.DataFrame
     for (device_key,device) in Device_dict
         if device_key in map(String, names(df))
             ts_raw = df[Symbol(device_key)]
-            Device_dict[device_key]["scalingfactor"] = TimeSeries.TimeArray(df[:DateTime],ts_raw)
+            if maximum(ts_raw) <= 1.0
+                @info "assumed time series is a scaling factor for $device_key"
+                Device_dict[device_key]["scalingfactor"] = TimeSeries.TimeArray(df[:DateTime],ts_raw)
+            else
+                @info "assumed time series is MW for $device_key"
```
jd-lara: This is a nice addition.

PR #116, opened 2018-11-07
Update to v0.2.0

kdheepak:


jd-lara:
Are we tagging a new version soon?

jd-lara:
Yes, the issues opened about printing and modifying data are important ones to figure out

PR #117, opened 2018-11-14
Create a detailed show for PowerSystemDevice

jjstickel:
I created a generic long version print for power system devices. There are quite a few, and so this will help until custom print functions are created for each device (as desired). All it does is print the fieldnames and their values, but because the fieldnames are mostly descriptive already, it seems useful.


jjstickel:
@jd-lara, please see my new (amended) commit.

Since the members of TechnicalParams do not have a "name" field, the "short" print for those just show the type right now. I suggest we merge this and then I can work on customized print functions for various types.

PR #123, opened 2018-11-28
Fix cdmparser

claytonpbarrows:
- updates the cdm parser and fixes bugs
 - updates the times eries assignment to improve flexibilitiy
 - requires update of powersystemstestdata

### DIFF
--- a/src/parsers/cdm_parser.jl
+++ b/src/parsers/cdm_parser.jl
@@ -26,14 +26,13 @@ function read_csv_data(file_path::String)
     for d_file in files
         try
             if match(REGEX_IS_FOLDER, d_file) != nothing
-                @info "Parsing csv timeseries files in $d_file ..."
+                @info "Parsing csv files in $d_file ..."
                 d_file_data = Dict{String,Any}()
                 for file in readdir(joinpath(file_path,d_file))
                     if match(REGEX_DEVICE_TYPE, file) != nothing
-                        @info "Parsing csv timeseries data in $file ..."
+                        @info "Parsing csv data in $file ..."
                         fpath = joinpath(file_path,d_file,file)
                         raw_data = CSV.File(fpath) |> DataFrame
-                        #raw_data = DataFrames.DataFrame(Pandas.read_csv(fpath))
                         d_file_data[split(file,r"[.]")[1]] = raw_data
                     end
                 end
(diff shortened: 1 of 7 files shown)


src/parsers/cdm_parser.jl
```diff
@@ -148,93 +252,108 @@ function gen_csv_parser(gen_raw::DataFrames.DataFrame, Buses::Dict{Int64,Any})
...
-                                                                        "reactivepowerlimits" => (min=float(gen_raw[gen,14]),max=float(gen_raw[gen,13])),
-                                                                        "ramplimits" => (up=gen_raw[gen,17],down=gen_raw[gen,17]),
-                                                                        "timelimits" => (up=gen_raw[gen,16],down=gen_raw[gen,15])),
-                                            "econ" => Dict{String,Any}("capacity" => gen_raw[gen,11],
+                                            "tech" => Dict{String,Any}("activepower" => gen_raw[gen,Symbol("MW Inj")],
+                                                                        "activepowerlimits" => (min=tryparse(Float64,"""$(gen_raw[gen,Symbol("PMin MW")])"""),max=pmax),
+                                                                        "reactivepower" => tryparse(Float64,"""$(gen_raw[gen,Symbol("MVAR Inj")])"""),
+                                                                        "reactivepowerlimits" => (min=tryparse(Float64,"""$(gen_raw[gen,Symbol("QMin MVAR")])"""),max=tryparse(Float64,"""$(gen_raw[gen,Symbol("QMax MVAR")])""")),
```
jd-lara: These type conversions to Symbol can be expensive computationally. Is there a reason to use symbols and not strings in the resulting dict?

claytonpbarrows: In this particular case, the symbol conversion is to access the column headers in a dataframe which are symbols. The previous implementation used numeric column indices that were hard-coded. Perhaps the better approach would be to create a column index map so that numeric indices are used to access columns, but the parser doesn't depend on a certain order of columns.... this could also be made a bit more flexible to accommodate different column names.

claytonpbarrows: only RTS... seems fine so far. I'll implement my suggested changes. There are a few areas that I'm worried about performance, but those can be addressed as we scale up.

jd-lara:
@claytonpbarrows what's the status of this branch? is it ready to merge?

claytonpbarrows:
Yep, it should be good to go once approved

claytonpbarrows:
@jd-lara @kdheepak can you review/approve?

jd-lara:
@claytonpbarrows this code needs more comprehensive testing. Coverage of the CDM parser is only at +63.86%

claytonpbarrows:
Based on my analysis, and the issue here: https://github.com/JuliaCI/Coverage.jl/issues/187 
Despite the low coverage report, I'm fairly confident that the tests actually touch most, if not all, of the code in the cdmparser.jl
@jd-lara or @kdheepak unless you have objections or solutions, I'm going to merge.

jd-lara:
I am ok with the merge

PR #124, opened 2018-11-28
ravamped printing of PowerSystem type

jjstickel:
- converted NamedTuple for generators to a dedicated structure, "GenClasses" 
- smart printing of PowerSystem type
- overload of summary command (in addition to show) for PowerSystem (and the new GenClasses) -- helpful with output of varinfo()

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -1,14 +1,11 @@
 ### Struct and different Power System constructors depending on the data provided ####
 
-struct PowerSystem{T <: Union{Nothing,Array{ <: ThermalGen,1}},
-                   R <: Union{Nothing,Array{ <: RenewableGen,1}},
-                   H <: Union{Nothing,Array{ <: HydroGen,1}},
-                   L <: ElectricLoad,
+struct PowerSystem{L <: ElectricLoad,
                    B <: Union{Nothing,Array{ <: Branch,1}},
                    S <: Union{Nothing,Array{ <: Storage,1}}
                    }
     buses::Array{Bus,1}
-    generators::NamedTuple{(:thermal, :renewable, :hydro), Tuple{T, R, H}}
+    generators::GenClasses
     loads::Array{L,1}
     branches::B
     storage::S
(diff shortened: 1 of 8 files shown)


jjstickel:
I have in mind to create some print tests before merging this, but I wanted to make this preview available now.

First, please let me know if my changed type for PowerSystem.generators is OK. Scanning to see how it is later used, I _think_ my changes are OK. The existing tests in PowerSystems passed, but I haven't yet tried running tests in PowerSimulations with this change.

src/base.jl
```diff
@@ -1,14 +1,11 @@
...
-                   R <: Union{Nothing,Array{ <: RenewableGen,1}},
-                   H <: Union{Nothing,Array{ <: HydroGen,1}},
-                   L <: ElectricLoad,
+struct PowerSystem{L <: ElectricLoad,
                    B <: Union{Nothing,Array{ <: Branch,1}},
                    S <: Union{Nothing,Array{ <: Storage,1}}
                    }
     buses::Array{Bus,1}
-    generators::NamedTuple{(:thermal, :renewable, :hydro), Tuple{T, R, H}}
+    generators::GenClasses
```
kdheepak: I like the change but I'm not sure what this will break in PowerSimulations.

jd-lara: This change will break PowerSimulations. We had an approach like this before but it is preferable to have a NamedTuple so it is possible to make iterations.

jjstickel:
@jd-lara I had that exact concern. In my initial search for uses of "generators" in PowerSimulations, I did not find where iteration was used, but I see it now in `device_retreval.jl`. Do you have an objection to use the struct type if I implement an iterator for it? See latest commit efaaf22906346b4e990ef44dbca4798d85d9548c . I tested this so far with:
```for source in sys.generators
    if source isa Nothing
        println("nothing")
    else
        println(source)
    end
end
```
I recognize this _almost_ a case of the tail wagging the dog, but I think there are advantages (in addition to printing) for having PowerSystem.generators be a custom type rather than a generic NamedTuple. It cleans up the PowerSystem constructor code, and is consistent with all the other subcomponents being custom types. If you are amenable, I will gladly run PowerSimulations tests to make sure nothing breaks.

jd-lara:
@kdheepak wrote some tests for printing a few months ago. It looks kind off like this 

```Julia
@assert "$PowerSystem" == "PowerSystems.PowerSystem(buses=5, generators=7, branches=6)"
```

Basically compares the output with a string that resembles what you expect to see when printing.

jjstickel:
Please check whether the last two commits here look OK to you:
https://github.com/NREL/PowerSystems.jl/pull/124/commits/dcced39462edaaf18e200d1838751f3662d2f0ff I overwrote the printing of TimeArray from TimeSeries because their implementation does not differentiate between REPL (text/plain) output and print() output.
https://github.com/NREL/PowerSystems.jl/pull/124/commits/00c31e72339772846cfb8ead287d969927b3c795 Adds the printing tests. I think I did the right thing here, but let me know if I missed something important or if I did something naive. If I missed a test for the printing of your favorite subcomponent, I would prefer to add it later.

claytonpbarrows:
This looks good to me. I think you can merge

PR #130, opened 2018-12-27
Bug fix parsing matpower files with gentype/genfuel fields + case ins…

### LINKED ISSUE #128: parsestandardfiles parses all generators as thermal generators
-- at least from Matpower inputs (e.g., ACTIVSg2000.m, RTS_GMLC.m)
-- keys in output dictionary (generator names) have extra spaces at end

kdayday:
…ensitive
Fixes #128 for Matpower files (separate from the fix for RTS-GMLC repo)

### DIFF
--- a/src/parsers/pm2ps_parser.jl
+++ b/src/parsers/pm2ps_parser.jl
@@ -268,20 +268,20 @@ function read_gen(data,Buses)
         gen_name =[]
         type_gen =[]
         for (d_key,d) in data["gen"]
-            haskey(d,"fuel") ? fuel =d["fuel"] : fuel = "generic"
-            haskey(d,"type") ? type_gen = d["type"] : type_gen = "generic"
+            haskey(d,"fuel") ? fuel =d["fuel"] : haskey(data,"genfuel") ? fuel = data["genfuel"][d_key]["col_1"] : fuel = "generic"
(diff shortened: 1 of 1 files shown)


src/parsers/pm2ps_parser.jl
```diff
@@ -268,20 +268,20 @@ function read_gen(data,Buses)
...
+            elseif uppercase(fuel) in ["SOLAR","WIND"] || uppercase(type_gen) in ["W2","PV"]
                 bus = find_bus(Buses,d)
-                if type_gen == "PV"
+                if uppercase(type_gen) == "PV"
                     Generators["Renewable"]["PV"][gen_name] =  make_ren_gen(gen_name, d, bus)
-                elseif type_gen == "RTPV"
+                elseif uppercase(type_gen) == "RTPV"
                     Generators["Renewable"]["RTPV"][gen_name] = make_ren_gen(gen_name, d, bus)
-                elseif type_gen in ["WIND","W2"]
+                elseif uppercase(type_gen) in ["WIND","W2"]
```
claytonpbarrows: This seems fine, but we should open an issue to be able to pass in a mapping of `gen_type=>device_type` such as `"WIND"=>"RenewableCurtail"` so that we don't hardcode a new mapping every time we parse a new file.

claytonpbarrows: see #131

PR #136, opened 2019-01-23
enable tests for printing

jd-lara:


jd-lara:
Fix the basepower in the printing tests

PR #143, opened 2019-02-01
Docstrings

jjstickel:
Some work on docstrings for PowerSystems (the module), the PowerSystem struct,  and structs in topological_elements, and tech_common. 
@jd-lara suggested merging and then continue adding/revising docstrings as we work on other tasks. I used the keyword `DOCTODO` to bring attention to docstring items to address.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -1,9 +1,39 @@
 ### Struct and different Power System constructors depending on the data provided ####
 
+"""
+    PowerSystem
+
+A power system defined by fields for buses, generators, loads, branches, and
+overall system parameters.
+
+# Constructor
+```julia
+PowerSystem(buses, generators, loads, branches, storage, basepower; kwargs...)
+```
+
+# Arguments
+
+* `buses`::Array{Bus,1} : an array of buses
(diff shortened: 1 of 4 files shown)


src/base.jl
```diff
@@ -133,6 +167,10 @@ struct PowerSystem{L <: ElectricLoad,
 
 end
 
+# DOCTODO JJS What is the purpose of this statement? OK, it looks like a
```
jd-lara: Correct, this is for missing fields and also to enable keyword arguments in the constructors

src/models/generation/tech_common.jl
```diff
@@ -16,26 +23,21 @@ TechRenewable(; InstalledCapacity = 0, reactivepowerlimits = nothing, powerfacto
 
 Data Structure for the economical parameters of thermal generation technologies.
     The data structure can be called calling all the fields directly or using named fields.
-    Two examples are provided one with minimal data definition and a more comprenhensive one
+    Two examples [DOCTODO only one example exists below] are provided one with minimal data definition and a more comprenhensive one
```
jd-lara: Take the example data from the `data5_pu.jl` file.

PR #152, opened 2019-02-19
remove name::String argument from make_forecast_dict()

jjstickel:
The `name` argument was not being used.


jjstickel:
It looks like one of the parsing tests is failing with this change. I will look into it. My guess is that the test was not really correctly implemented.

jjstickel:
OK, the calls to `make_forecast_dict()` in `test/readforecastdata.jl` were using the `name` argument that I deleted. As far as I can tell, using this argument does nothing, hence this pull request, and so I removed that argument in the tests as well. If I am missing something here, do let me know! Anyway, Travis is passing now.

PR #153, opened 2019-02-21
Added coding style guide.

daniel-thom:
The content here is based on discussions I had with Dheepak.  Consider any rule up for discussion.


CodingStyleGuide.md
```diff
@@ -0,0 +1,106 @@
...
+
+## One-line Conditionals
+Instead of writing a complete if statement the Julia code base uses this idiom
+frequently:  ```<cond> && <statement>```
+[Example](https://docs.julialang.org/en/v1.0/manual/control-flow/#Short-Circuit-Evaluation-1):
+>
+    function fact(n::Int)
+       n >= 0 || error("n must be non-negative")
+       n == 0 && return 1
+       n * fact(n-1)
```
kdheepak: I think using shortcutting `&&` is acceptable, whereas shortcutting `||` is not very readable. As a general rule though I'd avoid both.

PR #157, opened 2019-03-04
Generate Documentation on CI

kdheepak:
This will generate documentation on Travis-CI and push to gh-pages automatically.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,9 +31,6 @@ script:
   - julia --check-bounds=yes --depwarn=no --color=yes --project -e "import Pkg; Pkg.test(coverage=true);"
 
 after_success:
-  # push coverage results to Codecov
-  - julia -e 'using Pkg; cd(Pkg.dir("PowerSystems")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
-  # push coverage results to Codecov
   - julia -e 'using Pkg; cd(Pkg.dir("PowerSystems")); Pkg.add("Coverage"); using Coverage; Codecov.submit(Codecov.process_folder())'
 
 jobs:
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,5 +32,17 @@ script:
 
 after_success:
   # push coverage results to Codecov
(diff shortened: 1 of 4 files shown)


docs/Manifest.toml
```diff
@@ -0,0 +1,80 @@
...
+deps = ["Random", "Serialization", "Sockets"]
+uuid = "8ba89e20-285c-5b6f-9357-94700520ee1b"
+
+[[DocStringExtensions]]
+deps = ["LibGit2", "Markdown", "Pkg", "Test"]
+git-tree-sha1 = "1df01539a1c952cef21f2d2d1c092c2bcf0177d7"
+uuid = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
+version = "0.6.0"
+
+[[Documenter]]
```
jd-lara: I don't think you need to add the documenter to the package dependencies. You can call `Pkg.add` when calling the Documentation Stage

kdheepak: This is a separate Project.toml just for docs. So it's not part of package dependencies. There are other documentation package extensions (see [DocStringExtensions.jl](https://github.com/JuliaDocs/DocStringExtensions.jl)) that I'd like to add in the future and having a separate Project.toml for this can be useful.

.travis.yml
```diff
@@ -32,5 +32,17 @@ script:
 
 after_success:
   # push coverage results to Codecov
-  - julia --project -e 'import Pkg; Pkg.add("Documenter"); include("docs/make.jl")'
-  - julia --project -e 'import Pkg; Pkg.add("Coverage"); using Coverage; Codecov.submit(Codecov.process_folder())'
+  - julia -e 'using Pkg; cd(Pkg.dir("PowerSystems")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
```
jd-lara: We Should also remove this line since we don't use coveralls.

kdheepak: Resolved in 2e373c0.

kdheepak:
Okay, I'll merge this when tests pass.

Issue #125, opened 2018-12-04
Missing tests


jd-lara:
~~Add regression tests for parsing @kdheepak #107~~ 
~~`parsestandardfiles()` with a time series folder specification~~
- [x] `cdm_parser` @claytonpbarrows 
~~- [ ] `json2ps_parser.jl` @claytonpbarrows  can you confirm this is still necessary? ~~
- [x] @errors in `pm2ps_parser.jl`
- [x] `pm_io` @kdheepak can you take a look a this folder missing tests?  #102 
- [x] `ybus_calculations.jl` @jd-lara #80 #79 
- [x] `print.jl` @jjstickel

jd-lara:
@daniel-thom We had this issue opened about the parsing testing that is relevant to today's conversation. Many of the missing tests are relevant to the parsing.

daniel-thom:
@jd-lara Regarding the @error messages in pm2ps_parser.jl, is your point that we should have negative tests that trip these errors?  Secondly, if they are fatal errors then I propose we throw an exception instead of just logging it.  Do you agree?

jd-lara:
@claytonpbarrows and @daniel-thom what's the status of this issue?

PR #169, opened 2019-03-19
Fix cdmparser

### LINKED ISSUE #126: create services and DC lines with cdm_parser
extension of #122

claytonpbarrows:
Contributes to #125 and closes #126

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -184,7 +184,7 @@ PowerSystem(; buses = [Bus()],
 function PowerSystem(ps_dict::Dict{String,Any}; kwargs...)
         Buses, Generators, Storage, Branches, Loads, LoadZones, Shunts, Services = ps_dict2ps_struct(ps_dict)
         sys = PowerSystem(Buses, Generators,Loads,Branches,Storage,ps_dict["baseMVA"]; kwargs...);
-        return sys
+        return sys , Services
 end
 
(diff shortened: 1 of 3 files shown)


src/base.jl
```diff
@@ -184,7 +184,7 @@ PowerSystem(; buses = [Bus()],
 function PowerSystem(ps_dict::Dict{String,Any}; kwargs...)
         Buses, Generators, Storage, Branches, Loads, LoadZones, Shunts, Services = ps_dict2ps_struct(ps_dict)
         sys = PowerSystem(Buses, Generators,Loads,Branches,Storage,ps_dict["baseMVA"]; kwargs...);
-        return sys
+        return sys , Services
```
kdheepak: Are we sure we want to return `sys` and `Services` here? This is going to make the return type across the various constructors type unstable. If you absolutely want to do this, you can create a new function `function get_powersystem_and_services(args...) end` and `return sys, Services`.

src/base.jl
```diff
@@ -184,7 +184,7 @@ PowerSystem(; buses = [Bus()],
 function PowerSystem(ps_dict::Dict{String,Any}; kwargs...)
         Buses, Generators, Storage, Branches, Loads, LoadZones, Shunts, Services = ps_dict2ps_struct(ps_dict)
```
kdheepak: These variable names should all be lower case.

PR #170, opened 2019-03-20
LODF radial line fix

### LINKED ISSUE #165: LaPack error in buildlodf()
Hi folks,

Greetings from the Spine project! 

https://github.com/Spine-project
www.spine-model.org/spine_toolbox.htm

I've successfully calculated ptdfs for my network using `PowerSystems.buildptdf()`

However, I'm getting Lapack errors in `buildlodf()` at `LODF = gemm('N','N',H,getri!(Dem,dipiv))`. I have a suspicion it's for outages that result in islanding and/or orphaned tails.

Any suggestions?

sourabhdalvi:
Allows for LODF calculations with radial lines.
Also added test for LODF calculation with radial lines.
fixes #165


claytonpbarrows:
> Instead of using -1 to substitute for `NaN` use `nothing`. Julia has a pretty good interface to handle Arrays with missing values.

But in this case, replacing where MATPOWER places `NaN` and `-Inf`, with `0.0` and `-1.0` is correct since is is reflecting the proportion of a radial line's flow that gets distributed to other lines (`0.0`) and to itself (`-1.0`) when the radial line is removed from service.
Right?

Issue #78, opened 2018-09-14
Move parsing info/warnings to a Log file 


jd-lara:
We should move the printouts of the warnings/info from the parsing into a log file and remove the printouts on the screen.

jd-lara:
@claytonpbarrows can you look at this while fixing the parsing? We have too many printouts while parsing and it is not really informative to the user.

claytonpbarrows:
@kdheepak did you resolve this?

jd-lara:
@kdheepak can we close this now?

daniel-thom:
Implemented by pull request #171

PR #171, opened 2019-03-21
Logging improvements

daniel-thom:
This will fix issue #78.  The MultiLogger object could be moved to a 3rd-party module.  I'm proposing that we experiment with it here first.

### DIFF
--- a/src/utils/logging.jl
+++ b/src/utils/logging.jl
@@ -1,6 +1,7 @@
 
 import Logging
 
+export configure_logging
 export open_file_logger
 export MultiLogger
 export LogEvent
@@ -8,6 +9,72 @@ export LogEventTracker
 export report_log_summary
 export get_log_events
 
+
+"""
+    configure_logging([console, console_stream, console_level,
+                       file, filename, file_level, file_mode,
+                       tracker, set_global])
+
+Creates console and file loggers per caller specification and returns a MultiLogger.
+
+**Note:** If logging to a file users must call Base.close() on the returned MultiLogger to
+ensure that all events get flushed.
+
+# Arguments
+- `console::Bool=true`: create console logger
+- `console_stream::IOStream=stderr`: stream for console logger
+- `console_level::Logging.LogLevel=Logging.Error`: level for console messages
+- `file::Bool=true`: create file logger
+- `filename::String=log.txt`: log file
+- `file_level::Logging.LogLevel=Logging.Info`: level for file messages
+- `file_mode::String=w+`: mode used when opening log file
+- `tracker::Union{LogEventTracker, Nothing}=LogEventTracker()`: optionally track log events
+- `set_global::Bool=true`: set the created logger as the global logger
+
+# Example
+```julia
+logger = configure_logging(filename="mylog.txt")
+```
+"""
+function configure_logging(;
+                           console=true,
+                           console_stream=stderr,
+                           console_level=Logging.Error,
+                           file=true,
+                           filename="log.txt",
+                           file_level=Logging.Info,
+                           file_mode="w+",
+                           tracker=LogEventTracker(),
+                           set_global=true
+                          )::MultiLogger
+    if !console && !file
+        error("At least one of console or file must be true")
+    end
+
+    loggers = Array{Logging.AbstractLogger, 1}()
+    if console
+        console_logger = Logging.ConsoleLogger(console_stream, console_level)
+        push!(loggers, console_logger)
+    end
+
+    if file
+        io = open(filename, file_mode)
+        try
+            file_logger = Logging.SimpleLogger(io, file_level)
+            push!(loggers, file_logger)
+        catch
+            close(io)
+            rethrow() 
+        end
+    end
+
+    logger = MultiLogger(loggers, tracker)
+    if set_global
+        Logging.global_logger(logger)
+    end
+
+    return logger
+end
 
 """
     open_file_logger(func, filename[, level, mode])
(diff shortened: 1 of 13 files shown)


daniel-thom:
> I am fine in general with the code, but I wouldn't put the testing code in the README file. This should be a docstring.

I wasn't entirely certain this was the best place.  I received a request about how to do these customizations.  They aren't necessarily obvious by looking at the code; this makes it very clear.  Where exactly do you suggest putting the docstrings?

jd-lara:
Inside of the Logging code. We can also keep a Copy in the SIIP repo for reference.

daniel-thom:
> Inside of the Logging code. We can also keep a Copy in the SIIP repo for reference.
We may not be talking about the same thing.  The content I put into the README is specific to running PowerSystems unit tests in the REPL.  It has nothing to do with the code in logging.jl.

jd-lara:
Then create a page in the docs?

daniel-thom:
I'd be OK with that.  @claytonpbarrows, @kdheepak any thoughts?

claytonpbarrows:
Sounds good.

test/runtests.jl
```diff
@@ -60,15 +61,29 @@ macro includetests(testarg...)
...
+
+    return log_level
 end
-global_logger(ConsoleLogger(gl.stream, log_level))
 
-# Testing Topological components of the schema
-@testset "Begin PowerSystems tests" begin
-    @includetests ARGS
+console_level = get_logging_level("PS_LOG_LEVEL", "Error")
+file_level = get_logging_level("PS_CONSOLE_LOG_LEVEL", "Info")
```
kdheepak: Is the environment name here right? 

```
file_level = get_logging_level("PS_CONSOLE_LOG_LEVEL", "Info")
```

We getting a `PS_CONSOLE_LOG_LEVEL` but storing the value in variable called `file_level`.

test/runtests.jl
```diff
@@ -60,15 +61,29 @@ macro includetests(testarg...)
...
+
+open_file_logger(LOG_FILE, file_level) do file_logger
+    console_logger = ConsoleLogger(stderr, console_level)
+    multi_logger = MultiLogger([console_logger, file_logger],
+                               LogEventTracker((Logging.Info, Logging.Warn, Logging.Error)))
+    global_logger(multi_logger)
```
kdheepak: I would go one step further and create utility functions around this:

```
    console_logger = ConsoleLogger(stderr, console_level)
    multi_logger = MultiLogger([console_logger, file_logger],
                               LogEventTracker((Logging.Info, Logging.Warn, Logging.Error)))
    global_logger(multi_logger)
```

So that the user can call say `create_logger(file=true, console=false)` and use that function in this context. This is not high priority.

test/runtests.jl
daniel-thom: Doh.  Fixed, thanks.

kdheepak: If a user sets these environment variables, at the moment, they will only take effect when they run tests. If I want to create a file logger when I'm using `PowerSystems` to parse some file, I wouldn't get these logging levels used unless I implemented the same as above, right?

daniel-thom: That is correct.  Logging has to be configured by the application, whatever that is.  I'm guessing that that is going to be some higher-level entity, like a consumer of PowerSimulations, right?  I'm totally open to making this easier, but I probably need to better understand the use cases.

daniel-thom:
The test instructions are now in the docs directory.

src/utils/logging.jl
```diff
@@ -0,0 +1,188 @@
...
+    count::Int
+    suppressed::Int
+end
+
+LogEvent(file, line, id, message, level) = LogEvent(file, line, id, message, level, 1, 0)
+
+struct LogEventTracker
+    events::Dict{Logging.LogLevel, Dict{Symbol, LogEvent}}
+
+    LogEventTracker(events::Dict{Logging.LogLevel, Dict{Symbol, LogEvent}}) = new(events)
```
kdheepak: Is this inner constructor necessary?

test/runtests.jl
kdheepak: In the future we need a way to make sure other packages can use a `PowerSystems` logger in conjunction with a logger that they create. I'm not sure how that would work though?

daniel-thom: I added an exported function called configure_logging.  I ran a test with this function in PowerSimulations - it worked fine.  
logger = PowerSystems.configure_logging(...)

Every log statement is tagged with @<module name> <file:line>.
Getting different loggers (and log files) for different modules is possible but not practical for this PR.

src/utils/logging.jl
daniel-thom: Without it I get a method overwrite error for the default constructor that takes Any.  I'm choosing to leave it for now.  I'm open to discussion, though.

daniel-thom:
I think I've addressed the open questions and actions for this PR.  I propose that we merge it and start trying it out.  PowerSimulations (and associated notebooks) can start using the functionality by simply importing the desired functions and objects from PowerSystems.  It may be as simple as `import PowerSystems: configure_logging`.

claytonpbarrows:
Sounds good to me.

kdheepak:
Looks great!

test/runtests.jl
kdheepak: Resolved in aa166cd

src/utils/logging.jl
kdheepak: In this sort of a situation, I would add a comment for why a inner constructor is necessary.

daniel-thom: Good point, added.

src/utils/logging.jl
```diff
@@ -0,0 +1,276 @@
...
+function report_log_summary(logger::MultiLogger)::String
+    if logger.tracker == nothing
+        error("log event tracking is not enabled")
+    end
+
+    return report_log_summary(logger.tracker)
+end
+
+"""Flush any file streams."""
+function Base.flush(logger::MultiLogger)
```
daniel-thom: Note that I added this after our discussion today.  This is what Julia devs have frequently posted on forums when people ask why log files haven't been updated.

Issue #173, opened 2019-03-22
Julia reports method overwrite warnings during execution of tests.


daniel-thom:
Julia reports the following warnings when the tests are run.  The code needs to be fixed.

[ Info: Recompiling stale cache file /home/dthom/.julia/compiled/v1.1/PowerSystems/pY4gF.ji for PowerSystems [c512b964-a17e-11e8-344c-11d7c2eac237]
WARNING: Method definition summary(IO, Base.Dict{String, Any}) in module PowerSystems at /mnt/c/Users/dthom/Sandboxes/daniel-thom-PowerSystems.jl/src/parsers/pm_io/data.jl:188 overwritten at /mnt/c/Users/dthom/Sandboxes/daniel-thom-PowerSystems.jl/src/parsers/im_io/data.jl:132.
WARNING: Method definition #summary(Any, typeof(PowerSystems.summary), IO, Base.Dict{String, Any}) in module PowerSystems overwritten.
┌ Warning: Replacing docs for `PowerSystems.summary :: Tuple{IO,Dict{String,Any}}` in module `PowerSystems`
└ @ Base.Docs docs/Docs.jl:223
WARNING: Method definition show(IO, TimeSeries.TimeArray{T, N, D, A} where A<:AbstractArray{T, N} where D<:Dates.TimeType where N where T) in module TimeSeries at /home/dthom/.julia/packages/TimeSeries/bbwst/src/timearray.jl:167 overwritten in module PowerSystems at /mnt/c/Users/dthom/Sandboxes/daniel-thom-PowerSystems.jl/src/utils/print_timearray.jl:134.

daniel-thom:
@jjstickel It looks like you intentionally overwrote the show method in print_timearray.jl.  What do you think should be done here?  Does your improvement need to be pushed upstream?  It isn't great that we get this warning constantly, but if this is the best behavior, we can probably live with it.

jd-lara:
@daniel-thom looks like the PR was already merged in TimeSeries.jl

https://github.com/JuliaStats/TimeSeries.jl/pull/399

We just need to update our code and the TimeSeries version.

jjstickel:
That's right. @jd-lara beat me to reporting on my PR upstream in TimeSeries.jl. Do we wait for TimeSeries to release another version, or use a commit-tag for our package dependencies? It doesn't look like it releases regularly.

jd-lara:
Maybe ping in the PR for the release?

[...]

PR #176, opened 2019-03-22
Fixed method overwrite problem in power model parser

daniel-thom:
This will partially fix issue #173.


jd-lara:
We could probably do the same for the isapprox method

PR #183, opened 2019-03-26
Improve timeseries

### LINKED ISSUE #179: update PowerSystem struct
related to #112  and #121

### LINKED ISSUE #180: remove scalingfactor fields
related to #112 and the discussion on [gitter](https://gitter.im/NREL/PowerSystems.jl)

### LINKED ISSUE #181: populate forecasts with time series data instead of scalingfactors
related to #180 and #112

claytonpbarrows:
- [x] resolves #179 
 - [x] resolves #180 
 - [x] resolves #181


src/base.jl
```diff
@@ -44,30 +46,44 @@ struct PowerSystem{L <: ElectricLoad,
...
     function PowerSystem(buses::Array{Bus,1},
                         generators::Array{G,1},
                         loads::Array{L,1},
                         branches::Nothing,
                         storage::Nothing,
-                        basepower::Float64; kwargs...) where {G <: Generator, L <: ElectricLoad}
+                        basepower::Float64,
+                        forecasts::F,
+                        services::V,
+                        annex::Union{Nothing,Dict{Any,Any}}; kwargs...) where {G <: Generator,
```
daniel-thom: So how you do feel about complying with our new coding standards about whitespace, particularly around commas?  :)

src/base.jl
```diff
@@ -44,30 +46,44 @@ struct PowerSystem{L <: ElectricLoad,
...
+                                                                                F <: Union{Nothing,Dict{Symbol,Array{ <: Forecast,1}}}}
         
+        sources = genclassifier(generators);
         runchecks = in(:runchecks, keys(kwargs)) ? kwargs[:runchecks] : true
         if runchecks
-                generators = checkramp(generators, minimumtimestep(loads))
+                if (!isa(forecasts,Nothing)) & (length(forecasts)>0)
+                        generators = checkramp(generators, minimumtimestep(loads))
+                        time_length = timeseriescheckload(loads)
+                        !isa(sources.renewable, Nothing) ? timeserieschecksources(sources.renewable, time_length) : true
```
daniel-thom: In the coding standards we are trying to move away from this idiom.  You could do 
if !isa(...)
    timeserieschecksources()
end

or 
!isa(...) && timeserieschecksources()

src/models/forecasts.jl
```diff
@@ -2,13 +2,44 @@ abstract type
...
+    resolution::Dates.Period        # resolution
+    initialtime::Dates.DateTime     # forecast availability time
+    data::TimeSeries.TimeArray      # TimeStamp - scalingfactor
+
+    function Deterministic(device::PowerSystemDevice,
+                            field::Symbol,
+                            resolution::Dates.Period,
+                            initialtime::Dates.DateTime,
+                            data::TimeSeries.TimeArray; kwargs...)
+        runchecks = in(:runchecks, keys(kwargs)) ? kwargs[:runchecks] : true
```
daniel-thom: This idiom commonly exists as well.  Suggest moving toward this:

runchecks = get(kwargs, :runchecks, true)

src/models/forecasts.jl
```diff
@@ -2,13 +2,44 @@ abstract type
...
+
+function Deterministic(device::PowerSystemDevice, field::Symbol, resolution::Dates.Period, initialtime::Dates.DateTime, time_steps::Int; kwargs...)
+    data = TimeSeries.TimeArray(initialtime:Dates.Hour(1):initialtime+resolution*(time_steps-1), ones(time_steps))
+    Deterministic(device, field, resolution, initialtime, data; kwargs...)
+end
+
+function Deterministic(device::PowerSystemDevice, field::Symbol, data::TimeSeries.TimeArray; kwargs...)
+    resolution = getresolution(data)
+    initialtime = TimeSeries.timestamp(data)[1]
+    time_steps = length(data)
```
daniel-thom: time_steps isn't used...

src/models/forecasts.jl
```diff
@@ -2,13 +2,44 @@ abstract type
...
+
+"""
 struct Deterministic <: Forecast
-    device::PowerSystemDevice
-    horizon::Int
-    resolution::Dates.Period
-    interval::Dates.Period
-    initialtime::Dates.DateTime
-    data::Dict{Any,Dict{Int,TimeSeries.TimeArray}}
+    device::PowerSystemDevice       # device
```
daniel-thom: Would it make more sense to store the instance of Deterministic within the PowerSystemDevice?

jd-lara: That was the previous implementation, but it didn't allow to make quick changes to the time series data for the same device

daniel-thom: Clayton and I discussed this a few minutes ago.  Maybe it should be neither.  The device and the forecast don't need to know about each other.  The operation to be executed needs to know about both, so perhaps that's where they should be tied together.

src/base.jl
jd-lara: now there is a function `isnothing` we should use in this instances also.

src/models/forecasts.jl
jd-lara: One of the key design features in PowerSystems is precisely to make the time series a defining factor in the device specification. For instance, a particular solar plant is defined in part for the availability of the resource a described by a time series or collection of time series

claytonpbarrows: The more that I think about this, the more I return to the implementation here. I tried an implementation like:
```
struct Deterministic <: Forecast
    devicename::String              # device
    fieldpath::Array{Symbol,1}      # field of device
    resolution::Dates.Period        # resolution
    initialtime::Dates.DateTime     # forecast availability time
    data::TimeSeries.TimeArray      # TimeStamp - scalingfactor
end
```
and I don't like having the `device` field as a `string` because there could be multiple devices with the same `name::String` in the System. I've returned to the position, despite the somewhat odd and backward reference of having a the `device::PowerDevice` field.

jd-lara: I don't see a problem with having the device reference instead of the string. This will avoid issues of duplicate naming.

src/base.jl
```diff
@@ -44,30 +46,44 @@ struct PowerSystem{L <: ElectricLoad,
...
+                                                                                L <: ElectricLoad,
+                                                                                V <: Union{Nothing,Array{ <: Service,1}},
+                                                                                F <: Union{Nothing,Dict{Symbol,Array{C,1} where C <: Forecast}}}
         
+        sources = genclassifier(generators);
         runchecks = in(:runchecks, keys(kwargs)) ? kwargs[:runchecks] : true
         if runchecks
-                generators = checkramp(generators, minimumtimestep(loads))
+                if !isnothing(forecasts)
+                        if length(forecasts)>0
```
kdheepak: These two expressions after the `if` can be combined into one line of code.

```
if ! ( isnothing(forecasts) || isempty(forecasts) )
```

src/base.jl
```diff
@@ -76,34 +92,43 @@ struct PowerSystem{L <: ElectricLoad,
...
-        sources = genclassifier(generators);
-        time_length = timeseriescheckload(loads)
-        !isa(sources.renewable, Nothing) ? timeserieschecksources(sources.renewable, time_length) : true
-        !isa(sources.hydro, Nothing) ? timeserieschecksources(sources.hydro, time_length) : true
-
-        runchecks = in(:runchecks, keys(kwargs)) ? kwargs[:runchecks] : true
-        if runchecks
+                
+                if !isnothing(forecasts)
+                        if length(forecasts)>0
```
kdheepak:
```
if ! ( isnothing(forecasts) || isempty(forecasts) )
```

src/base.jl
```diff
@@ -112,25 +137,37 @@ struct PowerSystem{L <: ElectricLoad,
...
+                                                                                V <: Union{Nothing,Array{ <: Service,1}},
+                                                                                F <: Union{Nothing,Dict{Symbol,Array{C,1} where C <: Forecast}}}
 
-        
+        sources = genclassifier(generators);
         runchecks = in(:runchecks, keys(kwargs)) ? kwargs[:runchecks] : true
         if runchecks
-                generators = checkramp(generators, minimumtimestep(loads))
+                if !isnothing(forecasts)
+                        if length(forecasts)>0
```
kdheepak:
```
if ! ( isnothing(forecasts) || isempty(forecasts) )
```

src/base.jl
```diff
@@ -139,29 +176,43 @@ struct PowerSystem{L <: ElectricLoad,
...
+                                                                                F <: Union{Nothing,Dict{Symbol,Array{C,1} where C <: Forecast}}}
         
+        sources = genclassifier(generators);
         runchecks = in(:runchecks, keys(kwargs)) ? kwargs[:runchecks] : true
         if runchecks
                 slackbuscheck(buses)
                 buscheck(buses)
                 pvbuscheck(buses, generators)
+                if !isnothing(forecasts)
+                        if length(forecasts)>0
```
kdheepak:
```
if ! ( isnothing(forecasts) || isempty(forecasts) )
```

PR #187, opened 2019-03-26
Fixed parsing of matpower files that don't end in newline

daniel-thom:
The function parse_matlab_string created line indices based on how many
"\n" characters were present in the file. There was an off-by-one error
whenever the last character wasn't "\n".  The code would have not worked
at all if a file used line endings of "\r" (pre Mac OS X).


daniel-thom:
Please check my changes.  If you agree then I can also submit a PR to PowerModels.

Issue #182, opened 2019-03-26
pm2ps_dict populates ramp limits as (up = 0.0, down = 0.0) 


jd-lara:
While using the data set ACTIVEg2000.m I noticed that the conversion between dicts is not using `nothing` for the ramp constraints when they are not defined.

daniel-thom:
@jd-lara Is this the same as issue #148 ?

jd-lara:
No, #148 is about another field.

claytonpbarrows:
also PM does not convert ramp limits to P.U. when parsing the files. We need to divide pm_dict ramp limits by `baseMVA` in the conversion to `ps_dict`

Issue #148, opened 2019-02-14
Missing ramp_agc is not handled in make_thermal_gen


daniel-thom:
The function make_thermal_gen assumes that "ramp_agc" is a key in the passed dictionary.  Sometimes it is not present and so the code fails with an unhandled exception.

@claytonpbarrows @jd-lara What should ramplimits be set to in this case?

jd-lara:
We can ignore that for now. This should be part later of the AGC service definition.

daniel-thom:
@claytonpbarrows Based on the comment by @jd-lara I recommend moving this to a milestone that will include implementation of the AGC service definition.

jd-lara:
I am closing this issue since it seems that the ramp_agc field is not used in matpower https://www.mail-archive.com/matpower-l@cornell.edu/msg06231.html and although the information is from 2017 there isn't anything in the current manual https://matpower.org/docs/MATPOWER-manual.pdf that contradicts the older information.

Issue #43, opened 2018-07-17
Units of PowerSystems.


jd-lara:
Should PowerSystems be stored in p.u. or in MW? For now, we have a mix, but we should be consistent and avoid issues in modeling. This is related to #10 and in particular, complicates the modeling of line flows.

claytonpbarrows:
The [standardfiles_parser.jl](https://github.com/NREL/PowerSystems.jl/blob/master/src/parsers/standardfiles_parser.jl#L12) calls `make_mixed_units()` on the `pm_dict`

jd-lara:
We have already determined that all the values in a System Struct are in p.u.

PR #189, opened 2019-04-03
Parse pu

claytonpbarrows:
This PR removes the `make_mixed_units()` conversion in the `parsestandardfiles` routine and makes the CDM parsing routines poulate per unit data. 
Addresses:
 - #182 
 - #148 
 - #43 
 - https://github.com/NREL/PowerSimulations.jl/issues/118

### DIFF
--- a/src/parsers/cdm_parser.jl
+++ b/src/parsers/cdm_parser.jl
@@ -15,7 +15,7 @@ Returns:
     as values
 
 """
-function read_csv_data(file_path::String, baseMVA = 100.0)
+function read_csv_data(file_path::String, baseMVA::Float64)
     files = readdir(file_path)
     REGEX_DEVICE_TYPE = r"(.*?)\.csv"
     REGEX_IS_FOLDER = r"^[A-Za-z]+$"
@@ -220,8 +220,8 @@ Returns:
         "BaseKV" => ..
         ...
 """
-function csv2ps_dict(file_path::String)
-    data =  read_csv_data(file_path)
+function csv2ps_dict(file_path::String, baseMVA::Float64)
+    data =  read_csv_data(file_path, baseMVA)
     ps_dict = csv2ps_dict(data)
     return ps_dict
 end
--- a/src/parsers/cdm_parser.jl
+++ b/src/parsers/cdm_parser.jl
@@ -149,7 +149,7 @@ function csv2ps_dict(data::Dict{String,Any})
          ps_dict["branch"] = nothing
     end
     if haskey(data,"dc_branch")
-        ps_dict["dcline"] =  PowerSystems.dc_branch_csv_parser(data["dc_branch"],ps_dict["bus"])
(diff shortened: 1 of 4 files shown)


src/parsers/cdm_parser.jl
```diff
@@ -317,23 +321,24 @@ function gen_csv_parser(gen_raw::DataFrames.DataFrame, Buses::Dict{Int64,Any},co
...
-                                                                        "activepowerlimits" => (min=_get_value_or_nothing(gen_raw[gen,colnames["PMin MW"]]),max=pmax),
-                                                                        "reactivepower" => _get_value_or_nothing(gen_raw[gen,colnames["MVAR Inj"]]),
-                                                                        "reactivepowerlimits" => (min=_get_value_or_nothing(gen_raw[gen,colnames["QMin MVAR"]]),max=_get_value_or_nothing(gen_raw[gen,colnames["QMax MVAR"]])),
-                                                                        "ramplimits" => (up=gen_raw[gen,colnames["Ramp Rate MW/Min"]],down=gen_raw[gen,colnames["Ramp Rate MW/Min"]]),
-                                                                        "timelimits" => (up=gen_raw[gen,colnames["Min Up Time Hr"]],down=gen_raw[gen,colnames["Min Down Time Hr"]])),
-                                            "econ" => Dict{String,Any}("capacity" => pmax,
+                                            "tech" => Dict{String,Any}("activepower" => gen_raw[gen,colnames["MW Inj"]]/baseMVA,
```
daniel-thom: Maybe this is a dumb question, but can you apply <divide by baseMVA> to the specific columns in the dataframe in one shot?

claytonpbarrows: Indeed we could. The thought had crossed my mind, but I thought this might be more obvious, and there probably isn't a real performance penalty. But if you prefer, I'll do what you suggest.

daniel-thom: Performance should be the same either way.  I would look at it from a code maintenance perspective:  if you ever have to change the calculation again, what's easier and less error-prone?

src/parsers/cdm_parser.jl
```diff
@@ -15,14 +15,16 @@ Returns:
     as values
 
 """
-function read_csv_data(file_path::String)
+function read_csv_data(file_path::String, baseMVA = 100.0)
```
jd-lara: We shouldn't have a default value here. It might be convenient for most cases but the user should be aware of what the base is

src/parsers/cdm_parser.jl
```diff
@@ -15,14 +15,16 @@ Returns:
...
 """
-function read_csv_data(file_path::String)
+function read_csv_data(file_path::String, baseMVA::Float64)
     files = readdir(file_path)
     REGEX_DEVICE_TYPE = r"(.*?)\.csv"
     REGEX_IS_FOLDER = r"^[A-Za-z]+$"
     data =Dict{String,Any}()
 
     if length(files) == 0
         @error "No test files in the folder"
```
daniel-thom: I know this is existing code, but should this error log print be changed to error() (which terminates execution)?

Issue #121, opened 2018-11-18
Improve the container for many devices. 


jd-lara:
Currently, we store PowerSystem devices in Arrays when creating a `PowerSystem`, this is efficient from the Memory POV. However, this forces us to access the data by using a list comprehension over the array or by having to index explicitly `ArrayName[ix]`. This implementation defeats the purpose of what PowerSystems does by improving the access and use of Power System data. This is evident when taking a deeper look at how things are done in PowerSimulations to manipulate the data from PowerSystems. 

We need to test several container options that we can index with a better format in order to avoid this practice. I am thinking of `AxisArrays`, `AbstractArrays` or eventually `Dicts` but we need to make performances tests for this.

ccoffrin:
Can you give an example of one of these list comprehension operations?  I presume it is used to build a list of a specific field of a component?

PS: `Dict` FTW!

jd-lara:
for instance this to get the name sets 

`gen_names = [g.name for g in ArrayofGens if g.available]`

ccoffrin:
Can you give a little more context on why this is slow?  Other than allocating arrays over and over it's not clear to me how it can be made faster.

If array allocation is the time sink, then something like this could work around it,
```
for g in ArrayofGens
   if g.available
       <various computations related to g>
   end
end
```

jd-lara:
The problem is when combined with JuMP and we want to access data then it leads to this kind of code or even more complicated versions. 

```Julia 
for (ix, devices) in enumerate(ArrayofDevices)
     @contraint(m, Var[ArrayofDevices[ix].name] >= ArrayofDevices[ix].data)
end
```

ccoffrin:
What about something like this,
```
for device in ArrayofDevices
   if device.available
      @constraint(m, Var[device.name] >= device.data)
   end
end
```
I think that would be pretty fast.

jd-lara:
That will work for many of the cases, I will implement that as I refactor to anonymous variables soon. Thanks @ccoffrin

jd-lara:
@ccoffrin I am leaning towards using dicts similar to PowerModels but store the device using PowerSystems types. This will also make easier the integration and DataWrangling.

jd-lara:
@ccoffrin Exactly. Still researching if using strings for indexing is the best alternative. I made some tests with AxesArrays, the features are cool but the speed is not great, and it complicates interoperability with other languages.

claytonpbarrows:
[...]

PR #190, opened 2019-04-08
Refactor PowerSystem constructors

daniel-thom:
This will now fix issue #121


src/base.jl
```diff
@@ -9,17 +9,21 @@ overall system parameters.
...
 ```
 
 # Arguments
 
-* `buses`::Array{Bus,1} : an array of buses
-* `generators`::Array{Generator,1} : an array of generators of (possibly) different types
-* `loads`::Array{ElectricLoad,1} : an array of load specifications that includes timing of the loads
-* `branches`::Array{Branch,1} : an array of branches; may be `nothing`
-* `storage`::Array{Storage,1} : an array of storage devices; may be `nothing`
+* `buses`::Buses : an array of buses
```
jd-lara: Semantics Should be `BusesArray`

src/base.jl
```diff
@@ -30,170 +34,120 @@ PowerSystem(file, ts_folder; kwargs...)
...
-    """
-    docstrings for buses field
-    """
-    buses::Array{Bus,1}
+    buses::Buses
     generators::GenClasses
-    loads::Array{L,1}
-    branches::B
-    storage::S
+    loads::ElectricLoads
```
jd-lara: Same here `ElectricLoadsArray`

src/validation/generator.jl
```diff
@@ -0,0 +1,25 @@
...
+    return is_valid
+end
+
+"""Validates the contents of a Generator."""
+function validate(generator::Generator)::Bool
+    is_valid = true
+
+    @debug "Generator validation" generator.name is_valid
+    return is_valid
+end
```
kdheepak: Is this incomplete? This function just returns `true`.

daniel-thom: Intentionally incomplete.  I wanted to get agreement on the design before including details.  This is roughly how it will look.  Another missing item is a config file containing default values, mins, and maxes.

daniel-thom:
This PR has all the changes we discussed this morning.  There is one area of feedback I'd like to get from everyone, but you will have to use this code first.  The ConcreteSystem dictionary members called components and components_by_type in are keyed by Type (instead of Symbol).  The REPL usage differs on two factors:
- using PowerSystems or import PowerSystems:  causes different names for Types but Symbols are always the same
- tab auto-completion works for Types but not Symbols

jd-lara:
> This PR has all the changes we discussed this morning. There is one area of feedback I'd like to get from everyone, but you will have to use this code first. The ConcreteSystem dictionary members called components and components_by_type in are keyed by Type (instead of Symbol). The REPL usage differs on two factors:
> 
> * using PowerSystems or import PowerSystems:  causes different names for Types but Symbols are always the same
> * tab auto-completion works for Types but not Symbols

If we are going to have a single field in the struct for the separated components, then it would be better to call the current `components` fields `data` and call `components_by_type` just components. Especially given that the `components` field will actually have all the data including other things that are not components.

src/base.jl
```diff
@@ -30,170 +33,190 @@ PowerSystem(file, ts_folder; kwargs...)
...
+    return PowerSystem(buses, generators, loads, branches, storage, ps_dict["baseMVA"];
+                       kwargs...);
+end
+
+"""A PowerSystem struct that stores all devices in arrays with concrete types.
+This is a temporary implementation that will allow consumers of PowerSystems to test the
+functionality before it is finalized.
+"""
+struct PowerSystemConcrete
+    devices::Dict{DataType, Vector{<: PowerSystemDevice}}
```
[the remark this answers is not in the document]
daniel-thom: You already get this for free with
sys.components_by_type[Device]
sys.components_by_type[Forecast]
sys.components_by_type[Service]

Secondly, Branch is a now a subtype of Device, so what you wrote is contradictory.  Do you disagree with this type hierarchy?

daniel-thom:
> > This PR has all the changes we discussed this morning. There is one area of feedback I'd like to get from everyone, but you will have to use this code first. The ConcreteSystem dictionary members called components and components_by_type in are keyed by Type (instead of Symbol). The REPL usage differs on two factors:
> > 
> > * using PowerSystems or import PowerSystems:  causes different names for Types but Symbols are always the same
> > * tab auto-completion works for Types but not Symbols
> 
> If we are going to have a single field in the struct for the separated components, then it would be better to call the current `components` fields `data` and call `components_by_type` just components. Especially given that the `components` field will actually have all the data including other things that are not components.

daniel-thom:
data and components sounds good to me.

src/base.jl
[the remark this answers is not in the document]
jd-lara: Yes, this is an outdated thing. I don't know why Github posted it when I added my later comment.

daniel-thom:
What is not (or should not be) a subtype of Component?

src/validation/generator.jl
kdheepak: Add a comment that says this will be implemented later.

src/base.jl
```diff
@@ -1,199 +1,321 @@
...
+"""Constructs System with Generators but no branches or storage."""
+function System(buses::Vector{<: Bus},
+                generators::Vector{<: Generator},
+                loads::Vector{<: ElectricLoad},
+                basepower::Float64; kwargs...)
+    return System(buses, generators, loads, nothing, nothing, basepower; kwargs...)
+end
+
+"""Constructs System with Generators but no storage."""
+function System(buses::Vector{Bus},
```
kdheepak: Should this be `Vector{<:Bus}`?

daniel-thom: Bus is a concrete type.

src/validation/generator.jl
daniel-thom: There is a comment in the System constructor.  Does that work for you?

PR #194, opened 2019-04-09
Add `branch` as keyword argument to download function

kdheepak:
- make `branch` keyword only arguments
- re-indent code to be consistent with Julia style guide


kdheepak:
This is ready to be merged.

PR #202, opened 2019-04-14
Fix data dict creation with parametric types

### LINKED ISSUE #201: ConcreteSystem with parametric subtypes
The [ConcreteSystem](https://github.com/NREL/PowerSystems.jl/blob/dev/src/base.jl#L169) breaks when the system includes parametric subtypes like [`Deterministic`](https://github.com/NREL/PowerSystems.jl/blob/dev/src/models/forecasts.jl#L8)

@daniel-thom, @jd-lara, @kdheepak I'm not sure how to proceed.

kdheepak:
Resolves #201 

This creates the keys of the dictionary on the fly. 
We don't know what all possible keys can be, so I think this is what we should do.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -207,7 +207,8 @@ function ConcreteSystem(sys::System)
         @debug "data: $(string(key)): count=$(string(length(value)))"
     end
 
-    components = _get_components_by_type(Component, data)
+    # components = _get_components_by_type(Component, data)
+    components = Dict{DataType, Any}()
     return ConcreteSystem(data, components, sys.basepower)
 end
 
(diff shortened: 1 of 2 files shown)


src/base.jl
```diff
@@ -200,7 +207,8 @@ function ConcreteSystem(sys::System)
         @debug "data: $(string(key)): count=$(string(length(value)))"
     end
 
-    components = _get_components_by_type(Component, data)
+    # components = _get_components_by_type(Component, data)
+    components = Dict{DataType, Any}()
```
kdheepak: I think we should have a further discussions of this components construction using recursion. If we have parameterized types in PowerSystems this may pose a challenge. On a tangent we should also discuss how parametrized types should be serialized. 

I've commented this out in this PR because I didn't want to make sweeping changes across the package without having a discussion.

claytonpbarrows:
@kdheepak can we close this?

kdheepak:
Yes. We are not storing the data this way anymore.

Issue #185, opened 2019-03-26
Many check_.* functions should be marked with ! because they mutate objects passed in


daniel-thom:
Many functions named check_.* mutate the objects passed in but are not marked with an exclamation point, which is the Julia standard.  This is misleading to readers.

Functions to check and potentially change:
git grep -E "function _?check\w+\("

daniel-thom:
This could be invalidated if we validate all values with our new proposal.

Issue #177, opened 2019-03-22
Setting validate=false in parse_file results in several failures.


daniel-thom:
Code that runs when validate=true adds and modifies data that end up being required by other code.  If this parameter remains optional then we should be able to execute correctly with either option.

daniel-thom:
Here is one example of a failure:
- If validate=true is passed to parse_file then check_thermal_limits() in pm2ps_parser.jl will be called.  If any of the passed branches don't have a key called "rate_a" then it adds one.
- pm2ps_dict() calls read_branch() which calls make_lines().  That function tries to read "rate_a" from the data dictionary.  A missing value causes a runtime exception.

Issue #193, opened 2019-04-09
Remove ramp check 


jd-lara:
See https://github.com/NREL/PowerSimulations.jl/issues/121

Issue #168, opened 2019-03-19
Remove ReadMe.md in `src/models`


kdheepak:

jd-lara:
closed via b57cd5c

PR #206, opened 2019-04-17
v0.4.0

### LINKED ISSUE #290: add SIL calculation to data validation
calculation in PM is here: 
https://github.com/lanl-ansi/PowerModels.jl/blob/357c29779499c215dc99654bde35200df7cba479/src/core/data.jl#L1054

claytonpbarrows:
- refactors ~~the storage of time series information~~ all the structs and the creation process. 
~~adds the `ConcreteSystem` representation~~ `ConcreteSystem` was made into `System`

As we approach the final stages to merge dev. This is the list of capabilities missing. 

- [x] Validation checks (includes selecting which are going to be kept from PM). #177 #265 #185 #294 (Partially completed, Branch Data needs to resolve #290 
- [x] Build System Dynamically. 
- [x] Add to_index methods in NetworkArrays #285 
- [x] Remove Python Dependency
- [ ] Increase Coverage Rates #125 
- [x] Address failing tests on Julia v1.2
- [x] Structs reviews #259 #288 
- [ ] Add Data Validation to unit test #280 
- [x] Add SIL to validations #290 https://neos-guide.org/sites/default/files/line_flow_approximation.pdf
- [x] Use UUIDs only for referencing buses #303 
- [x] Keep or bypass PM Validation code. 
- [x] Fix deprecations in DataFrames use
- ~~Use Project and TOML when building structs. (aspirational)~~ This is now in IS. 
- [x] Address @kdheepak about the poweflow function and #310 
- [x] Reorganize data files. @jd-lara (Waiting for IS to be registered) 
- [x] Fix print statements to return summary #325 
~~Add accessor functions for tech field @daniel-thom~~ Deemed unreasonable 
- [ ] Fix problem with reserves #334 (parsing problem to be addressed later)

Remember to merge https://github.com/NREL/PowerSystemsTestData/tree/improve-timeseries and change the target branch in `build.jl` before merging to master.


src/models/forecasts.jl
```diff
@@ -1,14 +1,61 @@
...
 end
 
+function Deterministic(component::Component, label::String, data::TimeSeries.TimeArray; kwargs...)
+    resolution = getresolution(data)
+    initialtime = TimeSeries.timestamp(data)[1]
+    time_steps = length(data)
+    Deterministic(component, label, resolution, initialtime, data; kwargs...)
+end
+
+#=
```
daniel-thom: Do we need to keep this around?

src/base.jl
```diff
@@ -156,169 +157,105 @@ This is a temporary implementation that will allow consumers of PowerSystems to
...
-"""Returns an array of components from the System. T must be a concrete type.
-
-# Example
-```julia
-devices = PowerSystems.get_components(ThermalDispatch, system)
-```
-"""
-function get_components(::Type{T}, sys::ConcreteSystem)::Vector{T} where {T <: Component}
+"""Adds a component to the system."""
+function add_component(sys::ConcreteSystem, component::T) where T <: Component
```
daniel-thom: I should have called this add_component!  Do you want to make the change?  Or I can do it in a subsequent commit.

src/models/generation/thermal_generation.jl
```diff
@@ -39,18 +39,16 @@ Data Structure for thermal generation technologies subjecto to seasonality const
     Examples
 
 """
-struct ThermalGenSeason <: ThermalGen
+struct ThermalGenSeason <: ThermalGen # TODO: Do we need this? Removal of scalingfactor makes it a duplicate of above
```
daniel-thom: Removing makes sense to me. It shouldn't cause any harm unless someone else is depending on it.

src/base.jl
claytonpbarrows: Done

src/models/generation/thermal_generation.jl
claytonpbarrows: Done

src/parsers/forecast_parser.jl
```diff
@@ -56,171 +56,55 @@ end
...
-function make_forecast_dict(time_series::Dict{String,Any},
-                            resolution::Dates.Period, horizon::Int,
-                            Devices::Array{ElectricLoad,1})
-    forecast = Dict{String,Any}()
-    for device in Devices
-        if haskey(time_series,"load")
-            if device.bus.name in  map(String, names(time_series["load"]))
-                df = time_series["load"][[:DateTime,Symbol(device.bus.name)]]
-                forecast[device.name] = make_device_forecast(device, df, resolution, horizon)
+function make_forecast_array(sys::Union{System,Array{Component,1}},ts_dict::Dict)
```
daniel-thom: This whole code path is a bit convoluted because you are trying to deal with either an array or a System.  It could be simpler if you had one path that created this array during parsing and then a second that creates it after ConcreteSystem exists.  You wouldn't have to do all of the introspection to find devices within System.

src/parsers/forecast_parser.jl
```diff
@@ -56,171 +56,55 @@ end
...
-    Forecasts = Array{Forecast}(undef, 0)
-    for (device_key,device_dict) in dict
-                push!(Forecasts,Deterministic(device_dict["device"],device_dict["horizon"],
-                                device_dict["resolution"],device_dict["interval"],
-                                device_dict["initialtime"],
-                                device_dict["data"]
-                                ))
-    end
-    return Forecasts
+function add_forecast!(sys::System,fc::Pair{Symbol,Array{Forecast,1}})
```
daniel-thom: I have a preference for putting functions that modify a struct in the same file as that struct.  Thoughts?  Also, you have a typo in the docstring - "modeified".

src/parsers/standardfiles_parser.jl
```diff
@@ -24,12 +24,13 @@ end
...
-
+    data = parsestandardfiles(file)
+    sys = System(data)
+    
     ts_data = read_data_files(ts_folder; kwargs...)
 
-    # assign_ts_data is in forecast_parser.jl
-    data = assign_ts_data(data, ts_data)
+    forecast = make_forecast_array(sys,ts_data)
+    pushforecast(sys,:default=>forecast)
```
daniel-thom: Is this a bug?  add_forecast!

src/parsers/forecast_parser.jl
claytonpbarrows: This is what I was getting at with my comment on gitter. I think the correct approach is to use the ConcreteSystem where we can. This might allow us to get rid of the `_access`, `get_device`, `_retrieve` methods too.

jd-lara:
We shouldn't merge this until is is compatible with Power Simulations

src/base.jl
```diff
@@ -41,13 +41,15 @@ struct System <: PowerSystemType
...
     storage::Union{Nothing, Vector{<:Storage}}
     basepower::Float64 # [MVA]
-    time_periods::Int64
+    forecasts::Union{Nothing, SystemForecasts}
+    services::Union{Nothing, Vector{ <: Service}}
+    annex::Union{Nothing,Dict{Any,Any}}
 
     function System(buses, generators, loads, branches, storage_devices, basepower,
-                    time_periods; kwargs...)
+                    forecasts, services, annex; kwargs...)
```
kdheepak: I didn't see this earlier. Are we sure they will be no issue when the inner constructor takes a `kwargs...`?

kdheepak: Also, in the inner constructor we are not propagating `kwargs...` any further. We can use `validate=false` here instead, but there are tradeoffs. We will have to use `pop!` instead of `get` for using values from `kwargs` when we don't want to it propagate further, and still use `get` when we do. This requires a clearer understanding of all dispatch methods. On the other hand we'll catch errors if someone types in an incorrect keyword. At the moment if the user makes a typo, the functions will silently choose a default value that might be related what they intend to do.

src/base.jl
```diff
@@ -64,7 +66,10 @@ function System(buses::Vector{Bus},
                 loads::Vector{<:ElectricLoad},
                 branches::Union{Nothing, Vector{<:Branch}},
                 storage::Union{Nothing, Vector{<:Storage}},
-                basepower::Float64; kwargs...)
+                basepower::Float64,
+                forecasts::Union{Nothing, SystemForecasts},
+                services::Union{Nothing, Vector{ <: Service}},
+                annex::Union{Nothing,Dict{Any,Any}}; kwargs...)
     runchecks = in(:runchecks, keys(kwargs)) ? kwargs[:runchecks] : true
```
kdheepak: Prefer this idiom: `get(kwargs, :runchecks, true)` instead?

kdheepak:
I think we can get rid of these functions

https://github.com/NREL/PowerSystems.jl/blob/c824b97c739a6ebf651624a54dfdb4035a89e462/src/base.jl#L96-L120

user either provides all the arguments or uses the keyword only argument function.

src/base.jl
daniel-thom: I vote that we defer conversation about refactoring System constructors.  Once we integrate workflows with ConcreteSystem in PowerSimulations and other consumers, we need to look at changing how users construct ConcreteSystem.  They should not have to construct System first.  In summary, ConcreteSystem will ultimately become System and will need new constructors.

src/base.jl
jd-lara: @kdheepak is right, inner constructors should take only specific keyword argument. Also, I am not sure that the checks should be optional. We should move some of them to PowerSimulations but not all of them are needed here.

jd-lara:
We are missing several features here before we can merge into master. We are missing #193, #168, #199 to complete the milestone.

daniel-thom:
I'm not sure we need the JSON serialization complete for this milestone.  This set of changes provides value that can be consumed.  The JSON work might not be complete until next week.

jd-lara:
I don't think I will have completed the integration of these changes into PowerSimulations until next week.

Issue #203, opened 2019-04-15
Add support for UUIDs in all PowerSystems structs.


daniel-thom:
In order to be able to allow users to delete components or forecasts from a PowerSystem each object will need to store a UUID.

When an object is first created it should call UUIDs.uuid4(). This UUID should be stored in the object as an internal attribute.  The user should be able to remove an object with something like:

delete!(system, forecast)

Our code can then retrieve the UUID and delete it from the underlying data structure.

We have three proposals to solve this:
1. Add a UUID field to every struct.
2. Create a new PowerSystemInternal struct that contains a UUID.  If we ever add new fields in the future, it will be much easier.
3. Implement a macro that automatically adds fields to a struct.

We should defer this work until after the JSON serialization work (PR #199) is complete.

daniel-thom:
Here is another reason to implement UUIDs:  better support serialization of a PowerSystem.

PowerSystems sometimes stores references to components in other data structures.  For example, ProportionalReserve has contributingdevices::Array{Device} that gets populated with references to actual devices.  When we serialize ProportionalReserve to JSON the information will be duplicated in the JSON string.  If a user were to change values in the main instance of the device but not in the secondary places then deserialization will not produce an identical System.

Solution:  Serialize out the UUIDs for these secondary references.

This would solve another problem.  Deserialization of abstract types is difficult with the current JSON libraries because there is no standard way of encoding the type of each object.

jd-lara:
We would then need to implement some of these changes with inner constructors in the device types such that the users can't create devices with custom UUIDs. In the proposal number 2, what do you mean with add new fields in the future?

daniel-thom:
In order to support deserialization we have to support a constructor that accepts a UUID (or PowerSystemInternal).

Adding this struct to every existing struct is very tedious work.  If we ever need to add a field in the future then we could add it to this one struct - one change in one file.

jd-lara:
Do we want this to be resolved for v0.4.0?

daniel-thom:
I'm working on it right now.  It's needed for (1) allowing removal of components and (2) JSON serialization.  It should not block 0.4.0, but can go in if it's ready.

jd-lara:
@daniel-thom thanks, if it is ready on time then we can add it.

daniel-thom:
Fixed by PR #209

PR #209, opened 2019-04-23
UUIDs and JSON Serialization

daniel-thom:
This PR adds UUIDs to all PowerSystems types and allows serialization and deserialization wtih JSON.  This supersedes the previous JSON PR.

There were quite a few special cases to handle with JSON serialization.  If you have concerns about future sustainability, definitely raise them.

In a subsequent change we could replace all JSON calls with JSON2.

This will fix issue #203.


src/models/serialization.jl
```diff
@@ -0,0 +1,94 @@
...
+function JSON2.write(resolution::Dates.Period)
+    return JSON2.write(encode_for_json(resolution))
+end
+
+function JSON2.write(io::IO, resolution::Dates.Period)
+    return JSON2.write(io, encode_for_json(resolution))
+end
+
+# TODO: should this actually be milliseconds? What is the lowest resolution we use?
+RESOLUTION_UNITS_FUNC = Dates.Second
```
daniel-thom: @claytonpbarrows  I need guidance on this point.  Are seconds sufficient?

claytonpbarrows: For our primary use case, seconds is sufficient. However, I'm trying to enable PowerDynamics.jl which will require milliseconds, so I'm inclined to stick with milliseconds.

jd-lara: Why would we have time series in the millisecond time steps? Even if we have Dynamic Simulations enabled.

jd-lara:
it might be necessary to rebase this branch onto dev, given the changes in the Project.TOML file

src/models/serialization.jl
daniel-thom: On second thought I should be able to serialize out a number plus a unit and then convert it back later.

src/models/branches/lines.jl
```diff
@@ -12,8 +12,15 @@ struct Line <: Branch
     b::From_To_Float # [pu]
     rate::Float64
     anglelimits::Min_Max #Degrees
+    internal::PowerSystemInternal
 end
 
+function Line(name, available, connectionpoints, r, x, b, rate, anglelimits::Min_Max)
```
kdheepak: Why is the last argument here alone typed?

src/models/products/reserves.jl
```diff
@@ -54,19 +59,23 @@ requirement - the required quantity of the product
...
-    function StaticReserve(name::String,
-                                contributingdevices::Array{Q},
-                                timeframe::Float64,
-                                generators::Array{G}) where {Q <: Device, G <: TechThermal}
+function StaticReserve(
+                       name,
+                       contributingdevices,
+                       timeframe,
+                       generators,
+                      ) where {Q <: Device, G <: TechThermal}
```
kdheepak: If the types from the function argument are removed, shouldn't the `where {Q<:Device, G<:TechThermal}` also be removed?

src/models/branches/lines.jl
jd-lara: Looks like a bug

src/models/products/reserves.jl
jd-lara: We should open an issue to check that all the functions are typed

src/models/serialization.jl
daniel-thom: Fixed.

src/models/branches/lines.jl
daniel-thom: Oversights. Thanks for catching them.  Fixed.

src/models/products/reserves.jl
daniel-thom: Fixed.  If someone thinks that these should actually be typed, speak up.

src/internal.jl
```diff
@@ -0,0 +1,20 @@
...
+"""Internal storage common to PowerSystems types."""
+struct PowerSystemInternal
+    uuid::Base.UUID
+end
+
+"""Creates PowerSystemInternal with a UUID."""
+PowerSystemInternal() = PowerSystemInternal(UUIDs.uuid4())
+
+"""Gets the UUID for any PowerSystemType."""
+function get_uuid(obj::T)::Base.UUID where T <: PowerSystemType
```
kdheepak: This can be `get_uuid(obj::PowerSystemType)` and does not need the `where T <: PowerSystemType` as we discussed.

src/internal.jl
```diff
@@ -0,0 +1,20 @@
...
+    return obj.internal.uuid
+end
+
+"""Gets the UUIDs for a vector of any PowerSystemType."""
+function get_uuids(objs::Vector{T})::Vector{Base.UUID} where T <: PowerSystemType
```
kdheepak: I have a minor preference for renaming this to `get_uuid`. 

I think I also prefer just removing this function. `get_uuid.(objs, uuids)` will call the previous method and return `Vector{Base.UUID}` appropriately because of the broadcasting operator.

src/models/generation/econ_common.jl
```diff
@@ -44,3 +55,23 @@ EconThermal(;   capacity = 0.0,
...
             annualcapacityfactor = nothing
         ) = EconThermal(capacity, variablecost, fixedcost, startupcost, shutdncost, annualcapacityfactor)
+
+"""Enables deserialization of EconThermal. The default implementation can't figure out the
+variablecost Union.
+"""
+function JSON2.read(io::IO, ::Type{T}) where {T <: EconThermal}
+    data = JSON2.read(io)
+    if data.variablecost isa Array
+        variablecost = Vector{Tuple{Float64, Float64}}()
```
kdheepak: I have a preference for using `Tuple{Float64, Float64}[]` instead of `Vector{Tuple{Float64, Float64}}()`.

src/models/serialization.jl
```diff
@@ -0,0 +1,92 @@
+
+"""Serializes a PowerSystemType to a JSON file."""
+function to_json(obj::T, filename::String) where {T <: PowerSystemType}
+    return open(filename, "w") do io
+        return to_json(io, obj)
+    end
+
+    @info "Serialized $T to $filename"
```
kdheepak: Shouldn't this be `@debug`?

test/test_internal.jl
```diff
@@ -0,0 +1,58 @@
...
+"""Recursively validates that the object and fields have UUIDs."""
+function validate_uuids(obj::T) where T
+    if !(obj isa PowerSystemType)
+        return true
+    end
+
+    result = true
+    if !(PowerSystems.get_uuid(obj) isa Base.UUID)
+        result = false
+        @error "object does not have a UUID" obj
```
kdheepak: This `@error` macro does not throw an error, just logs an error.

kdheepak:
Looks good to me! One other thing I was thinking was whether we should be using `getfield`, `fieldnames` etc versus `getproperty`, `propertynames`. Using `getproperty` and `propertynames` might be preferred, since it resolves to `getfield` and `fieldnames` anyway and will allow more flexibility.

src/models/serialization.jl
daniel-thom: I intended to always log it.  It won't happen very often.

test/test_internal.jl
daniel-thom: My intention was to log every violation instead of failing on the first one.

src/internal.jl
daniel-thom: Fixed

src/internal.jl
daniel-thom: Thanks!  I forgot about the broadcast operator.  Deleted.

src/models/generation/econ_common.jl
daniel-thom: Left as-is per our discussion.

src/models/serialization.jl
daniel-thom: I'm leaving it here unless you still disagree.

test/test_internal.jl
daniel-thom: Ditto.

PR #213, opened 2019-04-26
Annex nothing cases

### LINKED ISSUE #212: Missing methods in ConcreteSystem
When building the `ConcreteSystem` the method fails when fields have `nothing`. 

```Julia
sys5 = PSY.System(nodes14, thermal_generators5, loads5, branches5, nothing,  100.0, forecasts5, nothing, nothing);
sys14 = PSY.System(nodes14, thermal_generators14, loads14, branches14, nothing,  100.0, forecasts14, nothing, nothing);
c_sys5 = PSY.ConcreteSystem(sys5)
c_sys14 = PSY.ConcreteSystem(sys14)
```

Error stack trace returns 

```Julia
MethodError: no method matching get(::Nothing, ::Symbol, ::Nothing)
Closest candidates are:
  get(!Matched::Base.TTY, ::Symbol, ::Any) at stream.jl:482
  get(!Matched::REPL.Terminals.TTYTerminal, ::Any, ::Any) at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/REPL/src/Terminals.jl:176
  get(!Matched::IdDict{K,V}, ::Any, ::Any) where {K, V} at abstractdict.jl:594
  ...

Stacktrace:
 [1] ConcreteSystem(::System) at /Users/jdlara/.julia/packages/PowerSystems/VxcFx/src/base.jl:181
 [2] top-level scope at In[18]:3
```

jd-lara:
Fixes #212

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -178,9 +178,9 @@ function ConcreteSystem(sys::System)
         end
     end
 
-    !isnothing(sys.annex) ? loadZones = get(sys.annex, :LoadZones, nothing) : loadZones = nothing
-    if !isnothing(loadZones)
-        for lz in loadZones
+    load_zones = isnothing(sys.annex) ? nothing : get(sys.annex, :LoadZones, nothing)
+    if !isnothing(load_zones)
+        for lz in load_zones
             add_component!(concrete_sys, lz)
         end
     end
--- a/src/base.jl
+++ b/src/base.jl
@@ -234,9 +234,9 @@ function get_components(::Type{T}, sys::ConcreteSystem) where {T <: Component}
     if isconcretetype(T)
         components = get(sys.components, T, nothing)
         if isnothing(components)
-            return Iterators.flatten([], 0)
+            return Iterators.flatten([])
         else
-            return Iterators.flatten([components], length(components))
+            return Iterators.flatten([components])
         end
     else
         types = [x for x in get_all_concrete_subtypes(T) if haskey(sys.components, x)]
--- a/src/base.jl
+++ b/src/base.jl
@@ -178,7 +178,7 @@ function ConcreteSystem(sys::System)
         end
     end
 
-    loadZones = get(sys.annex, :LoadZones, nothing)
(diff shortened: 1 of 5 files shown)


src/base.jl
```diff
@@ -178,7 +178,7 @@ function ConcreteSystem(sys::System)
         end
     end
 
-    loadZones = get(sys.annex, :LoadZones, nothing)
+    !isnothing(sys.annex) ? loadZones = get(sys.annex, :LoadZones, nothing) : loadZones = nothing
```
daniel-thom: How about this instead?  (also noticed that I changed loadZones to load_zones)

load_zones = isnothing(sys.annex) ? nothing : get(sys.annex, :LoadZones, nothing)

jd-lara: The change looks good to me. I haven't seen the change in load zone naming.

daniel-thom:
This reveals the fact that we don't have any tests for this condition.  Would you mind adding invocations of ConcreteSystem() to all constructions of System in test/powersystemconstructors?

src/base.jl
daniel-thom: I was just referring to this local variable.  Our convention is snake_case.

jd-lara:
> This reveals the fact that we don't have any tests for this condition. Would you mind adding invocations of ConcreteSystem() to all constructions of System in test/powersystemconstructors?

Ok

daniel-thom:
Also, I was working on fixing this before you pushed.  I noticed that you can't run the powersystemconstructors test by itself.  It needs to have an <include("../data/data_5bus_pu.jl")> at the top.  Otherwise, nodes5 is undefined.  Might want to make that change also.

jd-lara:
> This reveals the fact that we don't have any tests for this condition. Would you mind adding invocations of ConcreteSystem() to all constructions of System in test/powersystemconstructors?

I just added the missing tests.

daniel-thom:
Everything looks good except for my comment about <include("../data/data_5bus_pu.jl")> in the test.

daniel-thom:
Why don't I see an "approve" button?

PR #221, opened 2019-05-01
update output of get_components

### LINKED ISSUE #219: get_components inconsistent return type
There is some inconsistent behavior in the `get_components` function for a concrete system. 

```Julia
include(joinpath(base_dir,"data/data_5bus_pu.jl"));
sys5 = PSY.System(nodes5, thermal_generators5, loads5, branches5, nothing,  100.0, forecasts5, nothing, nothing);

#Concrete System type
devices = PSY.get_components(PSY.ThermalDispatch, c_sys5);
typeof(devices)
> Base.Iterators.Take{Array{ThermalDispatch,1}} 

#Abstract Type 
devices = PSY.get_components(PSY.ThermalGen, c_sys5);
typeof(devices)
> Base.Iterators.Flatten{Base.Generator{Array{DataType,1},getfield(PowerSystems, Symbol("##84#87")){ConcreteSystem}}}
```

This is not consistent with the return types in `get_components`

jd-lara:
closes #219

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -228,8 +228,7 @@ end
 """
 const ComponentIterator{T} = Base.Iterators.Flatten{Array{Array{T,1},1}} where {T <: Component}
 
-
-"""Returns an ComponentIterator{T} from the System. T can be concrete or abstract.
+"""Returns a ComponentIterator{T} from the System. T can be concrete or abstract.
 
 # Examples
 ```julia
--- a/src/base.jl
+++ b/src/base.jl
@@ -239,9 +239,14 @@ function get_components(::Type{T}, sys::ConcreteSystem) where {T <: Component}
             return Iterators.take(components, length(components))
         end
     else
-        types = [x for x in get_all_concrete_subtypes(T) if haskey(sys.components, x)]
-        abstract_type_vector = Vector{T}(reduce(vcat, sys.components[x] for x in types))
-        return Iterators.take(abstract_type_vector, length(abstract_type_vector))
+        types = [x for x in get_all_concrete_subtypes(T) if haskey(sys.components, x)]  
+        components_arrays = [sys.components[x] for x in types]
+        if  !isempty(components_arrays) 
+            abstract_type_vector = Vector{T}(reduce(vcat, components_arrays))
+            return Iterators.take(abstract_type_vector, length(abstract_type_vector))
+        else
+            return Iterators.take(Vector{T}(), 0)
+        end
     end
 end
 
--- a/src/base.jl
+++ b/src/base.jl
@@ -221,6 +221,13 @@ end
 # need each PowerSystemType to store a UUID.
 # GitHub issue #203
 
+# Const Definitions
+
+
+"""Defines the Iterator to contain variables of type Component.
+"""
+const ComponentIterator{T} = Base.Iterators.Flatten{Array{Array{T,1},1}} where {T <: Component}
+
 
 """Returns an iterable of components from the System. T can be concrete or abstract.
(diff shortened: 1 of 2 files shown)


src/base.jl
```diff
@@ -234,13 +234,14 @@ function get_components(::Type{T}, sys::ConcreteSystem) where {T <: Component}
...
-            return Iterators.flatten([[]])
+            return Iterators.take(Vector{T}(), 0)
         else
-            return Iterators.flatten([components])
+            return Iterators.take(components, length(components))
         end
     else
         types = [x for x in get_all_concrete_subtypes(T) if haskey(sys.components, x)]
-        return Iterators.flatten(sys.components[x] for x in types)
+        abstract_type_vector = Vector{T}(reduce(vcat, sys.components[x] for x in types))
```
daniel-thom: Doesn't this allocate a new array? (which is what we were trying to avoid?

jd-lara: It does, I guess it doesn't really help to fix the issue.

jd-lara: removed with f31e366

src/base.jl
```diff
@@ -234,13 +234,13 @@ function get_components(::Type{T}, sys::ConcreteSystem) where {T <: Component}
     if isconcretetype(T)
         components = get(sys.components, T, nothing)
         if isnothing(components)
-            return Iterators.flatten([[]])
+            return Iterators.flatten([Vector{T}()])
```
daniel-thom: For readability it would be good to set the type of this function as ComponentIterator{T} - and describe it explicitly in the docstring.  I was very vague the first time around because the return type wasn't consistent.

jd-lara: addressed in 197f19e

daniel-thom:
I guess I wasn't clear on my last comment.  I was suggesting that you set the return type of the function get_components like this:

```
function get_components(::Type{T}, sys::ConcreteSystem)::ComponentIterator{T} where {T <: Component}
```

And update the function docstring.

src/base.jl
```diff
@@ -221,26 +221,32 @@ end
...
 generators = PowerSystems.get_components(Generator, system)
 ```
 """
-function get_components(::Type{T}, sys::ConcreteSystem) where {T <: Component}
+function get_components(::Type{T}, sys::ConcreteSystem)::ComponentIterator{T} where {T <: Component}
     if isconcretetype(T)
         components = get(sys.components, T, nothing)
         if isnothing(components)
-            return Iterators.flatten([[]])
+            return ComponentIterator{T}(Vector{T}())
```
daniel-thom: I just noticed that you've changed this section again.  There are three cases where you create a ComponentIterator.  In the first you explicitly create a Vector{T} instead of [].  In the next two you create the object from [...]. These should be consistent.  If using [] works, then you can just delete Vector.  Also, your type definition uses Array{T, 1} instead of Vector.  These should also be consistent.  I think we should always use Vector{T} instead of Array{T, 1}.

jd-lara: If I don't specify the type inside of the ComponentIterator then it returns a type `Iterators.flatten[Array{Any,1}]` which made the tests fail. This syntax will always return an iterator with the correct Array Type parametrization even if it needs to be empty instead of an empty Array parametrized with Any. 

I can change the ComponentIterator to use Vector that is not an issue.

daniel-thom: Ok, makes sense.  In cases 2 and 3 does it still do the right thing if the vectors are empty?

jd-lara: Yes, the return is fine

PR #232, opened 2019-05-10
Elevate system

claytonpbarrows:
This PR:
 - renames `System` to `_System`
 - renames `ConcreteSystem` to `System` and routes all construction of `System` through the construction of `_System`

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -243,7 +243,7 @@ function System(buses::Vector{Bus},
                 basepower::Float64; kwargs...)
 
     
-    return = System(buses, generators, loads, nothing, nothing, basepower, nothing, nothing, nothing; kwargs...)
+    return System(buses, generators, loads, nothing, nothing, basepower, nothing, nothing, nothing; kwargs...)
 end
 
 """Constructs System with default values."""
--- a/src/base.jl
+++ b/src/base.jl
@@ -236,6 +236,16 @@ function System(buses::Vector{Bus},
     return System(_sys)
 end
 
+"""System constructor without nothing-able arguments."""
+function System(buses::Vector{Bus},
+                generators::Vector{<:Generator},
+                loads::Vector{<:ElectricLoad},
+                basepower::Float64; kwargs...)
+
+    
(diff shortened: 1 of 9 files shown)


src/base.jl
```diff
@@ -104,55 +104,56 @@ function System(; buses=[Bus()],
...
+    A _System struct with a modified forecasts field
 """
-function add_forecast!(sys::System,fc::Pair{Symbol,Array{Forecast,1}})
+function add_forecast!(sys::_System,fc::Pair{Symbol,Array{Forecast,1}})
     sys.forecasts[fc.first] = fc.second
 end
 
-"""A System struct that stores all devices in arrays with concrete types.
+"""
+A _System struct that stores all devices in arrays with concrete types.
```
daniel-thom: We probably shouldn't refer to _System.  You could copy its docstring here and then change _System to say that it is an internal implementation detail.

src/base.jl
```diff
@@ -192,8 +193,59 @@ function ConcreteSystem(sys::System)
...
+                basepower::Float64,
+                forecasts::Union{Nothing, SystemForecasts},
+                services::Union{Nothing, Vector{ <: Service}},
+                annex::Union{Nothing,Dict}; kwargs...)
+
+    
+    _sys = _System(buses, generators, loads, branches, storage, basepower, forecasts, services, annex; kwargs...)
+    return System(_sys)
+end
+
```
daniel-thom: I like your choice of constructors.  But didn't @jd-lara say that he wanted explicit constructors that dealt with optional Branches and Loads?

daniel-thom:
Looks good to me except for the docstring.

jd-lara:
Please do not merge this change just yet even if approved. I want to add the support for the ED and UC formulations before having to perform this update.

src/base.jl
claytonpbarrows: 073620e5840a3e2abc5179f3ac3e7499bc23ae9e resolves

src/base.jl
claytonpbarrows: 08abf44e91b479f54c265022309060ef23582340 resolves

Issue #228, opened 2019-05-07
make a get_forecast function


jd-lara:
we need a function that gets the relevant forecast given an array of devices

jd-lara:
closed via #233

PR #233, opened 2019-05-11
Forecast storage and access improvements

### LINKED ISSUE #231: make ComponentIterator a struct
Currently, the return of get components is a `ComponentIterator{T}` defined as 

```Julia
const ComponentIterator{T} = Base.Iterators.Flatten{Vector{Vector{T}}} where {T <: Component}
```

It will be good to make this as struct that includes a length field such as:

```Julia
struct ComponentIterator{T <: Component}
    it::Base.Iterators.Flatten{Vector{Vector{T}}}
    length::Int64
end
```

such that a method 

```Julia
length(it::ComponentIterator) = it.length 
```

can also be defined

daniel-thom:
Changes data structure for forecasts and adds functions to add, get, and remove forecasts.

Fixes #231, #228.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -136,7 +136,7 @@ end
     * `loads`::Vector{ElectricLoad} : an array of load specifications that includes timing of the loads
     * `branches`::Union{Nothing, Vector{Branch}} : an array of branches; may be `nothing`
     * `storage`::Union{Nothing, Vector{Storage}} : an array of storage devices; may be `nothing`
-    * `basepower`::Float64 : the base power value for the system 
+    * `basepower`::Float64 : the base power value for the system
     * `ps_dict`::Dict{String,Any} : the dictionary object containing System data
     * `forecasts`::Union{Nothing, SystemForecasts} : dictionary of forecasts
     * `services`::Union{Nothing, Vector{ <: Service}} : an array of services; may be `nothing`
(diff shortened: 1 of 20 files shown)


src/base.jl
```diff
@@ -280,15 +284,116 @@ function add_component!(sys::ConcreteSystem, component::T) where T <: Component
...
 """
 Args:
     A ConcreteSystem struct
     A :Symbol=>Array{ <: Forecast,1} Pair denoting the forecast name and array of device forecasts
 Returns:
     A System struct with a modified forecasts field
 """
-function add_forecast!(sys::ConcreteSystem,fc::Pair{Symbol,Array{Forecast,1}})
-    sys.forecasts[fc.first] = fc.second
+function add_forecast!(sys::ConcreteSystem, fc::Pair{Symbol, Vector{Forecast}})
```
daniel-thom: We do need a new add_forecast function.  Should it accept System and a vector of forecasts?  We can determine IssueTime and then store accordingly.  Do we need a function that accepts one forecast?

src/base.jl
```diff
@@ -280,15 +284,116 @@ function add_component!(sys::ConcreteSystem, component::T) where T <: Component
...
+end
+
+"""Returns a vector of forecasts that match the components and label."""
+function get_forecasts(sys::ConcreteSystem, issue_time::IssueTime, components_iterator,
+                       label::Union{String, Nothing}=nothing)
+    forecasts = Vector{Forecast}()
+
+    # This code caches the passed component UUIDs in a dict so that we don't have to
+    # iterate across them for each forecast.
+    components = LazyDictFromIterator(Base.UUID, Component, components_iterator, get_uuid)
```
daniel-thom: Another option is to let the user create the LazyDictFromIterator and pass it in.  This could be useful if they are going to call this function multiple times with the same components.  However, I don't like the added complexity.  I think LazyDictFromIterator should remain an implementation detail rather than a public object.

src/base.jl
claytonpbarrows: Given that we have an `add_component` function, I think we should have an `add_forecast!` function that accepts a single forecast. Otherwise, your proposal for arguments seems fine to me.

src/models/forecasts.jl
```diff
@@ -1,7 +1,16 @@
...
 
-const Forecasts = Vector{ <: Forecast}
-const SystemForecasts = Dict{Symbol, Forecasts}
+const Forecasts = Vector{<:Forecast}
+# This is deprecated because only the legacy System uses it. Parsing code need to change to
+# build the new format.
+const SystemForecastsDeprecated = Dict{Symbol, Forecasts}
+const IssueTime = NamedTuple{(:resolution, :initialtime),
+                             Tuple{Dates.Period, Dates.DateTime}}
+const SystemForecasts = Dict{IssueTime, Forecasts}
```
claytonpbarrows: I like this structure.

src/base.jl
```diff
@@ -280,15 +284,116 @@ function add_component!(sys::ConcreteSystem, component::T) where T <: Component
...
+function get_forecasts(sys::ConcreteSystem, issue_time::IssueTime)
+    if !haskey(sys.forecasts, issue_time)
+        error("forecast issue_time {issue_time} does not exist")
+    end
+
+    return Iterators.take(sys.forecasts[issue_time], length(sys.forecasts[issue_time]))
+end
+
+"""Returns a vector of forecasts that match the components and label."""
+function get_forecasts(sys::ConcreteSystem, issue_time::IssueTime, components_iterator,
```
daniel-thom: This implementation might not be exactly what we want.  It will return all the forecasts that match the components and label.  However, it will not throw an error if it does not find a forecast matching one of the components. Should it?

claytonpbarrows: In my opinion, yes, it should throw an error if it doesn’t find a forecast for one of the items in the iterator

daniel-thom: What about the scenario where the user passes the result of get_components() to get_forecasts()?  They could pass something very broad like Component to get_components and get back components that are not tied to forecasts, right?  In that case they may not care if some are not matched.  I have no problem restricting the behavior, but just want to ensure all use cases are covered.

claytonpbarrows: That seems valid. Perhaps the default should error if some components aren’t matched with forecasts and we could have a kwarg like filter=false to enable the behavior you mention

claytonpbarrows: That seems valid. Perhaps the default should error if some components aren’t matched with forecasts and we could have a kwarg like filter=false to enable the behavior you mention

daniel-thom:
@claytonpbarrows @jd-lara This is now ready for final review.

src/base.jl
```diff
@@ -345,17 +334,170 @@ function add_component!(sys::System, component::T) where T <: Component
...
+        end
+
+        component_uuid = get_uuid(forecast.component)
+        if in(component_uuid, components)
+            push!(forecasts, forecast)
+            push!(matched_components, component_uuid)
+        end
+    end
+
+    if length(components) != length(matched_components)
```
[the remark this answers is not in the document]
daniel-thom: I don't follow this comment. The code does throw an error if it does not find a forecast with one of the components. Can you be more specific?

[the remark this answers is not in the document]
jd-lara: This is an old comment, don't know why it just showed up after I approved. You can disregard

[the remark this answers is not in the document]
jd-lara: This is an old comment, don't know why it just showed up after I approved. You can disregard

PR #241, opened 2019-05-17
add rating field to devices

### LINKED ISSUE #127: add device rate in MVA in data models

jd-lara:
closes #127


jd-lara:
@claytonpbarrows `rating` is a parameter in relation to the current carrying capacity of the device, generally limited by P^2 + Q^2 <= S^2. `installedcapacity` is related to the total available power installed and applied to both devices and collections of devices. For instance, a power plant has an installed capacity but the generators inside have ratings. 

In models with reactive power supply, we care for the rating of the units in the same way the rating of the lines is not the same as the limits imposed for other reasons. 

I would even argue that `installedcapacity` is more of an economic term than an electrical one.

Issue #242, opened 2019-05-17
parsing creates new instances of structs in fields that reference other structs


claytonpbarrows:
e.g. parsing results in lines where the endpoints are buses, but they have different uuid values that the busses in `get_components(Bus,sys)`
 - same for `LoadZones.buses`
 - presumably same for any struct that references another struct
 - happens when parsing (from cdm dict, or pm dict) 
 - this needs to be addressed to fix #240

daniel-thom:
To be clear the problem isn't about references to other values.  When we fill in the field contributingdevices for Services we use an existing device reference.  So, ProportionalReserve.contributingdevices has devices with UUIDs that match devices stored inside System.components.

Buses are a different story.  During parsing the function make_bus() is repeatedly called with the same values, so the instances get different UUIDs.  It does not use an existing reference.

Two potential solutions:  Check if we have the same bus during parsing before creating another (in make_bus) or consolidate duplicate buses at the end.

claytonpbarrows:
to clarify:
```julia
julia> sys5 = System(nodes5, thermal_generators5, loads5, branches5, nothing,  100.0, nothing,
       nothing, nothing)
System(Dict{DataType,Array{#s308,1} where #s308<:Component}(PowerLoad=>PowerLoad[PowerLoad(name="Bus2"), PowerLoad(name="Bus3"), PowerLoad(name="Bus4")],ThermalDispatch=>ThermalDispatch[ThermalDispatch(name="Alta"), ThermalDispatch(name="Park City"), ThermalDispatch(name="Solitude"), ThermalDispatch(name="Sundance"), ThermalDispatch(name="Brighton")],Line=>Line[Line(name="1"), Line(name="2"), Line(name="3"), Line(name="4"), Line(name="5"), Line(name="6")],Bus=>Bus[Bus(name="nodeA"), Bus(name="nodeB"), Bus(name="nodeC"), Bus(name="nodeD"), Bus(name="nodeE")]), nothing, 100.0, PowerSystems.PowerSystemInternal(UUID("af56c8ae-6794-45cb-992e-e45572dd7b6c")))

julia> lines = get_components(Line,sys5)|>collect
6-element Array{Line,1}:
 Line(name="1")
 Line(name="2")
 Line(name="3")
 Line(name="4")
 Line(name="5")
 Line(name="6")

julia> buses = get_components(Bus,sys5)|>collect
5-element Array{Bus,1}:
 Bus(name="nodeA")
[...]

Issue #240, opened 2019-05-16
parsing of load zones


jd-lara:
When we parse the data from RTS using load_zones for the time_series information it doesn't create forecasts for individual loads. As a result when we run `get_forecast` it returns an error 

`PowerSystems.InvalidParameter("did not find forecasts for one or more components")`

This implies that when parsing data with Load Zones, there has to be a way to assign it to the loads for modeling.

PR #243, opened 2019-05-17
fix dict parsing

claytonpbarrows:
this PR addresses #240 and #242, but it does not ensure the uniqueness of the struct instances via a check (as suggested by @daniel-thom in #242).


jd-lara:
This PR needs #241 to pass since the data has been already updated in PowerSystemsData

daniel-thom:
Do we end up with duplicates because of the calls to make_bus() in src/parsers/pm2ps_parser.jl?

claytonpbarrows:
yeah, basically in both dict parsing routes (PM, and CDM), the first thing that happens is reading the Buses. We had a bunch of checks to make sure that a bus existed before referencing it in lines an other devices, but this just formalizes that and actually creates the `Bus` in the first step and then directly references those items rather than just an index or ~matching data.

PR #247, opened 2019-05-19
Struct updates

### LINKED ISSUE #226: Rename ThermalDispatch
We need to rename the type PSY.ThermalDispatch struct since both PowerSimulations and PowerSystems export the same name. 

@claytonpbarrows any ideas?

### LINKED ISSUE #244: InterruptibleLoad missing `econ` field
See https://github.com/NREL/PowerSimulations.jl/issues/152#issuecomment-493705735 .

### LINKED ISSUE #245: Rename `status` argument to `available` in `ThermalDispatch` constructor?
I noticed that the argument `status` for the `ThermalDispatch` constructor doesn't match the `available` field it its struct: https://github.com/NREL/PowerSystems.jl/blob/570c5d08a20241216a3789a76f209b1d9520ce93/src/models/generation/thermal_generation.jl#L17-L29

It looks like this inconsistency occurs for at least one other type: https://github.com/NREL/PowerSystems.jl/blob/570c5d08a20241216a3789a76f209b1d9520ce93/src/models/generation/thermal_generation.jl#L42-L56

jd-lara:
Closes #226
Closes #244 
Closes #245 

After this update I can add testing in PowerSimulations for Renewable Reactive Power Dispatch and Interruptible Loads.


src/PowerSystems.jl
```diff
@@ -48,19 +48,20 @@ export RenewableGen
 export TechRenewable
 export EconRenewable
 export RenewableFix
-export RenewableCurtailment
-export RenewableFullDispatch
+export RenewableDispatch
+export RenewableDispatch
```
claytonpbarrows: Is this duplicated?

src/models/loads/econ_common.jl
```diff
@@ -0,0 +1,20 @@
...
+Data Structure for the economical parameters of interruptible loads.
+    The data structure can be called calling all the fields directly or using named fields.
+    All the limits are defined by NamedTuples and some fields can take ```nothing```
+
+    ## Examples
+
+
+"""
+struct EconLoad <: TechnicalParams
+    curtailpenalty::Float64 # [$/event]
```
claytonpbarrows: What about ‘VOLL’ instead? Too specific?

src/models/generation/hydro_generation.jl
```diff
@@ -84,16 +84,16 @@ function HydroCurtailment(name, available, bus, tech, econ)
     return HydroCurtailment(name, available, bus, tech, econ, PowerSystemInternal())
 end
 
-function HydroCurtailment(name::String, status::Bool, bus::Bus, tech::TechHydro, curtailcost::Float64)
+function HydroCurtailment(name::String, available::Bool, bus::Bus, tech::TechHydro, curtailcost::Float64)
```
claytonpbarrows: We should make this consistent with renewable and call it HydroDispatch

src/PowerSystems.jl
```diff
@@ -48,19 +48,20 @@ export RenewableGen
...
 
 export ThermalGen
 export TechThermal
 export EconThermal
-export ThermalDispatch
+export StandardThermal
```
claytonpbarrows: Id like this to start with Thermal... ThermalStandard would be ok.

src/models/loads/econ_common.jl
jd-lara: I would say too specific

jd-lara:
@claytonpbarrows I made the changes. The open question is if we apply the rule of DeviceName First also to Loads?

claytonpbarrows:
You’re right. LoadStandard reads weird. But it’s nice to be able to look through the different devices in a consistent way. I’m not sure. Opinions from @kdheepak or @daniel-thom or others?

jd-lara:
How about we keep that rule for the generators ?

src/models/generation/renewable_generation.jl
```diff
@@ -34,37 +35,20 @@ struct RenewableCurtailment <: RenewableGen
...
-RenewableFullDispatch(; name = "init",
-                status = false,
-                bus= Bus(),
-                rating = 0.0,
-                econ = EconRenewable()) = RenewableCurtailment(name, status, bus, rating, econ)
+"""Accepts rating as a Float64 and then creates a TechRenewable."""
+function RenewableDispatch(name::String, available::Bool, bus::Bus, rating::Float64, econ::Union{EconRenewable,Nothing})
+    tech = TechRenewable(rating, nothing, 1.0)
+    return RenewableDispatch(name, available, bus, tech, econ)
+end
```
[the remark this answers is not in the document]
jd-lara: Fixed via 97e5023 and also changes the settings in my editor

PR #250, opened 2019-05-22
 Added iterators for components and forecasts.

daniel-thom:


src/base.jl
```diff
@@ -414,7 +455,8 @@ function get_forecasts(
         end
     else
         keys_ = [_ForecastKey(initial_time, x.forecast_type)
-                 for x in keys(sys.forecasts.data) if x.forecast_type <: T]
+                 for x in keys(sys.forecasts.data) if x.initial_time == initial_time &&
```
daniel-thom: This was a bug in my previous PR.

src/models/forecasts.jl
```diff
@@ -99,7 +99,7 @@ function SystemForecasts(data::NamedTuple)
     return SystemForecasts(_Forecasts(), initial_time, resolution, horizon, interval)
 end
 
-function Base.show(io::IO, forecasts::SystemForecasts)
+function Base.summary(io::IO, forecasts::SystemForecasts)
```
daniel-thom: On second thought, I think summary should be used here, which is what it was originally.

Issue #249, opened 2019-05-22
Parsing of forecast data upgrades


jd-lara:
The current implementation of parsing forecast data in `cdm_parser.jl` has the forecasts names hardcoded and generates the keys for the `ps_dict` based on the name of the files. As a result, a `ps_dict` can contain more than one set of time series and more than one resolution, yet the name of the time series doesn't reflect the distinction between sets of time series. 

This creates a challenge when creating SystemForecasts since a System should contain forecasts of the same resolution. Also, this parsing scheme for the forecasts won't allow the user to specify a look-ahead and an interval to slice the time series data according to their specifications. 

The proposal to tackle these shortcomings is as follows: 

1.  Assuming the time series data is provided as columns as currently stored in the CDM data format, when parsing the forecast data, eliminate the use ok hardcoded keys "DA" and "RT" and substitute that with a tuple of the resolution of the data and the length of the time series. Hence, we will have ~DA~ => (1 Hour, 8760) and ~RT~ => (5 Minutes, 105120). 

2. With this information, we store in `ps_dict` a configuration for the forecast arrays that specify the time series to be used, interval and length of the lookahead. Probably this will require some interface functions to avoid direct manipulation of the `ps_dict`. 

3. The `make_forecat_array` function takes the selected time series data set and the specification (interval, lookahead) and generates the `SystemForecast` struct accordingly. This function will throw exceptions/warnings or errors if there are inconsistencies in the data lengths or time_steps. 

Simple functions like `missing_data = setdiff(whole_year,dates_of_parsed_data)` can show if there are inconsistencies across the time series data or missing data points.

daniel-thom:
This comment is a suggestion for a general change in how we store the forecast data while parsing.  We currently build a deeply nested dictionary like this:

```
dict["forecasts"]["DA"]["gen"]["Renewable"]["PV"] = {
    name => TimeSeries.TimeArray(),
}
```

This introduces a few problems with the code:
- complexity
- hard to extend
    * What happens if you want to add something in between "gen" and "Renewable?"
    * Other people will have difficulty matching this format with their data.

Instead, we could store arrays of dictionaries with known fields like the example below.  This would make it easier to build our System and Forecast structs, easier to extend for other people's data, etc.

```
dict["forecasts"] = [
    {
        "name" => "324_PV_3",
         "label" => "DA",
        "generator" => "Renewable",
        "type" => "PV",
        "resolution" => Dates.Minute(5),
        "length" => 8760,
[...]

PR #251, opened 2019-05-24
 Prototype for simplifying forecast parsing.

daniel-thom:
This is a starting point for discussion on issue #249.  I changed the forecast parsing code to build a flat array of forecast dictionaries instead of a deeply nested dictionary keyed by the forecast attributes.

They array of forecasts gets passed to the System constructor.  The System breaks up the forecasts into its own optimized data structure.

I have several open questions about some specifics (resolution, initial_time, etc.), but am hoping that this is a more extensible design.  Let me know what you think.

I did not update the tests because it would require a bit of refactoring.  I want to get agreement first.  To test the code I performed these steps:
```
cdm_dict = PowerSystems.csv2ps_dict("./data/RTS_GMLC", 100.0);
sys = System(cdm_dict)
for forecast in iterate_forecasts(sys)
    @show forecast
end
```

Current status as of 1pm on 5/24:
- Changed the target of this PR to PowerSystems/forecast-parsing instead of dev.  @jd-lara may implement some additions.  If he does, he will need to perform a merge of this PR.
- Need to have discussion with Clayton about these interfaces before proceeding much further.
- Need to discussion with Clayton and Dheepak about longer-term interfaces for parsing.  The intermediate dictionary we produce currently (ps_dict) is not ideal.  Jose and I discussed some possibilities today.  Parsing could produce either existing PowerSystems structs or simplified, mutable versions that can be fed into the System constructor.  The forecast parsing points out why this could be desirable.  With the current plan the user will need to filter out forecasts to ensure only one resolution exists.  Doing this with structs would be much clearer than with dictionaries.

### DIFF
--- a/src/models/forecasts.jl
+++ b/src/models/forecasts.jl
@@ -40,8 +40,8 @@ end
 """Constructs SystemForecasts from the flat vector of forecasts resulting from parsing."""
 function SystemForecasts(forecasts::Forecasts)
     forecasts_by_type = ForecastsByType()
-    initial_time = Dates.DateTime(Dates.Second(1))
-    resolution = Dates.Period(Dates.Second(1))
+    initial_time = Dates.DateTime(Dates.Second(0))
+    resolution = Dates.Period(Dates.Second(0))
     initialized = false
     horizon::Int64 = 0
 
--- a/src/models/forecasts.jl
+++ b/src/models/forecasts.jl
@@ -1,18 +1,15 @@
 abstract type Forecast <: PowerSystemType end
 
 const Forecasts = Vector{<:Forecast}
-# This is deprecated because only the legacy System uses it. Parsing code need to change to
-# build the new format.
-const SystemForecastsDeprecated = Dict{Symbol, Forecasts}
 
-struct _ForecastKey
+struct ForecastKey
     initial_time::Dates.DateTime
     forecast_type::DataType
 end
 
-const _Forecasts = Dict{_ForecastKey, Vector{<:Forecast}}
+const ForecastsByType = Dict{ForecastKey, Vector{<:Forecast}}
 
-function _get_forecast_initial_times(data::_Forecasts)::Vector{Dates.DateTime}
+function _get_forecast_initial_times(data::ForecastsByType)::Vector{Dates.DateTime}
     initial_times = Set{Dates.DateTime}()
     for key in keys(data)
         push!(initial_times, key.initial_time)
@@ -21,8 +18,8 @@ function _get_forecast_initial_times(data::_Forecasts)::Vector{Dates.DateTime}
     return sort!(Vector{Dates.DateTime}(collect(initial_times)))
 end
 
-function _add_forecast!(data::_Forecasts, forecast::T) where T <: Forecast
-    key = _ForecastKey(forecast.initial_time, T)
+function _add_forecast!(data::ForecastsByType, forecast::T) where T <: Forecast
+    key = ForecastKey(forecast.initial_time, T)
     if !haskey(data, key)
         data[key] = Vector{T}()
     end
@@ -33,22 +30,22 @@ end
 """Container for forecasts and their metadata. Implementation detail that is not exported.
 Functions to access the data should go through the System."""
 struct SystemForecasts
-    data::_Forecasts
+    data::ForecastsByType
     initial_time::Dates.DateTime
     resolution::Dates.Period
     horizon::Int64
     interval::Dates.Period
 end
 
-"""Constructs SystemForecasts from the previous version of the data structure."""
-function SystemForecasts(old_forecasts::SystemForecastsDeprecated)
-    forecasts = _Forecasts()
+"""Constructs SystemForecasts from the flat vector of forecasts resulting from parsing."""
+function SystemForecasts(forecasts::Forecasts)
+    forecasts_by_type = ForecastsByType()
     initial_time = Dates.DateTime(Dates.Second(1))
     resolution = Dates.Period(Dates.Second(1))
     initialized = false
     horizon::Int64 = 0
 
-    for forecast in Iterators.flatten(values(old_forecasts))
+    for forecast in forecasts
         if !initialized
             initial_time = forecast.initial_time
             resolution = forecast.resolution
@@ -56,10 +53,7 @@ function SystemForecasts(old_forecasts::SystemForecastsDeprecated)
             initialized = true
         else
             if forecast.resolution != resolution
-                # TODO parsing code currently produces multiple resolutions.
-                @error("Skipping forecast because it has a different resolution",
-                       resolution, forecast.resolution, maxlog=1)
-                #throw(DataFormatError("found multiple resolution values in forecasts"))
+                throw(DataFormatError("found multiple resolution values in forecasts"))
                 continue
             end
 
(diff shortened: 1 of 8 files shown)


src/parsers/cdm_parser.jl
```diff
@@ -736,6 +699,71 @@ function services_csv_parser(rsv_raw::DataFrames.DataFrame,gen_raw::DataFrames.D
...
 end
 
+"""Return a dictionary that includes a vector of parsed forecast dictionaries."""
+function forecast_csv_parser(ps_dict::Dict, data::Dict)
+    forecasts = Dict(
+        "forecasts" => Vector{Dict}(),
+        "loads" => Vector{Dict}(),
+    )
+
+    # A system can only contain forecasts with one resolution. Skip any that don't
```
daniel-thom: I need some help here.

src/parsers/cdm_parser.jl
```diff
@@ -736,6 +699,71 @@ function services_csv_parser(rsv_raw::DataFrames.DataFrame,gen_raw::DataFrames.D
...
+                first_resolution = resolution
+            elseif resolution != first_resolution
+                @error("found timeseries with a different resolution", resolution,
+                       first_resolution, maxlog=3)
+                continue
+            end
+            push!(forecasts["forecasts"], forecast)
+        end
+
+        # TODO DT: Is this used anywhere?
```
daniel-thom: The structure could be changed depending on how we will use this data.

src/parsers/cdm_parser.jl
jd-lara: At a Dict level, we can load forecasts with more than one resolution. We should have some sort of interface that tells the user that more than one resolution is available and choose one. 

We would actually use the resolution of the selected forecast to actually build and test the whole system data.

src/parsers/cdm_parser.jl
jd-lara: This looks good to me. I would just enable the possibility of having more than one forecast dict to accommodate more than one resolution when parsing the CSV's

jd-lara:
This PR is pretty great, it is a significant improvement on the clarity of the forecast parsing.

src/parsers/dict_to_struct.jl
```diff
@@ -57,57 +57,64 @@ function ps_dict2ps_struct(data::Dict{String,Any})
...
+                # Recurse.
+                _retrieve(value, key_of_interest, output)
+            end
         end
-        iter_result = Base.iterate(dict, state)
     end
-    return output
 end
 
+# TODO DT: alternate implementation
```
daniel-thom: I first simplified the existing function to this implementation.  Then I decided what we actually want is a reference to the actual object, so went with the implementation above.  If we actually do want to use the array of keys, let me know.  But I find this simpler.

Something else to think about:  we could change ps_dict along the same lines as the forecasts in this PR.  We could store arrays of component dictionaries instead of nested dictionaries keyed by attributes.  Aside from code complexity, the main problem with the nested dictionaries is that it is hard to add layers.

src/models/forecasts.jl
```diff
@@ -33,33 +30,30 @@ end
...
 end
 
-"""Constructs SystemForecasts from the previous version of the data structure."""
-function SystemForecasts(old_forecasts::SystemForecastsDeprecated)
-    forecasts = _Forecasts()
+"""Constructs SystemForecasts from the flat vector of forecasts resulting from parsing."""
+function SystemForecasts(forecasts::Forecasts)
+    forecasts_by_type = ForecastsByType()
     initial_time = Dates.DateTime(Dates.Second(1))
     resolution = Dates.Period(Dates.Second(1))
```
jd-lara: I had missed this from before, but why do we have the resolution here in Seconds?

src/parsers/cdm_parser.jl
```diff
@@ -736,6 +699,71 @@ function services_csv_parser(rsv_raw::DataFrames.DataFrame,gen_raw::DataFrames.D
     return services_dict
 end
 
+"""Return a dictionary that includes a vector of parsed forecast dictionaries."""
+function forecast_csv_parser(ps_dict::Dict, data::Dict)
```
jd-lara: This one could be split by resolution

src/parsers/forecast_parser.jl
```diff
@@ -108,42 +62,57 @@ Returns:
...
+function make_forecast_array(sys::System, parsed_forecasts::Vector{Dict})
+    return make_forecast_array(get_components(Component, sys))
+end
+
+function make_forecast_array(all_components, parsed_forecasts::Vector{Dict})
+    # TODO this code could be cleaner if the incoming data was only DataFrame or TimeArray
+    forecasts = Vector{Forecast}()
+    for forecast in parsed_forecasts
+        data = forecast["data"]
+        if data isa DataFrames.DataFrame && size(data, 2) > 2
```
jd-lara: I think this is trying to find if the DataFrame has a forecast for more than one device. Looking at the RTS data each forecast CSV contains data for all the devices of the same type.

src/models/forecasts.jl
daniel-thom: That's just to initialize the variables.  They are always overwritten.

src/parsers/dict_to_struct.jl
jd-lara: 1. we added a yaml file that is supposed to create a mapping between generators to parse the CDM https://github.com/NREL/PowerSystems.jl/pull/164 maybe it is reasonable to do something like this or extend the yaml to account for the forecast? 

2. I am not sure about the rules to modify the ps_dict since we got those from power models. But I certainly support the idea of simpler intermediate dicts.

src/models/forecasts.jl
daniel-thom: I'll change the values to 0.

src/parsers/cdm_parser.jl
```diff
@@ -736,6 +699,71 @@ function services_csv_parser(rsv_raw::DataFrames.DataFrame,gen_raw::DataFrames.D
...
+            else
+                len = length(value)
+                @assert len > 2
+                timestamps = TimeSeries.timestamp(value)
+                resolution = timestamps[2] - timestamps[1]
+            end
+
+            component = get(component_map, name, nothing)
+            if isnothing(component)
+                @error "$name is not in component_map, skipping forecast"
```
daniel-thom: This trips in one case - name = 212_CSP_HEAD_STORAGE.  Seems like there needs to be a secondary lookup to match 212_CSP_1 (both names exist in data/RTS_GMLC/storage.csv).  What do you guys think?

src/parsers/cdm_parser.jl
daniel-thom: SystemForecasts is currently an immutable struct with a field called resolution.  We could make it mutable, add a field called available_resolutions, and let the user change it.  All forecasts could be stored within the data object.

jd-lara: When we create the system we want a single resolution. We don't want to carry around all possible time series for a system.

daniel-thom: Then we are missing a place to insert a new interface.  The user could specify the resolution to use when calling the System constructor.  But what if they don't know that multiple are available?  Are you suggesting a new, intermediate layer?  We could initially store all time series but force a decision if there are multiple, and then delete the unused ones.  (I don't really like that option.)

src/parsers/forecast_parser.jl
claytonpbarrows: I think that I didn’t understand time arrays when this was written. We could probably replace the data frame route with a timearray

claytonpbarrows: I think that I didn’t understand time arrays when this was written. We could probably replace the data frame route with a timearray

src/parsers/cdm_parser.jl
jd-lara: I think this is an issue of the RTS data @claytonpbarrows can confirm.

src/parsers/cdm_parser.jl
daniel-thom: The parsing code now stores all forecasts regardless of resolution.  The System construction will fail if there are multiple resolutions.  It will be up to the user to choose one resolution and filter out the rest before calling the System constructor.

PR #254, opened 2019-06-06
Type fixes

jd-lara:
Add new category for topological components and separate it from `injection`. Creates an explicit separation such that a `device` represents physical components of the system. 

This PR leads to thinking we need to rename `component` to a more generic type like element (?) or another name. Also because during the dynamic model components should be used to define parts inside of the device e.g. AVR is a component of a generator.

The tree will look like this

image

### DIFF
--- a/src/models/topological_elements.jl
+++ b/src/models/topological_elements.jl
@@ -1,5 +1,5 @@
 
-abstract type Topological <: Device end
+abstract type Topological <: Component end
 """
     Bus
 
--- a/src/models/topological_elements.jl
+++ b/src/models/topological_elements.jl
@@ -1,3 +1,5 @@
+
+abstract type Topological <: Device end
 """
     Bus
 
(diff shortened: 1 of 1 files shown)


jd-lara:
Topological is a broader term; it will enable creating collections of items like interties as collections of branches. Injection and branches can also be renamed one and two terminal devices. 

Noted about the Element vs component, still think we should rename the field components.

claytonpbarrows:
How about this:
image
Here, we move `Bus` (labeled `Buss` since `Bus` already exists) up to `Device` from `Injection`, and add a `Collection` that includes structs that define groups of devices. A collection could really be anything, including groups of `Bus`es that define `Substations`, `Generator`s in a state, generator categories, or anything else.

jd-lara:
What does buss mean?

claytonpbarrows:
`Buss` = `Bus`

jd-lara:
But Bus already exists, are you proposing to have both Bus and Buss?

claytonpbarrows:
No, see ^^^ : I just created `Buss` for demonstration because we already have `Bus` and I'm lazy. I'm proposing just moving `Bus` from under `Injection` to under `Device`.

jd-lara:
I will close this PR in favor of a more comprehensive proposal that tackles #259

PR #255, opened 2019-06-10
Rewrite CDM  and forecast parsing.

daniel-thom:
This PR implements the follow changes:
- Defines a parsing descriptor file in JSON.  This file lists each supported power system object that can be parsed in a CSV file.  Users can run a script to generate a customizable user descriptor file if their CSV column names do not match our standard names.
- CDM parsing loads the descriptor files to create a PowerSystem directly from the RTS_GMLC CSV files.  This parsing process no longer creates the intermediate PS Dict.  The details of reading specific column names and values from DataFrames is abstracted from the main flow.
- Defines timeseries format Julia types that we support.  Refactored the forecast-parsing code while trying to maintain existing behavior.  One deviation from collaboration with Clayton is that the code now uses a component category to help match forecasts to components.  Earlier, name collisions between components of different types would result selecting the wrong forecast.  Now, you can specify component category and component name.
- The forecast storage struct SystemForecasts is now mutable.  The System struct now always has a value for this.  Before, the value could be nothing or SystemForecasts.  This was required to support dynamic building of forecasts after the System is constructed.  All component structs are still immutable.

The branch is only compatible with the "improve-timeseries" branch of PowerSystemTestData.

Changes now should be considered final.


bin/generate_config_file.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python
```
daniel-thom: User documentation is needed for this.

src/common.jl
```diff
@@ -5,6 +5,20 @@
     POLYNOMIAL = 2
 end
 
+@enum AngleUnit begin
+    DEGREES
+    RADIANS
+end
+
+@enum BusType begin
```
daniel-thom: Note this change.  The code now uses this instead of strings.  It should lead to better robustness, but it could also break some existing code. Let me know if you disagree.

Also, note the comment below about "ISOLATED"

src/parsers/dict_to_struct.jl
```diff
@@ -131,55 +89,7 @@ function _get_device(name::Union{String,Symbol}, collection, devices = [])
...
-            @warn "No reference date given, assuming today"
-            startdatetime = Dates.today()
-        end
-        df[:DateTime] = collect(Dates.DateTime(startdatetime):Dates.Hour(1):Dates.DateTime(startdatetime)+Dates.Hour(size(df)[1]-1))
-    end
-    return df
-end
-
-
+# TODO DT: add_time_series and add_time_series_load are unused. Delete?
```
daniel-thom: Do we still need this code?

src/parsers/timeseries_formats.jl
```diff
@@ -0,0 +1,234 @@
...
+    for i in 1:length(file)
+        timestamp = get_timestamp(T, file, i) + step * (i - 1)
+        push!(timestamps, timestamp)
+    end
+
+    value_columns = get_value_columns(T, file)
+    vals = [getproperty(file, x) for x in value_columns]
+
+    if length(value_columns) == 1 && !isnothing(component_name) && string(value_columns[1]) != component_name
+        # TODO DT: Hack. Do we really want to support this?
```
daniel-thom: I think we should remove this.  I think we should make it a requirement that the column name be a valid component name.

src/parsers/timeseries_formats.jl
```diff
@@ -0,0 +1,234 @@
...
+function get_step_time(
+                       ::Type{T},
+                       file::CSV.File,
+                       period::AbstractArray,
+                      ) where T <: TimeseriesFormat
+    @assert period[end] == maximum(period) == length(period)
+    num_steps = period[end]
+
+    timestamps = get_unique_timestamps(T, file)
+    if length(timestamps) == 1
```
daniel-thom: What do you all think about this?

src/parsers/cdm_parser.jl
```diff
@@ -51,704 +129,732 @@ function read_csv_data(file_path::String, baseMVA::Float64)
...
 
-function _format_fcdict(fc,obj_map)
-    paths = Dict()
-    for (k,d) in fc
-        if haskey(obj_map, k)
-            _add_nested_dict!(paths,obj_map[k],d)
-        end
+"""Convert the dataframe's columns for the category to per_unit."""
+function convert_columns_per_unit!(data::PowerSystemRaw, category::InputCategory, columns)
+    # Asserting here is appropriate for the current code. It prevents someone from
```
daniel-thom: You might want something more advanced here.

src/parsers/cdm_parser.jl
```diff
@@ -51,704 +129,732 @@ function read_csv_data(file_path::String, baseMVA::Float64)
...
 """
-Args:
-    A DataFrame with the same column names as in RTS_GMLC gen.csv file
-    Parsed Bus PowerSystems dictionary
-Returns:
-    A Nested Dictionary with keys as generator types/names and values as
-    generator data dictionary with same keys as the device struct
+function bus_csv_parser!(sys::System, data::PowerSystemRaw)
+    for bus in iterate_rows(data, BUS::InputCategory)
+        # TODO DT:  confirm.
```
daniel-thom: Need some help here.

src/parsers/dict_to_struct.jl
```diff
@@ -57,58 +57,16 @@ function ps_dict2ps_struct(data::Dict{String,Any})
         @warn "key 'services' not found in PowerSystems dictionary, this will result in an empty services array"
     end
     if haskey(data,"forecasts")
-        devices = collect(vcat(buses,generators,storage,branches,loads,loadZones,shunts,branches,services))
-        forecasts = PowerSystems.forecasts_dict_parser(data["forecasts"],devices)
+        devices = vcat(buses,generators,storage,branches,loads,loadZones,shunts,branches,services)
+        forecasts = make_forecast_array(devices, data["forecasts"]["forecasts"])
```
daniel-thom: This is broken and must be fixed with the new format.

src/common.jl
```diff
@@ -5,6 +5,20 @@
...
+    REF
+    ISOLATED  # TODO DT: this was in a matpower file, verify
```
claytonpbarrows: There was some discussion on our support for islanded buses. @jd-lara I don't recall where we landed. If we don't include the `ISOLATED` bus type in the parser, does the data just get skipped?

src/parsers/timeseries_formats.jl
claytonpbarrows: Agreed

src/parsers/dict_to_struct.jl
claytonpbarrows: Don't think so

src/parsers/timeseries_formats.jl
claytonpbarrows: for `DateTime` with a single value, we should assume 1 hour resolution.

src/parsers/timeseries_formats.jl
daniel-thom: This means that the 6 files matching data/RTS_GMLC/Reserves/.*regional_Spin_Up_R1.csv will have to be corrected or skipped.

src/parsers/cdm_parser.jl
claytonpbarrows:
```
Ref = SF
ISOLATED = not supported
PQ = PQ
PV = PV
SF = SF
SLACK = SF
UNKNOWN = PQ
```

src/common.jl
daniel-thom: It will currently throw an error because it can convert "isolated" to a type.  I could make it parse to UNKNOWN.

src/parsers/timeseries_formats.jl
claytonpbarrows: Done

daniel-thom: Removed from code

PR #258, opened 2019-06-17
Automatically generate structs from a descriptor file.

daniel-thom:
This PR does the following:
1. Replaces PowerSystems device/component/forecast structs with files automatically generated from a descriptor file (JSON). We can still provide custom constructors for special cases - see src/models/supplemental_constructors.jl.
2. Changes the structs from immutable to mutable.


src/models/branches/lines.jl
```diff
@@ -1,75 +0,0 @@
...
-    internal::PowerSystemInternal
-end
-
-function MonitoredLine(name, available, connectionpoints, r, x, b, flowlimits, rate,
-                       anglelimits)
-    return MonitoredLine(name, available, connectionpoints, r, x, b, flowlimits, rate,
-                         anglelimits, PowerSystemInternal())
-end
-
-function MonitoredLine(; name="init",
```
daniel-thom: One oddity from before:  this constructor returned a Line.  The new code always returns a MonitoredLine.  It is a behavioral difference, so I'm pointing it out in case we really do want the old behavior.

jd-lara: We definitely need the Line constructor.

daniel-thom: More clarification is needed.

Old code:
- Call MonitoredLine with arguments -> return a MonitoredLine.
- Call MonitoredLine with keyword arguments -> return a Line.

New code:
- Always return a MonitoredLine.

What do you want from the new code?

jd-lara: Call MonitoredLine with arguments -> return a MonitoredLine.
Call MonitoredLine with keyword arguments -> return a MonitoredLine

bin/generate_structs.jl
```diff
@@ -0,0 +1,136 @@
+
```
jd-lara: I think this file goes in deps.

src/models/generated/Bus.jl
```diff
@@ -0,0 +1,61 @@
+#=
+This file is auto-generated. Do not edit.
```
jd-lara: Do we need this in every file?

src/models/generated/Bus.jl
```diff
@@ -0,0 +1,61 @@
...
+=#
+
+"""A power-system bus."""
+mutable struct Bus <: Injection
+    number::Int64  # number associated with the bus
+    name::String  # the name of the bus
+    bustype::Union{Nothing, BusType}  # bus type
+    angle::Union{Nothing, Float64}  # angle of the bus in degrees
```
jd-lara: the angle of the bus in degrees in radians, this was a mistake from old code.

src/models/generated/EconThermal.jl
```diff
@@ -0,0 +1,50 @@
...
+=#
+
+"""Data Structure for the economical parameters of thermal generation technologies."""
+mutable struct EconThermal <: PowerSystems.TechnicalParams
+    capacity::Float64
+    variablecost::Union{Tuple{Float64, Float64}, Array{Tuple{Float64, Float64}, N} where N}
+    fixedcost::Float64
+    startupcost::Float64
+    shutdncost::Float64
+    annualcapacityfactor::Union{Nothing, Float64}
```
jd-lara: This field should be removed. AnnucalCapacityFactor isn't relevant for operations.

src/models/generated/Bus.jl
daniel-thom: I think we want to do everything possible to prevent people from changing these files.  I should actually change the files to read-only in the generation script.

src/models/generated/Bus.jl
daniel-thom: Just to be crystal clear:  are you saying it should it be "angle of the bus in radians"?

jd-lara: yes

src/models/generated/Bus.jl
jd-lara: ok... I think that might be too restrictive while are still developing. I need to make some updates to the structs and test them before putting them in the schema. I don't want to re-generate the whole thing every time.

src/models/generated/EconThermal.jl
daniel-thom: Please file an issue.  This requires changes in the parsing code and is not relevant to this PR.

jd-lara: Ok, I will open an issue to do an overall review of all the structs.

bin/generate_structs.jl
daniel-thom: This topic merits some discussion.  We could invoke this script in build.jl and then throw an error if any file has changed from git's perspective.  We need to enforce the rule that the generated files match the descriptor file.  Of course, we could remove the generated files from git and always run this script during build.  But you guys didn't like that.

jd-lara: I don't get why these are mutually exclusive or why it should imply the removal of the generated code in git.

daniel-thom: 1. Generated code is not in git.  The build script runs the generation script.  The descriptor file is always the source of truth.  Whenever the user downloads the repo they have to run build.  They will always get the correct version of generated code.  If the user changes the descriptor file but doesn't rerun the script, that is their fault.
2. Generated code is in git.  The build script runs the generation script and checks for changes.  We have to prevent the situation where a developer commits changes to the descriptor file but forgets to run the script.

PR #261, opened 2019-06-20
Refactored PowerModels parsing

daniel-thom:
This PR is a follow-on to PR #260 (so it shows those commits also).

The changes here are:
- Refactored PM parsing to not use the intermediate ps_dict format.  The parsing process constructs an empty System and dynamically builds it while iterating over the pm_dict.  This makes it line up with the CDM parser.
- Updated the generator mapping config file.
- Changed the CDM parser to use the generator mapping file.
- Deleted now-unused code.  This includes the old _System struct as well as the file dict_to_struct.jl.


src/base.jl
```diff
@@ -143,65 +51,56 @@ function System(components, forecasts, basepower)
...
 
-    for (key, value) in concrete_sys.components
+    for (key, value) in sys.components
         @debug "components: $(string(key)): count=$(string(length(value)))"
     end
 
-    return concrete_sys
-end
+    runchecks = get(kwargs, :runchecks, true)
+    if runchecks
```
daniel-thom: @KyraDiF Note this code path.  I modified it so that it would run in our new parsing paradigm.  These checks and modifications should probably happen in the validation code you are writing.  We will likely need some discussion about it.

src/parsers/pm2ps_parser.jl
```diff
@@ -256,194 +259,233 @@ function make_thermal_gen(gen_name, d, bus)
...
+        elseif gen_type == HydroDispatch
+            generator = make_hydro_gen(gen_name, pm_gen, bus)
+        elseif gen_type == RenewableDispatch
+            generator = make_renewable_dispatch(gen_name, pm_gen, bus)
+        elseif gen_type == RenewableFix
+            generator = make_renewable_fix(gen_name, pm_gen, bus)
+        elseif gen_type == GenericBattery
+            @warn "Skipping GenericBattery"
+            continue
+            # TODO
```
daniel-thom: I'm not sure why this wasn't covered before. It should be simple to add.  Let me know if I should fill it in.

src/utils/IO/system_checks.jl
```diff
@@ -1,34 +1,27 @@
...
+function buscheck(buses)
     for b in buses
-        if b.bustype == nothing
+        if isnothing(b.bustype)
             @warn "Bus/Nodes data does not contain information to build an a network"
         end
     end
 
-    check_ascending_order([x.number for x in buses], "Bus")
+    # TODO DT: do we need this?
```
daniel-thom: Based on the change to storing components in dictionaries instead of arrays we can no longer guarantee ordering.  If this is OK, I will delete this check.  If we do want ordering then we could do one of the following:
1. Store the components in an OrderedDict or SortedDict.
2. Change get_components so that it returns a sorted array.  This would be much less efficient than what we have now.
3. Create a new function that performs the second option.

Let me know what you think.

src/utils/IO/system_checks.jl
```diff
@@ -38,17 +31,19 @@ end
...
+            pv_list[ix] = gen.bus.number
+        end
     end
 
     for b in buses
-        if b.bustype == "PV"
-            b.number in pv_list ? continue : 0 #@warn "The bus ", b.number, " is declared as PV without a generator connected to it"
-        else
-            continue
+        if b.bustype == PV::BusType
```
daniel-thom: The old code effectively did nothing.  Do we want to do something here?

src/parsers/pm2ps_parser.jl
```diff
@@ -115,112 +52,176 @@ function make_bus(bus_name, bus_number, d, bus_types)
...
-                                                "reactivepowerlimits" => (min=d["pmin"], max=d["pmax"]),
-                                                "powerfactor" => 1),
-                    "econ" => Dict{String,Any}("curtailcost" => 0.0,
-                                                "interruptioncost" => nothing)
-                    )
-    return gen_re
+function make_tech_renewable(d)
+    tech = TechRenewable(;
+        rating=float(d["pmax"]),
+        reactivepowerlimits=(min=d["qmin"], max=d["qmax"]),
```
daniel-thom: Something I'll call out is that the old dict_to_struct code ignored reactivepowerlimits, so even though there was matpower data, the TechRenewable got a nothing.  Now it will have something.  The matpower file case5_re.m has (0.0, 0.0).  The recent CDM parsing changes caused the same result.

src/parsers/cdm_parser.jl
```diff
@@ -605,25 +611,26 @@ function services_csv_parser!(sys::System, data::PowerSystemRaw)
...
-        generator = make_thermal_generator(data, gen, cost_colnames, bus_id)
-    elseif gen.fuel == "Hydro"
-        generator = make_hydro_generator(data, gen, bus_id)
-    elseif gen.fuel in ("Solar", "Wind")
-        generator = make_renewable_generator(data, gen, bus_id)
-    elseif gen.fuel == "Storage"
-        generator = make_storage(data, gen, bus_id)
+    gen_type = get_generator_type(gen.fuel, gen.unit_type, data.generator_mapping)
+
+    if gen_type == ThermalStandard
```
jd-lara: Do we have a fuel field in a generator?

src/parsers/pm2ps_parser.jl
jd-lara: We probably need to add another Battery type that fits with PM's Storage Data Model

src/utils/IO/system_checks.jl
jd-lara: Not ordering is ok. No models should rely on order to work. If something broke, as a result, the fix is in the model itself.

src/utils/IO/system_checks.jl
jd-lara: This is not necessary anymore.

src/utils/utils.jl
```diff
@@ -1,18 +1,26 @@
 import InteractiveUtils: subtypes
 
 
+g_cached_subtypes = Dict{DataType, Vector{DataType}}()
+
 """Returns an array of all concrete subtypes of T."""
 function get_all_concrete_subtypes(::Type{T}) where T
```
jd-lara: 👍

src/parsers/cdm_parser.jl
daniel-thom: nope

src/parsers/pm2ps_parser.jl
daniel-thom: I'll leave it for now.

src/utils/IO/system_checks.jl
daniel-thom: Removed.

src/utils/IO/system_checks.jl
daniel-thom: Removed.

Issue #264, opened 2019-07-05
LoadZones forecast parsing populates forecasts for Bus, instead of Load


claytonpbarrows:
```
julia> b = collect(get_components(Bus,sys))[1]
Bus:
   number: 202
   name: Bacon
   bustype: PV
   angle: -10.75906
   voltage: 1.04844
   voltagelimits: (min = 0.95, max = 1.05)
   basevoltage: 138.0
   internal: PowerSystems.PowerSystemInternal(UUID("c7d48343-c353-424b-a227-1c6a5ec7a7d4"))

julia> f = [f for f in collect(get_forecasts(Deterministic,sys,get_forecast_initial_times(sys)[1])) if f.component == b][1]
Deterministic{Bus}(Bus(name="Bacon"), "Requirement", 1 hour, 2020-01-01T00:00:00, 24×1 TimeSeries.TimeArray{Float64,1,DateTime,Array{Float64,1}} 2020-01-01T00:00:00 to 2020-01-01T23:00:00, PowerSystems.PowerSystemInternal(UUID("9c2a37f4-9cb0-4996-9612-3b89dbcd7485")))

julia> f.data
24×1 TimeSeries.TimeArray{Float64,1,DateTime,Array{Float64,1}} 2020-01-01T00:00:00 to 2020-01-01T23:00:00
│                     │ 2         │
├─────────────────────┼───────────┤
│ 2020-01-01T00:00:00 │ 1102.6759 │
│ 2020-01-01T01:00:00 │ 1082.9372 │
│ 2020-01-01T02:00:00 │ 1079.7969 │
│ 2020-01-01T03:00:00 │ 1078.4511 │
│ 2020-01-01T04:00:00 │ 1099.087  │
│ 2020-01-01T05:00:00 │ 1156.0601 │
│ 2020-01-01T06:00:00 │ 1233.2205 │
│ 2020-01-01T07:00:00 │ 1220.6595 │
   ⋮
│ 2020-01-01T17:00:00 │ 1409.0745 │
│ 2020-01-01T18:00:00 │ 1428.3646 │
│ 2020-01-01T19:00:00 │ 1406.8314 │
│ 2020-01-01T20:00:00 │ 1371.3915 │
│ 2020-01-01T21:00:00 │ 1296.9227 │
│ 2020-01-01T22:00:00 │ 1210.7902 │
│ 2020-01-01T23:00:00 │ 1138.5645 │

[...]

PR #266, opened 2019-07-08
Fix forecast parsing

daniel-thom:
This PR addresses these issues:
1. Fix issue #264 
2. Add automatic per-unit conversion for 
    a. CDM parsing of components
    b. CDM parsing of forecasts based on their components.
3. Allow user selection of per-unit conversion when adding forecasts from CSV files.


daniel-thom:
@claytonpbarrows @jd-lara @sourabhdalvi Shall we discuss the question above?

src/parsers/forecast_parser.jl
```diff
@@ -69,36 +70,55 @@ function _forecast_csv_parser!(sys::System, forecast_infos::ForecastInfos, resol
...
-        for component_ in forecast_components
-            timeseries = forecast.data[Symbol(forecast.component_name)]
-            forecast_ = Deterministic(component_, forecast.label, timeseries)
+        for component in forecast_components
+            sym = Symbol(get_name(forecast.component))
+            timeseries = forecast.data[sym]
+            if sym in per_unit_conversions
+                @assert forecast.per_unit
+            elseif forecast.per_unit
+                # PERF
```
daniel-thom: Note this.  It is unfortunate.  If we used DataFrames instead then we could avoid copies.

jd-lara: DataFrames instead of TimeSeriesArrays?

daniel-thom: Yes.  I'm not advocating for a switch.  I still like the choice of TimeArrays for this purpose.

src/parsers/forecast_parser.jl
```diff
@@ -140,34 +170,51 @@ end
...
                            )
-    if !haskey(infos.data_files, data_file)
-        if !isfile(data_file)
-            msg = "Timeseries file doesn't exist : $file_path"
-            throw(DataFormatError(msg))
-        end
+    timeseries = _add_forecast_data!(infos, data_file, get_name(component))
 
-        infos.data_files[data_file] = read_time_array(data_file, component_name)
+    normalize = component isa LoadZones ? true : false
```
daniel-thom: This is not good.  I'm not sure how to provide this customization.  I don't think it should go into timeseries_pointers.csv, as that describes the raw data, not what we should do with it.  There is no LoadZones config parameter because there is no LoadZones raw data.  Any opinion @claytonpbarrows

daniel-thom: Per discussion with Clayton I pulled out the normalization code.  It is now up to the user to supply timeseries data that is already normalized if that is what they want.

daniel-thom:
I think this is now ready for merging after you guys approve.  @claytonpbarrows  @jd-lara

src/descriptors/power_system_inputs.json
```diff
@@ -4,6 +4,15 @@
             "name": "simulation",
             "description": "Simulation name"
         },
+        {
```
jd-lara: These definitions are very tailored to the CDM data format. I am not sure if we could say this file is general anymore.

src/parsers/cdm_parser.jl
```diff
@@ -452,11 +432,6 @@ Add generators to the System from the raw data.
 
 """
 function gen_csv_parser!(sys::System, data::PowerSystemRaw)
-    pu_cols = ("active_power_limits_min", "active_power_limits_max",
```
jd-lara: This is a great improvement to avoid hard coded values

src/parsers/cdm_parser.jl
```diff
@@ -752,30 +724,90 @@ function make_storage(data::PowerSystemRaw, gen, bus)
...
 
-const CATEGORY_TO_TYPE = Dict{String, DataType}(
+"""Return the forecast label and whether to convert to per_unit from the descriptor."""
+function _get_label_info!(label_cache::Dict, data::PowerSystemRaw, component_type, forecast)
+    if forecast.label_source == "Category"
+        if component_type <: Generator
+            category = GENERATOR::InputCategory
+        elseif component_type <: Service
+            category = RESERVES::InputCategory
+        elseif component_type <: Bus
```
jd-lara: We shouldn't have forecasts for buses.

src/parsers/cdm_parser.jl
```diff
@@ -752,30 +724,90 @@ function make_storage(data::PowerSystemRaw, gen, bus)
...
+        if descriptor["name"] == name
+            return descriptor
+        end
+    end
+
+    error("Failed to find system descriptor category=$category name=$name")
+end
+
+const CATEGORY_STR_TO_COMPONENT = Dict{String, DataType}(
+    "Bus" => Bus,
```
jd-lara: Same here, this is pretty confusing.

src/parsers/cdm_parser.jl
daniel-thom: It's a bit confusing.  The timeseries_pointers.csv file indicates that LoadZone timeseries data comes from the MW Load column of a bus.

I'm open to walking through this in person.

jd-lara: I guess this is ok as long as we are clear that this applied to the CDM data format only. I am begging to worry we are making a lot of code generalizations using the specific CDM data as the bases.

daniel-thom: Keep in mind that this code is a CDM-specific implementation for parsing raw timeseries data into PowerSystems forecasts.  The code in forecast_parser.jl is supposed to apply to all data sources.

Issue #259, opened 2019-06-17
Review all structs. 


jd-lara:
related to #257 #252 #254

jd-lara:
Add PrimeMover to all Generation and Storage structs and Fuels to thermal as well.

PR #267, opened 2019-07-08
Update structs

jd-lara:
This PR partially addresses #259 but will fix current broken tests


src/parsers/pm2ps_parser.jl
```diff
@@ -370,7 +370,7 @@ function make_line(name, d, bus_f, bus_t)
     return Line(;
         name=name,
         available=Bool(d["br_status"]),
-        connectionpoints=(from=bus_f,to=bus_t),
+        arch=Arch(bus_f,bus_t),
```
daniel-thom: Spaces between commas and variables

src/PowerSystems.jl
```diff
@@ -30,8 +30,8 @@ export PhaseShiftingTransformer
 
 export Forecast
 export Deterministic
-export Scenarios
-export Probabilistic
+#export Scenarios
```
daniel-thom: Should we just delete these?  The structs are no longer defined.

Issue #256, opened 2019-06-13
ybus_calculations uses bus numbers as array indices


daniel-thom:
The code in ybus_calculations.jl causes runtime exceptions because it uses bus number as array indices.  The PR #255 fixed an issue where parsing matpower data caused the bus index to get stored as the bus number. Once I fixed that a test in test/network_matrices.jl started failing.

The code should create an array of bus indices to use with the Ybus.

PR #268, opened 2019-07-09
Cb/fix ybus

### LINKED ISSUE #253: PowerSystems RTS-GMLC parsing error for PTDF matrix
In PowerSystems when I build the RTS system with a ptdf matrix the function
buildptdf throws an error at this line 45
slacks = [num_bus[n.number] for n in nodes if n.bustype == "SF"]
The reason is that there is no bus in the RTS of type SF
There is a bus of type Ref
If I change the line of code to be
slacks = [num_bus[n.number] for n in nodes if n.bustype == "Ref"]
the code runs for the RTS.
Is this the correct fix or is the Ref bus in the RTS different then the Slack bus SF for the system?
I think the RTS is parsing the slack bus as type Ref when it should parse it as type SF

claytonpbarrows:
addresses #256 

 - [x] needs a Ybus5nonsequential_matpower for testing
 - [x] do the same for LODF

closes #253


src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -0,0 +1,200 @@
+
+struct Ybus <: PowerNetworkMatrix
+    data::SparseArrays.SparseMatrixCSC{Complex{Float64},Int64}
+    axes::NTuple{2,Array}
+    lookup::NTuple{2,Dict}
+end
+
+function ybus!(ybus::SparseArrays.SparseMatrixCSC{Complex{Float64},Int64},
```
jd-lara: make `_ybus!`

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -0,0 +1,200 @@
...
+    ybus[b.transformer.connectionpoints.to.number,
+        b.transformer.connectionpoints.from.number] += Y12;
+    Y22 = (y_a + y * (1 - c)) ;
+    ybus[b.transformer.connectionpoints.to.number,
+        b.transformer.connectionpoints.to.number] += Y22;
+
+end
+=#
+
+# Old Ybus creation
```
jd-lara: We can delete this

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -0,0 +1,200 @@
...
+end
+
+function Ybus(branches::Array{T}, nodes::Array{Bus}) where {T<:ACBranch}
+
+    #Get axis names
+    bus_ax = [get_number(bus) for bus in nodes]
+    axes = (bus_ax, bus_ax)
+    look_up = (_make_ax_ref(bus_ax),_make_ax_ref(bus_ax))
+
+    ybus, A = _buildybus(branches, nodes)
```
jd-lara: We don't need to return A here

src/utils/network_calculations/ptdf_calculations.jl
```diff
@@ -0,0 +1,89 @@
...
+    elseif length(slack_position) == 0
+        @warn("Slack bus not identified in the Bus/Nodes list, can't build PTDF")
+        S = Array{Float64,2}(undef,linecount,buscount)
+    end
+
+    return  S, A
+
+end
+
+function PTDF(branches::Array{T}, nodes::Array{Bus}, dist_slack::Array{Float64}=[0.1]) where {T<:ACBranch}
```
jd-lara: We should add a `PTDF(sys::System)` so we don't need to get the branches and buses to make the PTDF

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -0,0 +1,200 @@
...
+        ybus!(ybus, b, num_bus)
+
+
+    end
+
+    return  ybus, A
+
+end
+
+function Ybus(branches::Array{T}, nodes::Array{Bus}) where {T<:ACBranch}
```
jd-lara: We can also make a `Ybus(sys::System)`

claytonpbarrows:
@jd-lara the tests pass on this PR locally, but Travis is broken due to the powersystemsdata changes.

jd-lara:
@claytonpbarrows you need to do a rebase to address the conflicts as well.

PR #269, opened 2019-07-09
Cost functions update

jd-lara:
Support for generic cost functions


src/models/operational_cost.jl
```diff
@@ -0,0 +1,10 @@
+const var_cost_args = Union{Float64, NTuple{2,Float64}, Vector{NTuple{2,Float64}}}
```
daniel-thom: VarCostArgs?

src/models/serialization.jl
```diff
@@ -91,25 +91,24 @@ function encode_for_json(uuid::Base.UUID)
     return (value=string(uuid),)
 end
 
-"""Enables deserialization of EconThermal. The default implementation can't figure out the
-variablecost Union.
+"""Enables deserialization of ThreePartCost. The default implementation can't figure out the
```
daniel-thom: VariableCost instead of ThreePartCost

PR #271, opened 2019-07-10
Auto gen change

jd-lara:


.gitignore
```diff
@@ -16,3 +16,4 @@ data
 *.bak
 *~
 *.vscode
+.continuous-integration/tmp/
```
daniel-thom: Why is your editor doing this?

jd-lara: I added this myself since it risks uploading the tmp folder if someone runs the checks locally.

.travis.yml
```diff
@@ -26,8 +26,12 @@ matrix:
 #  - if [ $TRAVIS_OS_NAME = osx ]; then brew install gcc; fi
 
 ## uncomment the following lines to override the default test script
+before_script:
+  - cd bin
+  - julia --project --color=yes -e 'import Pkg; Pkg.instantiate(); include("../.continuous-integration/check_generated_structs.jl")'
```
daniel-thom: It seems odd to include the script like this.  Can we put the project logic in the script?

jd-lara: This will start the Julia session with the correct Manifest. I prefer this syntax since it separates the package manager aspects from the script itself.

.gitignore
daniel-thom: I meant the fact that there is now no newline at the end of the file.  Actually, the directory is guaranteed to be deleted by the script.  Did that not happen?

jd-lara: If the script fails for some reason the directory is not deleted.

daniel-thom: Ah, the call to read is not in the try block.  Can you move it there?  Then deletion is guaranteed.

daniel-thom: I could go either way.  The script to generate the output should never fail.

jd-lara: see daf3038. I think it doesn't hurt to leave it here

bin/Manifest.toml
```diff
@@ -0,0 +1,98 @@
+# This file is machine-generated - editing it directly is not advised
```
daniel-thom: Why is this file in bin?

.travis.yml
daniel-thom: Can you remove the "cd bin" and then just run the command from the root of the repo?

PR #273, opened 2019-07-12
Added function to remove a component from the system.

daniel-thom:


src/base.jl
```diff
@@ -280,6 +280,27 @@ function add_component!(sys::System, component::T) where T <: Component
...
 end
 
+"""
+    remove_component!(sys::System, component::T) where T <: Component
+
+Remove a component from the system.
+
+Throws InvalidParameter if the component is not stored.
+"""
+function remove_component!(sys::System, component::T) where T <: Component
```
jd-lara: So to remove the component this will require running a `get_component` first and then select which to remove

daniel-thom: We could add another method that allows you to specify a name.  But if you have the name then you should have the component.  Another option could be to remove all components of type T.

jd-lara: We can also add a remove components that get a device type

daniel-thom: New methods are now present.

PR #276, opened 2019-07-15
Probabilistic forecast

jd-lara:
Adds capability to use probabilistic forecasts.


src/base.jl
```diff
@@ -344,6 +344,39 @@ function add_forecasts!(sys::System, forecasts)
     _add_forecasts!(sys.forecasts, forecasts)
 end
 
+"""
+    add_forecast!(sys::System, forecasts)
```
daniel-thom: This is essentially a copy of add_forecasts!().  Do we need it?  The user can call add_forecasts!([forecast]) instead of add_forecast!(forecast).  The latter is a bit nicer.  If you want do want to keep it then I suggest changing add_forecasts so that it calls add_forecast and does not contain duplicate code.

src/descriptors/power_system_structs.json
```diff
@@ -1168,5 +1168,45 @@
...
+      },
+      {
+        "name": "quantiles",
+        "data_type": "Vector{Float64}",
+        "comment": "Quantiles for the probabilistic forecast"
+      },
+      {
+        "name": "data",
+        "data_type": "TimeSeries.TimeArray",
+        "comment": "TimeStamp - scalingfactor"
```
daniel-thom: nit: I think the most common usage of 'timestamp' is as one word.

src/models/forecasts.jl
```diff
@@ -4,8 +4,8 @@ const Forecasts = Vector{<:Forecast}
 const ForecastComponentLabelPair = Tuple{<:Component, String}
 const ForecastComponentLabelPairByInitialTime = Dict{Dates.DateTime,
                                                      Set{ForecastComponentLabelPair}}
-const UNITIALIZED_DATETIME = Dates.DateTime(Dates.Second(0))
-const UNITIALIZED_PERIOD = Dates.Period(Dates.Second(0))
+const UNITIALIZED_DATETIME = Dates.DateTime(Dates.Minute(0))
```
daniel-thom: Aren't these effectively the same values as before?

src/models/supplemental_constructors.jl
```diff
@@ -46,7 +46,40 @@ function Deterministic(component::Component,
...
+
+"""Constructs Probabilistic Forecast after constructing a TimeArray from initial_time and time_steps.
+"""
+function Probabilistic(component::Component,
+                       label::String,
+                       quantiles::Vector{Float64},  # Quantiles for the probabilistic forecast
+                       data::TimeSeries.TimeArray)
+
+    if !(length(TimeSeries.colnames(data)) == length(quantiles))
+        error("The size of the provided quantiles and data columns is incosistent")
```
daniel-thom: typo:  inconsistent

Also, we should have a discussion about consistency when using exceptions.  I've been using our own exception or Julia exceptions.  One of the Base exceptions might be more appropriate here.  There is a debate here:  https://discourse.julialang.org/t/please-stop-using-error-and-errorexception-in-packages-and-base/12096

jd-lara: Ok, I can use your exception. I wasn't aware of this debate.

src/models/forecasts.jl
jd-lara: No, this forced the System to store the resolution in Minutes, beforehand it was storing in seconds.

src/base.jl
jd-lara: see 96572e8

src/models/supplemental_constructors.jl
jd-lara: see 2a458fa

Issue #274, opened 2019-07-15
parsing non-deterministic forecast from JSON broken


jd-lara:
Line 
https://github.com/NREL/PowerSystems.jl/blob/f5ac685444741cff2b35a94998a296e3dfd36144/src/models/forecasts.jl#L249

Requires `issue_time` to parse the forecasts. However, when parsing from user defined forecasts `issue_time` is not defined.

jd-lara:
Changing to 
```
push!(forecasts.data[key], convert_type(forecast_type))
```

allows the code to continue but still throws an error. 

```
Stacktrace:
 [1] convert_type!(::PowerSystems.SystemForecasts, ::NamedTuple{(:data, :initial_time, :resolution, :horizon, :interval),Tuple{NamedTuple{(Symbol("PowerSystems.ForecastKey(2019-07-15T17:00:00, Probabilistic{RenewableDispatch})"), Symbol("PowerSystems.ForecastKey(0001-01-01T00:00:00, Deterministic{PowerLoad})")),Tuple{Array{Any,1},Array{Any,1}}},String,NamedTuple{(:value, :unit),Tuple{Int64,String}},Int64,NamedTuple{(:value, :unit),Tuple{Int64,String}}}}, ::PowerSystems.LazyDictFromIterator{Base.UUID,Component}) at /Users/jdlara/.julia/dev/PowerSystems/src/models/forecasts.jl:249
 [2] from_json(::IOStream, ::Type{System}) at /Users/jdlara/.julia/dev/PowerSystems/src/base.jl:256
 [3] #308 at /Users/jdlara/.julia/dev/PowerSystems/src/models/serialization.jl:24 [inlined]
 [4] #open#310(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::getfield(PowerSystems, Symbol("##308#309")){System}, ::String) at ./iostream.jl:375
 [5] open(::Function, ::String) at ./iostream.jl:373
 [6] from_json(::Type{System}, ::String) at /Users/jdlara/.julia/dev/PowerSystems/src/models/serialization.jl:23
 [7] System(::String) at /Users/jdlara/.julia/dev/PowerSystems/src/base.jl:144
 [8] top-level scope at In[23]:1
```

PR #279, opened 2019-07-15
Fix serialization of Probabilistic forecasts.

daniel-thom:
This PR fixes issue #274

### DIFF
--- a/src/models/serialization.jl
+++ b/src/models/serialization.jl
@@ -37,10 +37,22 @@ Deserialization can't determine the actual concrete type.
 function JSON2.read(io::IO, ::Type{T}) where {T <: TimeSeries.TimeArray}
     data = JSON2.read(io)
     timestamp = [Dates.DateTime(x) for x in data.timestamp]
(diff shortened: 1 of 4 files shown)


src/models/serialization.jl
```diff
@@ -195,8 +207,8 @@ function convert_type(
     return T{component_type}(values...)
 end
 
-function convert_type(::Type{T}, data::Any) where T <: Deterministic
-    error("This form of convert_type is not supported for Deterministic")
+function convert_type(::Type{T}, data::Any) where T <: Forecast
```
jd-lara: Does this error imply that if a user uses a custom forecast object then it needs to also define the custom `convert_type`?

daniel-thom: No.  I defined the function ```convert_type(::Type{T}, data::Any) where T``` in src/utils/utils.jl.  It will convert any type except for a Forecast or Service.  If a caller wants to convert one of those then they have to call the version specific to those types, which take other parameters.

This is only necessary because the functions have different parameter lists.

PR #286, opened 2019-07-18
add AER aggreated Fuel Codes

jd-lara:
These are just the categories

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -50,20 +50,20 @@ end
 
 "AER Aggregated Fuel Code From https://www.eia.gov/survey/form/eia_923/instructions.pdf"
 @enum ThermalFuels begin
-    COL	#Anthracite Coal and Bituminous Coal
-    WOC	#Waste/Other Coal (includes anthracite culm, gob, fine coal, lignite waste, waste coal)
-    DFO	#Distillate Fuel Oil (Diesel, No. 1, No. 2, and No. 4
-    WOO	#Waste Oil Kerosene and JetFuel Butane, Propane,
-    PC  #Petroleum Coke
-    RFO	#Residual Fuel Oil (No. 5, No. 6 Fuel Oils, and Bunker Oil)
-    NG	#Natural Gas
-    OOG	#Other Gas and blast furnace gas
-    NUC	#Nuclear Fission (Uranium, Plutonium, Thorium)
-    ORW	#Agricultural Crop Byproducts/Straw/Energy Crops
-    MLG	#Municipal Solid Waste – Biogenic component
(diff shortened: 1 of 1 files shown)


src/common.jl
```diff
@@ -48,6 +48,24 @@ end
...
 
+"AER Aggregated Fuel Code From https://www.eia.gov/survey/form/eia_923/instructions.pdf"
+@enum ThermalFuels begin
+    COL	#Anthracite Coal and Bituminous Coal
+    WOC	#Waste/Other Coal (includes anthracite culm, gob, fine coal, lignite waste, waste coal)
+    DFO	#Distillate Fuel Oil (Diesel, No. 1, No. 2, and No. 4
+    WOO	#Waste Oil Kerosene and JetFuel Butane, Propane,
+    PC  #Petroleum Coke
+    RFO	#Residual Fuel Oil (No. 5, No. 6 Fuel Oils, and Bunker Oil)
+    NG	#Natural Gas
```
daniel-thom: Can we use full words?  NATURAL_GAS

claytonpbarrows: These are the codes that are published by the Energy Information Agency and in many cases, the data will be in this format. I hate the codes because they are somewhat cryptic, but I'm not entirely opposed since they are standardized. Would full words get too verbose? `RESIDUAL_FUEL_OIL`...

jd-lara: Clayton is right, these are the codes used in the data originally.

daniel-thom: That doesn't preclude defining a mapping of cryptic code to PowerSystems enum in the parsing code.  I suggest that it's better to be able to understand the main PowerSystems code without having to look up the comments.

jd-lara: I added a different naming scheme that is more detailed.

src/common.jl
```diff
@@ -48,6 +48,24 @@ end
     WS #Wind Turbine, Offshore
 end
 
+"AER Aggregated Fuel Code From https://www.eia.gov/survey/form/eia_923/instructions.pdf"
+@enum ThermalFuels begin
+    COAL #COL	#Anthracite Coal and Bituminous Coal
+    WASTE_COAL #WOC	#Waste/Other Coal (includes anthracite culm, gob, fine coal, lignite waste, waste coal)
+    Distillate_Fuel_Oil #DFO #Distillate Fuel Oil (Diesel, No. 1, No. 2, and No. 4
```
daniel-thom: Needs capitalization

src/common.jl
daniel-thom: Looks good.  You might as well define the map of code-to-enum now.

PR #296, opened 2019-08-02
Automate split forecasts

claytonpbarrows:


src/PowerSystems.jl
```diff
@@ -80,6 +80,7 @@ export parse_file
 export add_forecasts!
 export add_forecast!
 export remove_forecast!
+export remove_all_forecasts!
```
jd-lara: I would call this function `clear_forecasts!`

src/parsers/forecast_parser.jl
```diff
@@ -293,3 +293,42 @@ function parse_json(filename,device_names)
...
 =#
+
+"""
+    remove_all_forecasts(sys::System, T::Type{T})::Vector{T} where T <: Forecast
+
+Remove all forecast objects of concrete type T from a System
+"""
+function remove_all_forecasts!(sys::System, ForecastType::Type{T})where T <: Forecast
+
+    ts = get_forecast_initial_times(sys)
```
jd-lara: Do we need to iterate to remove all the forecasts? Can't we just overwrite the data with an empty dict?

src/parsers/forecast_parser.jl
```diff
@@ -293,3 +293,42 @@ function parse_json(filename,device_names)
...
+function split_forecasts!(sys::System, 
+                            forecast::FlattenIteratorWrapper{T}, 
+                            interval::Dates.Period, 
+                            horizon::Int) where T <: Forecast
+
+    forecasts = make_forecasts(forecast, interval, horizon)
+
+    remove_all_forecasts!(sys, T)
+
+    add_forecasts!(sys,forecasts)
```
jd-lara: missing return

src/parsers/forecast_parser.jl
```diff
@@ -293,3 +293,42 @@ function parse_json(filename,device_names)
...
+Remove all forecast objects of concrete type T from a System
+"""
+function remove_all_forecasts!(sys::System, ForecastType::Type{T})where T <: Forecast
+
+    ts = get_forecast_initial_times(sys)
+
+    for t in ts
+        f = get_forecasts(ForecastType, sys, t)
+        foreach(x -> remove_forecast!(sys,x),f)
+    end
```
jd-lara: missing return

src/parsers/forecast_parser.jl
daniel-thom: Even better:  ```empty!(sys.forecasts.data)```

src/parsers/forecast_parser.jl
daniel-thom: Do we need that on a function that doesn't return anything?

src/parsers/forecast_parser.jl
```diff
@@ -293,3 +293,42 @@ function parse_json(filename,device_names)
...
+
+end
+
+"""
+    split_forecasts!(sys::System, 
+                    forecast::FlattenIteratorWrapper{T}, 
+                    interval::Dates.Period, 
+                    horizon::Int) where T <: Forecast
+
+Replaces system forecasts with ke a vector of forecasts by incrementing through a forecast
```
daniel-thom: nit: ke?

src/parsers/forecast_parser.jl
```diff
@@ -293,3 +293,42 @@ function parse_json(filename,device_names)
     return Devices
 end
 =#
+
+"""
+    remove_all_forecasts(sys::System, T::Type{T})::Vector{T} where T <: Forecast
+
+Remove all forecast objects of concrete type T from a System
+"""
+function remove_all_forecasts!(sys::System, ForecastType::Type{T})where T <: Forecast
```
daniel-thom: I thought we wanted to remove all forecasts regardless of type.  Is that not true?

src/parsers/forecast_parser.jl
```diff
@@ -293,3 +293,42 @@ function parse_json(filename,device_names)
...
+                    forecast::FlattenIteratorWrapper{T}, 
+                    interval::Dates.Period, 
+                    horizon::Int) where T <: Forecast
+
+Replaces system forecasts with ke a vector of forecasts by incrementing through a forecast 
+by interval and horizon.
+
+"""
+function split_forecasts!(sys::System, 
+                            forecast::FlattenIteratorWrapper{T},
```
daniel-thom: nit:  alignment

src/parsers/forecast_parser.jl
```diff
@@ -293,3 +293,42 @@ function parse_json(filename,device_names)
...
+    for t in ts
+        f = get_forecasts(ForecastType, sys, t)
+        foreach(x -> remove_forecast!(sys,x),f)
+    end
+
+end
+
+"""
+    split_forecasts!(sys::System, 
+                    forecast::FlattenIteratorWrapper{T},
```
daniel-thom: Enforcing FlattenIteratorWrapper may be too rigid. I could imagine a user calling this with a vector.  I suggest leaving this untyped and put a comment saying that it needs to be an iterable.  I think Julia will eventually add a way to specify an interface type for this case.

Also, forecast should probably be plural.

src/models/forecasts.jl
```diff
@@ -411,3 +411,18 @@ function make_forecasts(forecast::Forecast, interval::Dates.Period, horizon::Int
...
+                    
+Make a vector of forecasts by incrementing through a vector of forecasts 
+by interval and horizon.
+"""
+function make_forecasts(forecast::FlattenIteratorWrapper{T}, 
+                        interval::Dates.Period, horizon::Int) where T <: Forecast
+
+    forecasts = [make_forecasts(f, interval, horizon) for f in forecast]
+
+    return vcat(forecasts...)
```
daniel-thom: Does this need to be a vector?  You could avoid extra memory allocations if you used FlattenIteratorWrapper.

src/parsers/forecast_parser.jl
jd-lara: http://www.juliaopt.org/JuMP.jl/dev/style/#Return-statements-1

daniel-thom: Good point.  I forgot about that.  I don't love that rule and have been violating it...  :(

You would never do that in C or C++ unless you return early.

src/parsers/forecast_parser.jl
claytonpbarrows: I made this slightly more specific so that we can remove all forecasts of a particular type. If that's not a valid use case, I'll replace this with @daniel-thom's suggestion. Otherwise, I'll implement the suggestion in another method that just takes the `sys` argument.

src/parsers/forecast_parser.jl
claytonpbarrows: based on my tests, a system can have forecasts of different types that are inconsistent. Therefore, the need to remove forecasts from the system before adding split forecasts only applies to forecasts of the same type.

PR #298, opened 2019-08-03
Jd/new accessors

jd-lara:
Add new forecast retrieval methods.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -699,9 +699,12 @@ end
                             initial_time::Dates.DateTime,
                             ) where T <: Component
 
-Get the forecasts of a component of type T with initial_time. The resulting FlattenIteratorWrapper can contain Forecasts of dissimilar types.
+Get the forecasts of a component of type T with initial_time.
+    The resulting container can contain Forecasts of dissimilar types.
 
-Throws InvalidParameter if T is not an abstract type.
+Throws InvalidParameter if T is not a concrete type.
+
+See also: [`get_component`](@ref)
 """
 function get_component_forecasts(
                                  ::Type{T},
@@ -710,7 +713,7 @@ function get_component_forecasts(
                                 ) where T <: Component
 
     if !isconcretetype(T)
-        throw(InvalidParameter("get_component only supports concrete types: $T"))
+        throw(InvalidParameter("get_component_forecasts only supports concrete types: $T"))
     end
 
     return (f for k in keys(sys.forecasts.data) if k.initial_time == initial_time
--- a/src/base.jl
+++ b/src/base.jl
@@ -709,8 +709,8 @@ function get_component_forecasts(
                                  initial_time::Dates.DateTime,
                                 ) where T <: Component
 
-    if !isabstracttype(T)
-        throw(InvalidParameter("get_components_by_name only supports abstract types: $T"))
+    if !isconcretetype(T)
+        throw(InvalidParameter("get_component only supports concrete types: $T"))
     end
 
     return (f for k in keys(sys.forecasts.data) if k.initial_time == initial_time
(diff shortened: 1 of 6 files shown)


src/base.jl
```diff
@@ -557,6 +557,72 @@ function get_forecasts(
...
+# Examples
+```julia
+iter = PowerSystems.get_forecasts(RenewableFix, sys, initial_time)
+forecasts = PowerSystems.get_forecasts(RenewableFix, sys, initial_time) |> collect
+forecasts = collect(PowerSystems.get_forecasts(RenewableFix, sys))
+```
+
+See also: [`iterate_forecasts`](@ref)
+"""
+function get_component_forecasts(
```
daniel-thom: Here is a one-liner that could replace get_component_forecasts and get_components_forecasts:

```suggestion
function get_component_forecasts(
                                  ::Type{T},
                                 sys::System,
                                 initial_time::Dates.DateTime,
                                ) where T <: Component
    return (f for k in keys(sys.forecasts.data) if k.initial_time == initial_time
              for f in sys.forecasts.data[k] if get_component(f) isa T)
end
```

src/base.jl
```diff
@@ -657,6 +692,26 @@ function get_components_by_name(
     return components
 end
 
+"""
+    get_components_forecasts(
```
daniel-thom: nit:  Re-copy the function declaration on top of the docstring.  The docstring function has an extra 's' and the alignment is off by one space.

src/base.jl
```diff
@@ -657,6 +692,26 @@ function get_components_by_name(
...
+                            ::Type{T},
+                            sys::System,
+                            initial_time::Dates.DateTime,
+                            ) where T <: Component
+
+Get the forecasts of a component of type T with initial_time.
+
+Throws InvalidParameter if T is not an abstract type.
```
daniel-thom: This is no longer true.

src/base.jl
```diff
@@ -657,6 +692,26 @@ function get_components_by_name(
...
 end
 
+"""
+    get_components_forecasts(
+                            ::Type{T},
+                            sys::System,
+                            initial_time::Dates.DateTime,
+                            ) where T <: Component
+
+Get the forecasts of a component of type T with initial_time.
```
daniel-thom: Questions:
1. Should the docstring call out the fact that you will get back forecasts of all types?
2. Should there be an arg or kwarg that allows the user to specify a specific forecast type?  You could default it to Forecast.

jd-lara: I am working on this, I prefer to include the error if T is abstract. I think it's fine to return forescasts of different types.

src/base.jl
jd-lara: There is a problem with this implementation. The return type is too complicated to do dispatch on the forecast type afterwards.

jd-lara: I decided to do a collect call in PowerSimulations, this only happens once so it will be ok. If it becomes a bottle neck ir can be tackled later.

src/base.jl
```diff
@@ -693,13 +693,13 @@ function get_components_by_name(
...
+    get_component_forecasts(
                             ::Type{T},
                             sys::System,
                             initial_time::Dates.DateTime,
                             ) where T <: Component
 
-Get the forecasts of a component of type T with initial_time.
+Get the forecasts of a component of type T with initial_time. The resulting FlattenIteratorWrapper can contain Forecasts of dissimilar types.
 
 Throws InvalidParameter if T is not an abstract type.
```
daniel-thom: is not a concrete type

src/base.jl
```diff
@@ -708,6 +708,11 @@ function get_component_forecasts(
                                  sys::System,
                                  initial_time::Dates.DateTime,
                                 ) where T <: Component
+
+    if !isconcretetype(T)
+        throw(InvalidParameter("get_component only supports concrete types: $T"))
```
daniel-thom: get_component_forecasts

PR #300, opened 2019-08-06
fixing matpower bustype parsing, closes #299

### LINKED ISSUE #299: matpower parsing gets bustype wrong
in case5.m, bus 1 is a "PV" bus, but PSY assigns bus "1" as a "PQ" bus,

claytonpbarrows:


src/parsers/pm2ps_parser.jl
```diff
@@ -55,7 +55,7 @@ end
 function read_bus!(sys::System, data, )
     @info "Reading bus data"
     bus_number_to_bus = Dict{Int, Bus}()
-    bus_types = ["PV", "PQ", "REF","isolated"]
+    bus_types = ["PQ", "PV", "REF","isolated"]
```
daniel-thom: This feels a bit brittle because it relies on Julia array indices starting with 1. The fact the order has been wrong for a long time sorta proves the point.  Here is a suggestion:

 ```suggestion
"From http://www.pserc.cornell.edu/matpower/MATPOWER-manual.pdf Table B-1"
@enum MatpowerBusType begin
     PQ = 1
     PV = 2
     REF = 3
     ISOLATED = 4
end

bus_types = instances(MatpowerBusType)
```

jd-lara:
For some reason this PR seems to have broken the Ybus calculations.

jd-lara:
@claytonpbarrows you can merge this PR and then see if it addresses the failing tests. If they fail, we can address them with another PR into dev

PR #314, opened 2019-08-29
Change make_pf implementation

kdheepak:


src/utils/power_flow/make_pf.jl
```diff
@@ -41,124 +31,106 @@ res = NLsolve.nlsolve(pf!, x0)
...
-        Q_bal = Vector{Float64}(undef, $bus_count)
-    end
+    # Use vectors to cache data for closure
+    # These should be read only
+    P_GEN_BUS  = fill(0.0, N_BUS)
+    Q_GEN_BUS  = fill(0.0, N_BUS)
+    P_LOAD_BUS = fill(0.0, N_BUS)
+    Q_LOAD_BUS = fill(0.0, N_BUS)
+
+    BUSES = [b for b in enumerate(buses)]
```
jd-lara: There is no need to allocate a matrix here you can do  `BUSES = (b for b in enumerate(buses))` to use a generator or simple do enumerate buses below

src/utils/power_flow/make_pf.jl
```diff
@@ -41,124 +31,106 @@ res = NLsolve.nlsolve(pf!, x0)
...
 
-    var_count = 1
-    res_dict = Dict{String, Vector{Tuple{Symbol, Int}}}()
-    x0 = Array{Float64}(undef, bus_count*2)
+    state_variable_count = 1
```
jd-lara: Start at 0

src/utils/power_flow/make_pf.jl
```diff
@@ -41,124 +31,106 @@ res = NLsolve.nlsolve(pf!, x0)
...
-            # Reference for the results Dict
-            res_dict[b.name] = [var_ref1, var_ref2]
+            x0[state_variable_count]     = bus_voltage
+            x0[state_variable_count + 1] = bus_angle
+            state_variable_count += 2
         end
     end
 
-    balance_eqs = quote  end
+    @assert state_variable_count - 1 == N_BUS * 2
```
jd-lara: `@assert state_variable_count == N_BUS * 2`

src/utils/power_flow/power_flow.jl
```diff
@@ -133,33 +133,15 @@ Arguments available for `nlsolve`:
...
-            pf!, x0 = PowerSystems.make_pf($sys)
-            res = NLsolve.nlsolve(pf!, x0; $par)
-            show(res)
-            PowerSystems._write_pf_sol!($sys, res)
-    end)
-    return eval_code
+function solve_powerflow!(sys, args...)
+    pf!, x0 = PowerSystems.make_pf(sys)
+    res = NLsolve.nlsolve(pf!, x0; args...)
+    PowerSystems._write_pf_sol!(sys, res)
```
jd-lara: Missing return

src/utils/power_flow/power_flow.jl
```diff
@@ -95,16 +64,34 @@ function _update_branch_flow!(sys::System)
...
+                generator.reactivepower = Q_gen
+            end
+            bus.angle = θ
         elseif bus.bustype == PowerSystems.PV
-            _update_PV_bus(bus, sys, result, v)
+            Vm = result[2 * ix - 1]
+            θ = result[2 * ix]
+            bus.voltage = Vm
+            bus.angle = θ
         end
```
jd-lara: Missing Return

src/utils/power_flow/make_pf.jl
kdheepak: I didn't think we could use a generator because it is being consumed more than once. But it turns out you can, so I'll change it.

src/utils/power_flow/make_pf.jl
kdheepak: I'm using `state_variable_count` as an index, so I can't start at zero.

src/utils/power_flow/power_flow.jl
kdheepak: What should we return here? The `sys` object or the `res`?

src/utils/power_flow/make_pf.jl
jd-lara: Dan showed me some significant improvements in Julia in many cases that we use list comprehensions and we really just need generators.

src/utils/power_flow/power_flow.jl
kdheepak: What would you like to return here?

src/utils/power_flow/make_pf.jl
kdheepak: I think the easy readability of the rest of the code above this makes up for the awkwardness of this one line. Thoughts?

src/utils/power_flow/power_flow.jl
jd-lara: nothing, just return.

src/utils/power_flow/make_pf.jl
jd-lara: Maybe just leave a comment of why its there.

src/utils/power_flow/power_flow.jl
kdheepak: Resolved in 3ea1ff4

src/utils/power_flow/power_flow.jl
kdheepak: Resolved in 3ea1ff4

src/utils/power_flow/make_pf.jl
kdheepak: Resolved in be481bb

src/utils/power_flow/power_flow.jl
```diff
@@ -146,33 +133,16 @@ Arguments available for `nlsolve`:
...
 # Passing NLsolve arguments
-@rsolve_powerflow!(sys, method = :Newton)
+solve_powerflow!(sys, nlsolve, method = :Newton)
 
 ```
 
 """
-
-macro solve_powerflow!(sys, args...)
-    vals = filter((x) ->  x == :NLsolve, names(Main,imported=true))
```
jd-lara: I think we should keep these two lines, makes for a better user experience and it is not a super problematic thing to run

PR #316, opened 2019-08-29
Fix type inference on PowerNetworkMatrix

### LINKED ISSUE #310: Type of data retrieved from Network Matrices is not known during run time
The type of data retrieved from the Ybus is not known by the compiler. If this data is used in following code it results in poor type inference. 

Screen Shot 2019-08-27 at 2 51 28 AM

Related to implementation here:

https://github.com/NREL/PowerSystems.jl/blob/83ee65216c2df3eabc4a3e7c98e67a07029cd236/src/utils/network_calculations/ybus_calculations.jl#L1-L14

kdheepak:
Fixes #310

### DIFF
--- a/src/utils/network_calculations/common.jl
+++ b/src/utils/network_calculations/common.jl
@@ -1,5 +1,5 @@
 
-abstract type PowerNetworkMatrix <: AbstractArray{Float64,2} end
+abstract type PowerNetworkMatrix{T} <: AbstractArray{T,2} end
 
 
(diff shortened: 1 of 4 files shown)


src/utils/network_calculations/common.jl
```diff
@@ -1,5 +1,5 @@
 
-abstract type PowerNetworkMatrix <: AbstractArray{Float64,2} end
+abstract type PowerNetworkMatrix{T} <: AbstractArray{T,2} end
```
jd-lara: Use `AbstractMatrix{T}`

PR #320, opened 2019-08-29
Fix `@info` in documentation generation

kdheepak:


jd-lara:
This change has been incorporated in  f097330

PR #323, opened 2019-09-05
Auto-generate accessor functions for structs.

daniel-thom:
There is an outstanding request to export accessor functions for all auto-generated structs. This does that, but I don't love it.  I'm wondering if (1) it should be done at all or (2) there should be a flag in the JSON that specifies whether to export.  But I'm open.  What do you think?


jd-lara:
I don't love it either but must admit it is an annoyance to use PSY before EACH get function.

daniel-thom:
A point to consider is whether you want users to see these when they type `names(PowerSystems)` in the REPL.  We currently export 86 names. This adds 88 names.

jd-lara:
I don't think the number of exported names is a problem

Issue #325, opened 2019-09-10
Fix print statements 


jd-lara:
Make show statements return Summary by default to improve display in REPL and Notebooks.

claytonpbarrows:
@jjstickel can you help with this since you did it for a previous version?
Basically, we want `print(sys)` to show the output that you get when you run:
```
print(summary(sys))
```

jjstickel:
Sure. Is it urgent? I can probably take a look on Thursday.

daniel-thom:
Don't we won't to override show instead?

```julia
Base.show(io::IO, sys::System) = print(io, summary(sys))
@show sys
```

jd-lara:
Correct, I made the requirement more clear.

PR #328, opened 2019-09-12
fix summary() and printing for current version

jjstickel:
Some work to revamp `summary()` and printing (`print()` and repl output) for System, SystemCompoenents, and SystemForecasts. This to address #325.

I am not up-to-date with the current structure of PowerSystem objects, and so I leave it to others revise this code further to output what is desired for `summary()` and `print()`. Currently, `print()` does the default thing from Base. To use the custom output, uncomment the "Base.show..." lines as called out.

Note that there are also `Base.summary()` methods in utils/network_calculations/common.jl for PowerNetworkMatrix, but I did not modify those. It looks like that code was copied from elsewhere and may be OK as is.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -797,39 +797,6 @@ function get_components(
     return iter
 end
 
-"""Shows the component types and counts in a table."""
-function Base.summary(io::IO, sys::System)
-    println(io, "System")
-    println(io, "======")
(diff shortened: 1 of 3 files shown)


jd-lara:
This has been replaced with #331

PR #331, opened 2019-09-13
Printing enhancements

daniel-thom:
This extends Jonathan's work and makes it compatible with the new InfrastructureSystems module.


daniel-thom:
I apparently forgot to request reviews on this earlier.

@jd-lara Is the documentation build failure expected?

jd-lara:
@daniel-thom yes, it is expected the documentation to fail because IS isn't registered yet

PR #339, opened 2019-09-17
Move files

jd-lara:
This PR moves the .jl files for testing to PowerSystems since we won't support .jl data files anymore.


daniel-thom:
One general suggestion/comment.  In InfrastructureSystems I instituted a rule that says all test files must start with "test_".  The runtests.jl code skips any file that doesn't match.  We could do the same here and you could keep these jl files in /test.

test/test_aux_files/data_14bus_pu.jl
```diff
@@ -0,0 +1,210 @@
+using PowerSystems
```
daniel-thom: This can be deleted.

test/test_aux_files/data_5bus_pu.jl
```diff
@@ -0,0 +1,243 @@
+using PowerSystems
```
daniel-thom: ditto

jd-lara:
@daniel-thom I renamed the files but prefer to keep them as aux files so we can call them for other tests later if needed.

PR #349, opened 2019-09-23
Remove old files

jd-lara:


.gitignore
```diff
@@ -18,3 +18,4 @@ data
 src/descriptors/validation_config.*
 docs/src/man/data_requirements_table.md
 /docs/Manifest.toml
+Manifest.toml
```
daniel-thom: You're editor must still be messed up.

PR #353, opened 2019-10-02
enabling multiple loads per bus via pm parsing

claytonpbarrows:
- enables custom validation descriptors via pm parsing 
 - enables multiple loads per bus via pm parsing

### DIFF
--- a/src/parsers/pm2ps_parser.jl
+++ b/src/parsers/pm2ps_parser.jl
@@ -10,7 +10,7 @@ function pm2ps_dict(data::Dict{String,Any}; kwargs...)
 
     @info "Constructing System from Power Models" data["name"] data["source_type"]
 
-    sys = System(data["baseMVA"])
+    sys = System(data["baseMVA"]; kwargs...)
 
     bus_number_to_bus = read_bus!(sys, data)
     read_loads!(sys, data, bus_number_to_bus)
@@ -94,7 +94,7 @@ end
 
 function make_load(d, bus)
     return PowerLoad(;
-        name=bus.name,
+        name=join(d["source_id"], "_"), #bus.name,
         available=true,
         model = ConstantPower::LoadModel,
         bus=bus,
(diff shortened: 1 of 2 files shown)


claytonpbarrows:
One outstanding issue is that PSS/e supports non-unique bus names. So a user can encounter a situation where buses cannot be created because there is already one with the same name in the System. 
Currently, I'm able to get around it by parsing the PSS/e case into the pm_dict and then creating `b["name"] = string(b["name"]," ", busid)`, but we could automatically do this when parsing through pm. Thoughts?

jd-lara:
This issue with PSS/e is tricky, we can make a key with the name and the index and then that has to be unique.

claytonpbarrows:
Yeah, but should we always (from psse and matpower) do that, or just leave it to the user to do what I did?

daniel-thom:
This could be a very annoying issue for users.  It would be better if they didn't have to know about it.

A downside to this solution is that when you display the bus information the name will not actually be the original name.

I'll suggest another possible solution:  Create a new variable for buses called unique_name which is concatenation of name and number. Use that as the key for storing buses.  We would probably need to write a function called something like get_internal_storage_key.  The default implementation for Component would return the name.  The bus function would return unique_name.

claytonpbarrows:
@daniel-thom most data formats require unique names, so AFAIK PSS/E will be the only format that creates this issue.
But, I'm thinking of another solution. In most of our data, we'd like to preserve name as simply the name, but in some cases, such as the one that I'm encountering, I need to create a unique name. I'm doing that by concatenating the name with the id column. Could that be a user specified option? For example, we'd default to using name, but if a user found that that wasn't unique, they could add an option to specify name+baseKV to create uniqueness... if that didn't work name+id...
Thoughts?

claytonpbarrows:
@jd-lara and @daniel-thom this should be ready now

daniel-thom:
The code looks good.  How about a docstring for the user?

jd-lara:
@claytonpbarrows also update the Changelog, please.

PR #355, opened 2019-10-07
remove data download from build

### LINKED ISSUE #348: run download process of PowerSystems to run-tests
With the new data in the TestData repo, it is taking a long time to download. Considering that the new data is just needed to test we can reconsider the need to even download the data bu default in build and move it to test. Additionally, we can add the capability to obtain the data from the repo as a separate function. 

This can be addressed with #280

jd-lara:
This PR removes the data download from build to testing in order to reduce local building times. 

closes #348


test/runtests.jl
```diff
@@ -9,6 +9,9 @@ using PowerSystems
 import PowerSystems: PowerSystemTableData
 const PSY = PowerSystems
 
+include(joinpath(@__DIR__, "../src/utils/data.jl"))
+import .UtilsData: TestData
+download(TestData; branch = "master")
```
daniel-thom: This is already called about 10 lines down.

jd-lara: removed

Issue #356, opened 2019-10-10
power flow calculations don't support non-uniform bus indexing


claytonpbarrows:
buses must have numbers = 1:N.

PR #358, opened 2019-10-10
Pf indexing

claytonpbarrows:
addresses #356 and #357

### DIFF
--- a/src/utils/power_flow/make_pf.jl
+++ b/src/utils/power_flow/make_pf.jl
@@ -54,14 +54,15 @@ function make_pf(system)
         bus_number = get_number(b)::Int
         bus_angle = get_angle(b)::Float64
         bus_voltage = get_voltage(b)::Float64
-        generator = []
+        P_GEN_BUS[ix] = 0.0
+        Q_GEN_BUS[ix] = 0.0
         for gen in get_components(Generator, system)
             if gen.bus == b
-                push!(generator, gen)
+                P_GEN_BUS[ix] += get_activepower(gen)
+                Q_GEN_BUS[ix] += get_reactivepower(gen)
             end
         end
-        P_GEN_BUS[ix] = isempty(generator) ? 0.0 : sum(get_activepower.(generator))
-        Q_GEN_BUS[ix] = isempty(generator) ? 0.0 : sum(get_reactivepower.(generator))
+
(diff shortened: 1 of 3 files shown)


src/utils/power_flow/make_pf.jl
```diff
@@ -52,15 +54,14 @@ function make_pf(system)
         bus_number = get_number(b)::Int
         bus_angle = get_angle(b)::Float64
         bus_voltage = get_voltage(b)::Float64
-        generator = nothing
+        generator = []
```
daniel-thom: generators?

test/test_powerflow.jl
```diff
@@ -42,5 +42,7 @@ c_sys5_re = System(nodes5, vcat(thermal_generators5, renewable_generators5), loa
     @test res.zero ≈ result rtol=1e-3
 
     solve_powerflow!(c_sys14, nlsolve, method = :newton)
-    @test_throws PowerSystems.DataFormatError solve_powerflow!(c_sys5_re, nlsolve)
+    pf!, x0 = make_pf(c_sys5_re);
+    res = NLsolve.nlsolve(pf!, x0)
+    @test res.f_converged == false
```
daniel-thom: nit: to check a boolean it's typically better style to not include true/false.
`@test !res.f_converged`

src/utils/power_flow/make_pf.jl
jd-lara: Why empty vect?

src/utils/power_flow/make_pf.jl
```diff
@@ -52,15 +54,14 @@ function make_pf(system)
...
         for gen in get_components(Generator, system)
             if gen.bus == b
-                !isnothing(generator) && throw(DataFormatError("There is more than one generator connected to Bus $b.name"))
-                generator = gen
+                push!(generator, gen)
```
jd-lara: This is a very bad operation to had in an iterative algorithm, basically each iteration you will be allocating these matrices.

daniel-thom: If that is a problem, it can be easily solved.  Create two variables to hold sum_active_power and sum_reactive_power, initialized to 0.0.  Instead of pushing to a vector, add each generator's powers.

PR #368, opened 2019-10-25
Tabledata bugfixes

claytonpbarrows:
Resolves several bugs in tabledata parsing
Improves tabledata flexibility
Fixes PM parsing thermal gen PWL cost calculations


src/descriptors/power_system_structs.json
```diff
@@ -342,15 +342,15 @@
...
+        "valid_range": {"min":0.0, "max":4.0},
         "validation_action": "error"
       },
       {
         "null_value": "0.0",
         "name": "x",
         "data_type": "Float64",
         "comment": "System per-unit value",
-        "valid_range": {"min":0.0, "max":2.0},
+        "valid_range": {"min":-1.0, "max":4.0},
```
jd-lara: Negative reactance?

src/descriptors/power_system_structs.json
```diff
@@ -419,15 +419,15 @@
...
```
jd-lara: Negative reactance?

src/parsers/power_system_table_data.jl
```diff
@@ -602,17 +624,26 @@ function services_csv_parser!(sys::System, data::PowerSystemTableData)
             ))
         end
 
-        service = ProportionalReserve(reserve.name,
-                                      contributing_devices,
-                                      reserve.timeframe)
+        if isnothing(requirement)
+            service = ProportionalReserve(reserve.name,
```
jd-lara: contents are not aligned

src/parsers/pm2ps_parser.jl
```diff
@@ -280,7 +280,9 @@ function make_thermal_gen(gen_name::AbstractString, d::Dict, bus::Bus)
             power_p = [i for (ix,i) in enumerate(cost_component) if isodd(ix)]
             cost_p =  [i for (ix,i) in enumerate(cost_component) if iseven(ix)]
             cost = [(p,c) for (p,c) in zip(cost_p,power_p)]
-            fixed = cost[1][2]
+            fixed = max(0.0, cost[1][1] - (cost[2][1] - cost[1][1])/(cost[2][2] - cost[1][2])*cost[1][2])
+            cost = [(c[1] - fixed, c[2]) for c in cost]
+            #fixed = cost[1][2]
```
daniel-thom: stale line?

src/parsers/power_system_table_data.jl
```diff
@@ -550,7 +550,8 @@ function loadzone_csv_parser!(sys::System, data::PowerSystemTableData)
 
         buses = get_buses(sys, bus_numbers)
         name = string(zone)
-        load_zones = LoadZones(zone, name, buses, sum(active_powers), sum(reactive_powers))
+        zoneid = typeof(zone) <: Number ? zone : count
```
daniel-thom: A comment would be good here. What is count?

daniel-thom: Readability improvement:

```suggestion
        zoneid = zone isa Number ? zone : count
```

src/parsers/power_system_table_data.jl
```diff
@@ -566,34 +567,55 @@ function services_csv_parser!(sys::System, data::PowerSystemTableData)
...
-                        # There multiple categories, so we might not find a match in some.
-                        continue
-                    elseif length(components) == 1
-                        component = components[1]
-                    else
-                        msg = "Found duplicate names type=$component_type name=$name"
-                        throw(DataFormatError(msg))
+        if !isnothing(device_subcategories)
+            @info("Adding contributing generators for $(reserve.name) by category")
+            @warn("Adding contributing components by category only supports generators")
```
daniel-thom: It seems like this warning message could cause tons of noise.  Maybe add a maxlog=1?

src/parsers/power_system_table_data.jl
```diff
@@ -566,34 +567,55 @@ function services_csv_parser!(sys::System, data::PowerSystemTableData)
...
+                # There multiple categories, so we might not find a match in some.
+                continue
+            elseif length(components) == 1
+                component = components[1]
+            else
+                msg = "Found duplicate names type=$component_type name=$name"
+                throw(DataFormatError(msg))
+            end
+
+            push!(contributing_devices, component)
```
daniel-thom: Can you get here more than once in one function call?

src/parsers/power_system_table_data.jl
```diff
@@ -668,8 +699,19 @@ function make_thermal_generator(data::PowerSystemTableData, gen, cost_colnames,
     )
 
     capacity = gen.active_power_limits_max
-    startup_cost = gen.startup_heat_cold_cost * fuel_cost * 1000
-    shutdown_cost = 0.0
+    startup_cost = get(gen, :startup_cost, nothing)
+    if isnothing(startup_cost) && hasfield(typeof(gen), :startup_heat_cold_cost)
```
daniel-thom: This means that startup_heat_cold_cost is a required field.  Another way to do this is to define a default function get_startup_heat_cold_cost.  Then, users could extend in their own way.  We should be consistent on whether fields are required.

src/parsers/power_system_table_data.jl
```diff
@@ -845,6 +889,17 @@ function _get_field_infos(data::PowerSystemTableData, category::InputCategory, d
             custom_name = Symbol(item["custom_name"])
             name = item["name"]
             if custom_name in df_names
+                if !(name in descriptor_names)
+                    if (occursin("heat_rate_",name) | occursin("output_percent_",name))
```
daniel-thom: ||

src/parsers/power_system_table_data.jl
```diff
@@ -834,9 +876,11 @@ function _get_field_infos(data::PowerSystemTableData, category::InputCategory, d
     # The user's descriptors indicate that the raw data is already system-per-unit or not.
     per_unit = Dict{String, Bool}()
     unit = Dict{String,Union{String,Nothing}}()
+    descriptor_names = Vector{String}()
     for descriptor in data.descriptors[category]
         per_unit[descriptor["name"]] = get(descriptor, "system_per_unit", false)
         unit[descriptor["name"]] = get(descriptor, "unit", nothing)
+        push!(descriptor_names,descriptor["name"])
```
daniel-thom: Style guide says that there should be a space after comma.  Here and throughout the PR.

src/parsers/power_system_table_data.jl
```diff
@@ -845,6 +889,17 @@ function _get_field_infos(data::PowerSystemTableData, category::InputCategory, d
             custom_name = Symbol(item["custom_name"])
             name = item["name"]
             if custom_name in df_names
+                if !(name in descriptor_names)
+                    if (occursin("heat_rate_",name) | occursin("output_percent_",name))
+                        base = join(split(name, "_")[1:end-1], "_")
```
daniel-thom: I feel like this might be easier to follow if it was a regex.

src/parsers/power_system_table_data.jl
```diff
@@ -917,9 +975,14 @@ file may be customized, so this creates a mapping of (category, custom_name) to
 function _create_forecast_label_mapping(user_descriptors::Dict)
     mapping = Dict{Tuple{String, String}, String}()
     for (category, params) in user_descriptors
+        fields = [d["name"] for d in params]
+        ufields = unique(fields)
+        counts=[(i,count(x->x==i,fields)) for i in ufields if count(x->x==i,fields)>1]
+        if length(counts) < 0
```
daniel-thom: Seems like a bug.  How can a vector have a length < 0?  The exception string dereferences element 1.

src/parsers/power_system_table_data.jl
claytonpbarrows: Agreed. In this effort, I've realized that we have a pretty big mis-alignment on required table columns and required struct fields. I think we need a distinct effort to resolve that

src/parsers/power_system_table_data.jl
daniel-thom: Forget the regex.  I think findlast("_") is really what you want.

Issue #371, opened 2019-10-30
Generate Initial Times broken


claytonpbarrows:
[`generate_initial_times()`](https://github.com/NREL/PowerSystems.jl/blob/2f2feed7edce42a1a23eb5697ccd25f014a3fb73/src/base.jl#L570) fails.
This line:
```julia
    return generate_initial_times(sys.data, interval, horizon)
```
should be 
```julia
    return IS.generate_initial_times(sys.data, interval, horizon)
```

PR #372, opened 2019-10-30
Fixes call to generate_initial_times.

daniel-thom:
Fixes issue #371


jd-lara:
Should we add a test for this functionality, it is pretty critical.

daniel-thom:
There are tests in IS that cover the functionality.  See https://github.com/NREL/InfrastructureSystems.jl/blob/master/test/test_forecasts.jl#L265 and https://github.com/NREL/InfrastructureSystems.jl/blob/master/test/test_forecasts.jl#L302.

I got lazy and didn't repeat the effort in PSY.  I'll add a simple test to make sure that the call works.  It'd be nice if Julia had a static linter.

PR #375, opened 2019-11-04
draft of docstrings for JSON script that makes the structs

Lilyhanig:


src/descriptors/power_system_structs.json
```diff
@@ -1,7 +1,7 @@
 [
   {
     "struct_name": "TwoPartCost",
-    "docstring": "Data Structure Operational Cost Data in two parts fixed and variable cost.",
+    "docstring": "\n    mutable struct TwoPartCost <: OperationalCost\n        variable::VariableCost\n        fixed::Float64\n        _forecasts::InfrastructureSystems.Forecasts\n        internal::InfrastructureSystemsInternal\n    end\n\nData Structure Operational Cost Data in two parts: fixed and variable cost.\n",
```
jd-lara: After generating the files looks like this 

```
    mutable struct TwoPartCost &lt;: OperationalCost
        variable::VariableCost
        fixed::Float64
        _forecasts::InfrastructureSystems.Forecasts
        internal::InfrastructureSystemsInternal
    end

Data Structure Operational Cost Data in two parts: fixed and variable cost.
```

seems that the `<:` symbol isn't being parsed properly. 

@daniel-thom any suggestions?

daniel-thom: This will be fixed in the latest InfrastructureSystems update.

That being said, why are you all adding each field to the docstring?

jd-lara: For the docs

daniel-thom: It seems like this is duplicating all of the text.  If we really want this, we could auto-generate it so the the text isn't duplicated.

Separate from that, I'm not sure I understand what Julia is doing with the examples below (this does not include Lily's code).  TechHydro does not have a struct docstring but TechThermal does.  It seems like we should be able to rely on Julia.

```
help?> TechHydro
search: TechHydro

  No documentation found.

  Summary
  ≡≡≡≡≡≡≡≡≡

  mutable struct TechHydro <: PowerSystems.TechnicalParams

  Fields
  ≡≡≡≡≡≡≡≡

  rating              :: Float64
  primemover          :: PowerSystems.PrimeMovers
  activepowerlimits   :: NamedTuple{(:min, :max),Tuple{Float64,Float64}}
  reactivepowerlimits :: Union{Nothing, NamedTuple{(:min, :max),Tuple{Float64,Float64}}}
  ramplimits          :: Union{Nothing, NamedTuple{(:up, :down),Tuple{Float64,Float64}}}
  timelimits          :: Union{Nothing, NamedTuple{(:up, :down),Tuple{Float64,Float64}}}
  internal            :: InfrastructureSystems.InfrastructureSystemsInternal

  Supertype Hierarchy
  ≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡

  TechHydro <: PowerSystems.TechnicalParams <: PowerSystemType <: InfrastructureSystems.InfrastructureSystemsType <: Any

help?> TechThermal
search: TechThermal

  Data Structure for the technical parameters of thermal generation technologies.
```

jd-lara: Well, we sometimes also need to add comments about where the data is coming from and the units of the values. I don't see the problem in duplicating the text, it seems that is the common practice in Julia. 

As for the difference between Hydro and Thermal I don't know why Julia shows different content for each.

PR #376, opened 2019-11-05
Change forecast labels to component accessor functions

daniel-thom:
There is one thing not done here:  When calling add_forecast there is no validation that the label translates to a function that is callable on the component.  I'll wait to do that until you guys take a first pass on the implementation.

Note that it would be trivial to change label from accessor function to fields.

Also note that I DID create a branch in PowerSystemsTestData in support of these changes but NOT the RTS_GMLC repo.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -2,37 +2,36 @@
 const SKIP_PM_VALIDATION = false
 
 """
-    System
+System
 
-    A power system defined by fields for basepower, components, and forecasts.
+A power system defined by fields for basepower, components, and forecasts.
 
-    # Constructor
-    ```julia
-    System(basepower)
-    System(components, basepower)
-    System(buses, generators, loads, branches, storage, basepower, services, annex; kwargs...)
-    System(buses, generators, loads, basepower; kwargs...)
-    System(file; kwargs...)
-    System(; buses, generators, loads, branches, storage, basepower, services, annex, kwargs...)
-    System(; kwargs...)
-    ```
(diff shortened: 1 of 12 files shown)


src/base.jl
```diff
@@ -650,7 +645,7 @@ Return a TimeSeries.TimeArray where the forecast data has been multiplied by the
 component field.
 """
 function get_forecast_values(component::Component, forecast::Forecast)
-    return IS.get_forecast_values(component, forecast)
+    return IS.get_forecast_values(PowerSystems, component, forecast)
```
jd-lara: What is the variable PowerSystems?

src/models/generation.jl
```diff
@@ -13,3 +13,6 @@ function IS.get_limits(valid_range::Union{NamedTuple{(:min,:max)}, NamedTuple{(:
     return (min = valid_range.min, max = valid_range.max, zero = 0.0)
 end
 
+get_rating(gen::Generator) = get_rating(get_tech(gen))
```
jd-lara: ***This comment is not about the PR***

A get rating for branches is missing here. @claytonpbarrows what should we do about MonitoredLines

src/base.jl
daniel-thom: In order to acquire the function represented by the label (which is a string) we have to call getfield on the module containing the function.  In this case the function resides in a different module than the get_forecast_values function, so we have to pass in the module name (PowerSystems).

PR #383, opened 2019-11-10
Jd/docstrings

jd-lara:


jd-lara:
@Lilyhanig and @daniel-thom there is still some kink with the docstrings. They look good in the REPL, but in the docs aren't printing quite right https://nrel.github.io/PowerSystems.jl/latest/api/PowerSystems/#PowerSystems.GenericBattery

PR #385, opened 2019-11-12
add supplemental constructor for StaticReserve

jd-lara:


src/models/supplemental_constructors.jl
```diff
@@ -55,3 +55,11 @@ function Bus(number, name, bustype::String, angle, voltage, voltagelimits, basev
     return Bus(number, name, get_enum_value(BusType, bustype), angle, voltage,
                voltagelimits, basevoltage, InfrastructureSystemsInternal())
 end
+
+"""Allows to contruct a reserve from and iterator """
```
daniel-thom:
```suggestion
"""Allows construction of a reserve from an iterator."""
```

src/models/supplemental_constructors.jl
```diff
@@ -55,3 +55,11 @@ function Bus(number, name, bustype::String, angle, voltage, voltagelimits, basev
...
+function StaticReserve(name, contributingdevices::IS.FlattenIteratorWrapper,
+                       timeframe, requirement, _forecasts, InfrastructureSystemsInternal)
+
```
daniel-thom: Why do you always make the first line of a function a blank line?  Is that in some style guide?

jd-lara: I just find it easier to read, no specific reason

daniel-thom: Ok, no problem.  Just no that you are the only person who does it, so we can always tell who wrote a function.  That isn't necessarily a good thing.  It's better if all code in a package looks the same.

PR #386, opened 2019-11-20
Jd/psi service implementation

### LINKED ISSUE #334: Inconsistent definition of reserves and parsing of CDM data.
When using the TableDataFormat to parse CDM data from the RTS GMLC 7 `ProportionalReserves` are parsed into the system. However, the definitions of these reserves are done with a string in the struct and there is no consistency with the ramp products already defined in the reserves structs. 

For instance, a `ProportionalReserve` is used in the case of a static reserve. 

This issue goes into the Forecasts and creates forecasts that can't be properly attached to a reserve product.

jd-lara:
There is one test failing when creating the `rts_system()`. I am opening the PR to ask for some pointers @daniel-thom. 

@claytonpbarrows the PR includes the discussion we had about 2 reserve data templates. Closes   #334


src/descriptors/power_system_structs.json
```diff
@@ -1497,10 +1497,12 @@
         "validation_action": "error"
       },
       {
-        "name": "_forecasts",
-        "null_value": "InfrastructureSystems.Forecasts()",
-        "data_type": "InfrastructureSystems.Forecasts",
-        "default": "InfrastructureSystems.Forecasts()"
```
daniel-thom: The test is failing because you removed forecasts storage container from StaticReserve.  The raw files for the RTS system have time series data for this type.  Was that intentional?

Also, we could detect this error and display a helpful message.  I'll do that separately.

src/descriptors/power_system_structs.json
```diff
@@ -1486,7 +1486,7 @@
         "name": "contributingdevices",
         "comment": "devices from which the product can be procured",
         "null_value": "[ThermalStandard(nothing)]",
-        "data_type": "Vector{Device}"
+        "data_type": "Vector{<:Device}"
```
daniel-thom: The serialization test fails because of this change. I'd like to discuss why you made it.

src/models/supplemental_constructors.jl
```diff
@@ -57,8 +57,8 @@ function Bus(number, name, bustype::String, angle, voltage, voltagelimits, basev
 end
 
 """Allows construction of a reserve from an iterator."""
-function StaticReserve(name, contributingdevices::IS.FlattenIteratorWrapper,
-                       timeframe, requirement, _forecasts, InfrastructureSystemsInternal)
-    return StaticReserve(name, collect(contributingdevices),
+function VariableReserve(name, contributingdevices::IS.FlattenIteratorWrapper{D},
```
daniel-thom: I don't think you need to parameterize this.  Even if you do, since you aren't using D anywhere, you can just write ```FlattenIteratorWrapper{Device}```

test/test_internal.jl
```diff
@@ -45,5 +45,5 @@ end
 
 @testset "Test internal values" begin
     sys_rts = create_rts_system()
-    @test validate_uuids(sys)
+    @test validate_uuids(sys_rts)
```
daniel-thom: This is a good example of why global variables are bad, even in tests.  test_busnumberchecks.jl defines ```sys``` as a global and so that is why this test was passing.

src/descriptors/power_system_structs.json
jd-lara: The change was intentional. The idea is to make the definitions consistent #334, what we had defined as "Static" is in reality a Dynamic reserve and proportional never made too much sense. So replaced Static with dynamic and created a new Static reserve that has no forecast. 

The display message isn't clear. It just throws a MethodError

PR #389, opened 2019-11-21
add ext to structs

jd-lara:
requires https://github.com/NREL/InfrastructureSystems.jl/pull/36 to pass tests


kdheepak:
I would prefer `ext` to not be a `Union{Nothing, Dict{String, Any}}` and instead be a `Dict{String, Any}`. If it is a `Union{Nothing, Dict{String, Any}}` every time a user uses `get_ext`, they have to check if it is nothing or not, and then check if the dictionary has the key they need. Leaving it as an empty dictionary probably what we want here.

daniel-thom:
> I would prefer `ext` to not be a `Union{Nothing, Dict{String, Any}}` and instead just be a `Dict{String, Any}`. If it is a `Union{Nothing, Dict{String, Any}}` every time a user uses `get_ext`, they have to check if it is not `nothing`, and then check if the dictionary has the key they need. Leaving it as an empty dictionary probably what we want here.

I like the semantics shown here where the constructor has `ext=nothing`.  But you do make a good point.  Two possible ways around that:

1. get_ext is defined at the highest possible level and implements the isnothing check.
2. We create a custom type which behaves exactly like a dict even if `nothing` was passed.

src/descriptors/power_system_structs.json
```diff
@@ -24,6 +24,12 @@
         "data_type": "InfrastructureSystems.Forecasts",
         "default": "InfrastructureSystems.Forecasts()"
       },
+      {
+        "name": "ext",
+        "data_type": "Union{Nothing, Dict{String, Any}}",
+        "null_value": "nothing",
+        "default": "nothing"
+      },
```
daniel-thom: This should be defined before _forecasts so that the constructor that uses default values will work correctly (not the kwargs constructor).  The user will never pass _forecasts but will pass ext, so ext should be first.  This is true for every type here with _forecasts.

jd-lara:
I added to this PR an example of how the dynamics will look like 
https://github.com/NREL/PowerSystems.jl/blob/jd/add_ext/src/models/generated/AVRFixed.jl

src/descriptors/LITS-full-structs.json
```diff
@@ -0,0 +1,1082 @@
+[
```
daniel-thom: Should this file be called LITS_structs.json?

src/descriptors/power_system_inputs.json
```diff
@@ -219,37 +219,37 @@
         {
             "unit": "MW",
             "name": "active_power",
-            "description": "Real power injection setpoint",
+            "description": "Real power StaticInjection setpoint",
```
daniel-thom: Looks like you have a lot of accidental name replacements for injection.  You should review all of them.

src/descriptors/power_system_structs.json
```diff
@@ -1459,16 +1585,23 @@
...
+        "default": "Dict{String, Any}()"
+      },
       {
         "name": "_forecasts",
         "null_value": "InfrastructureSystems.Forecasts()",
         "data_type": "InfrastructureSystems.Forecasts",
         "default": "InfrastructureSystems.Forecasts()"
       },
-      {
+            {
```
daniel-thom: nit: whitespace

src/descriptors/power_system_structs.json
```diff
@@ -1591,9 +1744,54 @@
...
+        "internal_default": "InfrastructureSystemsInternal()"
+      }
+    ],
+    "supertype": "AVR"
   }
+
+
+
+
+
```
daniel-thom: nit: whitespace

src/models/dynamic_inverter.jl
```diff
@@ -0,0 +1,84 @@
+abstract type InverterComponent <: DynamicComponent end
+
+mutable struct DynamicInverter{C<:Converter,
+                               O<:OuterControl,
+                               VC<:VSControl,
+                               DC<:DCSource,
+                               P<:FrequencyEstimator,
+                               F<:Filter} <: DynInjection
```
daniel-thom: Parameterizing on 6 different types seems like a lot of complexity.  Maybe it's required. It also makes me wonder if the problem can be broken down into smaller pieces.
If you plan to use this for multiple dispatch, trying to tell the difference between those function calls with your eyes is going to be difficult.

src/utils/power_flow/power_flow.jl
```diff
@@ -72,16 +72,16 @@ function _write_pf_sol!(sys::System, nl_result)
         if bus.bustype == PowerSystems.REF
             P_gen = result[2 * ix - 1]
             Q_gen = result[2 * ix]
-            injection_components = get_components(Generator, sys)
-            devices = [d for d in injection_components if d.bus == bus]
+            StaticInjection_components = get_components(Generator, sys)
```
daniel-thom: more accidental replacements

src/descriptors/LITS-full-structs.json
jd-lara: This file got removed

src/models/dynamic_inverter.jl
jd-lara: This is because there are too many internal components for an Inverter, a simple inverter already has to make 19 states. But there are too many interactions to separate further.

jd-lara:
@daniel-thom  all the comments got addressed in ed2a187

daniel-thom:
How about a test that serializes and deserializes a struct with data stored in ext?

jd-lara:
Pending

jd-lara:
> How about a test that serializes and deserializes a struct with data stored in ext?

added in cc2e247

PR #390, opened 2019-11-21
Style fixes

kdheepak:


src/base.jl
```diff
@@ -291,15 +339,16 @@ function IS.add_forecast!(
 
     if component isa LoadZones
         uuids = Set([IS.get_uuid(x) for x in get_buses(component)])
-        for component_ in (load for load in IS.get_components(ElectricLoad, data)
-                          if get_bus(load) |> IS.get_uuid in uuids)
+        for _component in (load for load in IS.get_components(ElectricLoad, data)
```
daniel-thom: What's wrong with component_?

src/base.jl
```diff
@@ -291,15 +339,16 @@ function IS.add_forecast!(
...
+                          if IS.get_uuid(get_bus(load)) in uuids)
             if forecast isa IS.DeterministicInternal
-                forecast_ = IS.DeterministicInternal(IS.get_label(forecast), ts_data)
+                _forecast = IS.DeterministicInternal(IS.get_label(forecast), ts_data)
             else
                 # TODO: others
                 error("forecast type is not supported yet: $(typeof(forecast))")
             end
-            IS.add_forecast!(data, component_, forecast, ts_data)
+            # FIXME: Should this be `_forecast` or `forecast`?
```
daniel-thom: Poorly written code that will do the same thing either way.  I'll fix this separately.

src/base.jl
```diff
@@ -640,11 +697,11 @@ end
...
-                      ::Type{T},
-                      component::Component,
-                      initial_time::Dates.DateTime,
-                      label::AbstractString,
-                      ) where T <: Forecast
+    ::Type{T},
+    component::Component,
+    initial_time::Dates.DateTime,
+    label::AbstractString,
+) where T <: IS.Forecast
```
daniel-thom: We made a choice to re-export Forecast as a PowerSystems type so that the user doesn't have to know this is an IS type.  I think I still prefer what we had.

src/base.jl
```diff
@@ -698,17 +755,16 @@ end
...
 component field.
 """
-function get_forecast_values(component::Component, forecast::Forecast)
+function get_forecast_values(component::Component, forecast::IS.Forecast)
     return IS.get_forecast_values(PowerSystems, component, forecast)
 end
 
 """
-    get_forecast_initial_times(sys::System)::Vector{Dates.DateTime}
+    get_forecast_initial_times(sys::System) -> Vector{Dates.DateTime}
```
daniel-thom: Is the `->` documented somewhere as a Julia rule?

src/base.jl
```diff
@@ -1043,6 +1093,7 @@ end
 
 function handle_component_addition!(sys::System, bus::Bus)
     number = get_number(bus)
+    # FIXME: call `check_component_addition` here instead of `@assert`
```
daniel-thom: That already happens.  If this trips, it is a programmer mistake.

src/base.jl
```diff
@@ -1075,33 +1126,35 @@ function check_bus(sys::System, bus::Bus, component::Component)
...
-    if !IS.compare_values(x.data, y.data)
-        @debug "SystemData values do not match"
-        match = false
-    end
+function IS.compare_values(x::System, y::System)
 
     if x.basepower != y.basepower
         @debug "basepower does not match" x.basepower y.basepower
-        match = false
+        return false
```
daniel-thom: I wrote the original logic so that you would get error prints for both basepower mismatches and data mismatches.  I still prefer that.  That pattern exists in several places in the code.  Do you disagree with that pattern?

I suggest not changing logic in a "style fixes" PR.

src/models/supplemental_constructors.jl
```diff
@@ -9,56 +9,127 @@ function ThreePartCost(variable_cost::T, args...) where {T <: VarCostArgs}
...
+        rate,
+        (min=-anglelimits, max=anglelimits),
+    )
 end
 
 """Allows construction with bus type specified as a string for legacy code."""
-function Bus(number, name, bustype::String, angle, voltage, voltagelimits, basevoltage)
-    return Bus(number, name, get_enum_value(BusType, bustype), angle, voltage,
-               voltagelimits, basevoltage, InfrastructureSystemsInternal())
+function Bus(
```
daniel-thom: I don't understand this change.  If it fits on one line, it's OK, right?

jd-lara:
I would like to add to this function naming, and this is what I am proposing for our packages, which is a small departure from the agreed JuMP style guide. We have 3 types of code users: 
    1. Main developers (our team)
    2. Modelers, people who would want to do some customization and need to implement methods for their custom models. 
   3. Analysts, people who just want to run something with the exported functions. 

In this regard I propose the following 

functions meant to users type 1 must be preceded by and _ and not be exported 
functions meant for users of type 2 won't be preceded by _ and not be exported
functions meant for users of type 3 won't be preceded by _ and exported

src/models/supplemental_constructors.jl
```diff
@@ -9,56 +9,127 @@ function ThreePartCost(variable_cost::T, args...) where {T <: VarCostArgs}
...
+    number,
+    name,
+    bustype::String,
+    angle,
+    voltage,
+    voltagelimits,
+    basevoltage,
+)
+    return Bus(
```
daniel-thom: Maybe we should discuss this one.  Some flexibility at the call site would be OK.  I think I prefer the way it was.

src/parsers/pm2ps_parser.jl
```diff
@@ -248,23 +254,7 @@ function make_renewable_fix(gen_name, d, bus)
...
-    #    rating=,
-    #    activepower=,
-    #    inputactivepowerlimits=,
-    #    outputactivepowerlimits=,
-    #    efficiency=,
-    #    reactivepower=,
-    #    reactivepowerlimits=,
-    #)
-    #return battery
+    error("Not implemented yet.")
```
daniel-thom: Someone is going to have to re-type most of the code that you removed.  Might as well leave it.

src/utils/IO/base_checks.jl
```diff
@@ -5,7 +5,7 @@ function orderedlimits(limits::Union{NamedTuple{(:min, :max), Tuple{Float64, Flo
         @info "'$limitsname' limits defined as nothing"
     else
         if limits.max < limits.min
-            throw(DataFormatError("$limitsname limits not in ascending order"))
+            throw(IS.DataFormatError("$limitsname limits not in ascending order"))
```
daniel-thom: We re-exported DataFormatError intentionally.  We should discuss.

daniel-thom:
> I would like to add to this function naming, and this is what I am proposing for our packages, which is a small departure from the agreed JuMP style guide. We have 3 types of code users:
> 1. Main developers (our team)
> 2. Modelers, people who would want to do some customization and need to implement methods for their custom models.
> 3. Analysts, people who just want to run something with the exported functions.
> 
> In this regard I propose the following
> 
> functions meant to users type 1 must be preceded by and _ and not be exported
> functions meant for users of type 2 won't be preceded by _ and not be exported
> functions meant for users of type 3 won't be preceded by _ and exported

I definitely see where you're coming from and you have a point.  But I also have opinions that functions with preceding underscores (1) don't read well and (2) belong only in a file where the only call site is in that file.  This is pretty common.

In PowerSystems at least, I don't think we expect even power users to re-implement our methods. They will be implementing methods for their extended types.

src/base.jl
kdheepak: I have a preference for using `_component`, with underscore as a prefix, because to me, having a `_` in `_local_variable` provides some context about the scope of the `_local_variable` variable.

I have been thinking about this a little more and my thoughts are as follows. I think there's at least three places one can use `_` as a prefix to a symbol

1) Variable names
2) Field names
3) Function names

And each of these has slightly different meanings to someone trying to understand what the code is doing.

___

1) Variable names

Take the following trivial function as an example.


```julia
"""
Calculate total of elements that are less than or equal to 5
"""
function calculate_total(elements::Vector{Int})
    ignore = Int[]

    for (i, _element) in enumerate(elements)
        if _element > 5
            push!(ignore, i)
        end
    end

    total = 0

    for (i, _element) in enumerate(elements)
        if !(i in ignore)
            total += _element
        end
    end

   return total
end
```

There's two blocks of code here, a) the block that creates the ignore list and b) the block that sums the code.

Using the underscore in the `_element` variable clarifies that changing the variable name `_element` in block a) should not the use of `_element` in block b). It also implies that `_element` is of local scope to the `for` loop, since `_element` is not defined outside the body of the `for` loop.

```
julia> function calculate_total(elements::Vector{Int})
           ignore = Int[]

           for (i, _element) in enumerate(elements)
               if _element > 5
                   push!(ignore, i)
               end
           end


           @show _element # <- add this line to throw an error.


           total = 0

           for (i, _element) in enumerate(elements)
               if !(i in ignore)
                   total += _element
               end
           end

          return total
       end
calculate_total (generic function with 1 method)

julia> calculate_total([1, 2, 3, 4, 5, 6, 7, 8])
ERROR: UndefVarError: _element not defined
Stacktrace:
 [1] macro expansion at ./show.jl:562 [inlined]
 [2] calculate_total(::Array{Int64,1}) at ./REPL[18]:11
 [3] top-level scope at REPL[19]:1

julia>
```

This is not to say that all `for` loop element variable names should be prefixed with an underscore since that is just visual noise and is unnecessary. In a larger function body, if the terminology for a `element` needs to be repeated and used in separate contexts, using `_element` helps to clarify context and scope and can improve readability.

2) Field names

In various Object Oriented Programming (OOP) based languages, there are ways within the language to enforce visibility of fields that belong to a class. Dynamic languages (such as Julia, Python) tend to not support restricting of direct access to some of an object's fields. In this case, the convention for using an underscore is indicate that the field in question has some implementation details of the struct and off limits to an user as an interface. However, using this convention does not mean that field names without an underscore are all part of the public interface i.e. one should refer to the documentation to find what is part of the interface and what are fields are guaranteed to be stable.

Let's say we are implementing a `Vector3D` struct in Julia.



As an example, take the following code.

````julia
"""
Define Vector3D struct

Public API:

    v = Vector3D(x, y, z)
    println(v.x)
    println(v.y)
    println(v.z)

    println(get_magnitude(v))
    println(get_polar_angle(v))
    println(get_azimuth_angle(v))
"""

export Vector3D

"""
    Vector3D(x::Int, y::Int, z::Int)

# Example

```julia
v = Vector3D(1, 1, 1)
```
"""
struct Vector3D
    x::Int
    y::Int
    z::Int
    r::Float64       # private variable
    polar::Float64   # private variable
    azimuth::Float64 # private variable
    _r_xy::Float64   # implementation details
    _r_yz::Float64   # implementation details
    _r_zx::Float64   # implementation details

    function Vector3D(x, y, z)
        r = √(x^2 + y^2 + z^2)
        return new(
            x,
            y,
            z,
            r,
            arctan(y/x),
            arccos(z/r),
            √(x^2 + y^2),
            √(y^2 + z^2),
            √(z^2 + x^2),
        )
    end
end
get_magnitude(v::Vector3D) = v.r
get_polar_angle(v::Vector3D) = v.polar
get_azimuth_angle(v::Vector3D) = v.azimuth
````

If `Vector3D` is an `export` in say `v2.1.0` and if you want to change the field names to `position_x`, `position_y` and `position_z`, then based on semantic versioning you should release a new package with version `v3.0.0`.

But based on semantic versioning, it is technically acceptable to make the following change in a patch release (i.e. `v2.1.0` -> `v2.1.1`) since your public API according to your documentation has not changed.

````julia
"""
Define Vector3D struct

Public API:

    v = Vector3D(x, y, z)
    println(v.x)
    println(v.y)
    println(v.z)

    println(get_magnitude(v))
    println(get_polar_angle(v))
    println(get_azimuth_angle(v))
"""

export Vector3D

"""
    Vector3D(x::Int, y::Int, z::Int)

# Example

```julia
v = Vector3D(1, 1, 1)
x = v.x
y = v.y
z = v.z
println(get_magnitude(v))
println(get_polar_angle(v))
println(get_azimuth_angle(v))
```
"""
struct Vector3D
    x::Int
    y::Int
    z::Int

    Vector3D(x, y, z) = new(x, y, z)
end
get_magnitude(v::Vector3D) = √(v.x^2 + v.y^2 + v.z^2)
get_polar_angle(v::Vector3D) = arctan(v.y/v.x)
get_azimuth_angle(v::Vector3D) = arccos(v.z/get_magnitude(v))
````

A user using `Vector3D` in their application / package / library should not be surprised to find that their previous code that used `v._r_x` or even `v.a_angle` broke with a new patch update, since based on the documentation the public interface defines only `x`, `y` and `z` field attributes as the API, and the remaining fields can only be accessed using the functional getter interface.

3) Function names

Finally, function names with an underscore prefix are those seem to hint that there is some abstraction that reads cleaner as a separate function but is used just a handful of times, or is a hacky solution to a problem that you want to solve temporarily in this manner but will get rid off in the future without changing anything else in the package. These underscore prefixed functions are usually local to a file in which the function is defined.

Consider the following package:

```julia
# src/Package.jl

export report_metrics
export Type1
export Type2
export Type3

include("utils.jl")
include("type1_element.jl")
include("type2_element.jl")
include("type3_element.jl")
```

With the following `src/utils.jl` file:

```julia
# src/utils.jl
abstract type Element end

function _get_int64_np_array(e::Element)
   # convert Julia type into corresponding NumPy array for metrics calculation
   ...
   return array
end

function _py_calculate_metrics(e::Element)
    pymetrics = pyimport("pymetrics")
    return pymetrics.calculate_metrics(_get_int64_np_array(e))
end

function calculate_metrics(e::Element)
    ...
    # FIXME: rewrite pymetrics in Julia for better performance
    m = _hack_metrics(e)
    return m
end
```

And the following files:

```julia
# src/type1_element.jl

struct Type1 <: Element end

function report_metrics(t1::Type1)
    ...
    m = calculate_metrics(t1)
    ...
end
```

```julia
# src/type2_element.jl

struct Type2 <: Element end

function report_metrics(t2::Type2)
    ...
    m = calculate_metrics(t2)
    ...
end
```

```julia
# src/type3_element.jl

struct Type3 <: Element end

function report_metrics(t3::Type3)
    ...
    m = calculate_metrics(t3)
    ...
end
```

- `report_metrics` is clearly part of the public API since it is exported.
- `calculate_metrics` is not part of the public API and should not be used. If a user decides to use it, they may be subjecting themselves to breaking code in the future.
- `_py_calculate_metrics` is a hacky solution to get something working temporarily
- `_get_int64_np_array` is a implementation detail required in the package that is not related to what the package does, but exists just to get the `pymetrics` package to work.

Even though `calculate_metrics` is not an `export` and may never be an `export`, it (in my opinion) should not have an underscore prefix since API is only defined by what is or is not exported / is or is not documented. Underscores as prefix to function names are relegated to implementation details that are local to an aspect of the package that is not related to the "domain" or the "application" or some hacky solution that will be removed in the future.

___

That was a brain dump of an afternoon's thought. I'm happy to discuss this further if people have more thoughts or comments regarding this.

src/base.jl
kdheepak: Screen Shot 2019-11-21 at 4 47 59 PM

kdheepak: I added the use of DocStringExtensions in https://github.com/NREL/PowerSystems.jl/pull/390/commits/a58e6eca0d257a8a83ca01ef4bf106e9d39ae400 and https://github.com/NREL/PowerSystems.jl/pull/390/commits/2d9d2e023264c5ad02c5efec7e64b4e08b0a5842. See above screenshot for snippet of code in the file versus what shows up in the REPL when requesting for documentation. The same docstring that shows up in the REPL will show up in the documentation. 

We minimize the code we have to repeat in the function signature and documentation, however we lose some control on how the docstring is presented.

src/models/supplemental_constructors.jl
kdheepak: Resolved in https://github.com/NREL/PowerSystems.jl/pull/390/commits/5f9a17d80dda4ee219c3cb29a7c2a5fa845444f6.

src/base.jl
kdheepak: Resolved in https://github.com/NREL/PowerSystems.jl/pull/390/commits/03635a151b5c9a9580a476a0cc169029c8379ba2.

src/base.jl
kdheepak: Will be resolved in a separate PR.

src/models/supplemental_constructors.jl
kdheepak: Flexibility would be nice, but this above change prioritizes consistency with the rest of the file.

src/parsers/pm2ps_parser.jl
kdheepak: > Don't use block comments for Julia code. Prefer using the # prefix. If you are commenting code, consider deleting it instead.

We had something like this in our style guide. This isn't a significant amount of code, we can rewrite it without mistakes and without much effort if needed in the future.

jd-lara:
@kdheepak you should rebase this branch on top of https://github.com/NREL/PowerSystems.jl/pull/389 to account for the new content.

jd-lara:
@kdheepak seems that recent changes created a couple of conflicts.

src/base.jl
kdheepak: Resolved in https://github.com/NREL/PowerSystems.jl/pull/390/commits/59e59130edc3bdb963631ba666690f44811ba0b1.

src/base.jl
kdheepak: Resolved in https://github.com/NREL/PowerSystems.jl/pull/390/commits/aa6ad0775b1ef4dff3e2cf013bccad77e9be45a6.

src/utils/IO/base_checks.jl
kdheepak: Resolved in https://github.com/NREL/PowerSystems.jl/pull/390/commits/9294f19f8db4db08e2ffd1dc99395402513bbe78.

jd-lara:
@kdheepak and @daniel-thom I am merging this to make sure all the tests of PowerSimulations pass.

PR #394, opened 2019-11-22
Jd/lits structs

jd-lara:
@rodrigomha, this PR moves the LITS structs to PowerSystems. DynLines is not here since it has the same template as a regular line.


src/descriptors/power_system_structs.json
```diff
@@ -1771,13 +1771,51 @@
...
+        "comment": "power system internal reference, do not modify",
+        "data_type": "InfrastructureSystemsInternal",
+        "internal_default": "InfrastructureSystemsInternal()"
+      }
+    ],
+    "supertype": "AVR"
+  },
+  {
+    "struct_name": "AVRSimple",
+    "docstring": "Parameters of a simple proportional AVR in the derivative of EMF\ni.e. an integrator controller on EMF",
```
daniel-thom: I don't think we want to include formatting characters like \n in these strings.  A higher level could do that automatically.  But, I'm open.

jd-lara: I was getting issues with the JSON format. What do you suggest?

daniel-thom: Don't include \n in the string.  If we want line wrapping in the generated docstrings, we can do later in generate_structs.jl (later).

jd-lara: let's solve this later.

PR #396, opened 2019-11-24
Extended hr cols

claytonpbarrows:
- creates a superset of heat rate traunches to support parsing up to 13 point heat rate curves


claytonpbarrows:
@daniel-thom this seems like something we could do programmatically, but this solution was simple enough... let me know what you think.

PR #398, opened 2019-11-26
Parameterize Reserve structs with direction

daniel-thom:


src/base.jl
```diff
@@ -452,6 +452,105 @@ function get_components(
     return IS.get_components(T, sys.data)
 end
 
+# The following functions are reimplemented because Reserve subtypes are parameterized and
+# so we cannot auto-discover the types.
+# TODO: this needs performance improvement to reduce the number of extra arrays that get
+# created.
```
daniel-thom: @jd-lara @kdheepak I'd like you guys to review this and let me know if you have a better idea on how this can be done.

PR #402, opened 2019-12-05
skipping load creation for buses with zero load

claytonpbarrows:


src/parsers/power_system_table_data.jl
```diff
@@ -534,18 +534,19 @@ function load_csv_parser!(sys::System, data::PowerSystemTableData)
...
-            available = true,
-            bus = ps_bus,
-            model = ConstantPower::LoadModel,
-            activepower = active_power,
-            reactivepower = reactive_power,
-            maxactivepower = max_active_power,
-            maxreactivepower = max_reactive_power,
-        )
-        add_component!(sys, load)
+        if (max_active_power != 0.0) | (max_reactive_power != 0.0)
```
daniel-thom:
```suggestion
        if (max_active_power != 0.0) || (max_reactive_power != 0.0)
```

daniel-thom: Still correct here but you should really use a logical `or` instead of a bitwise `or`.

PR #405, opened 2020-01-07
Updating HydroDispatch

sourabhdalvi:
- Deleting HydroStorage
- Adding `storagecapacity` & `initial_storage` fields  to HydroDispatch struct
- Adding  `storage::Union{DataFrames.DataFrame,Nothing}` field to PowerSystemTableData to read in `storage.csv` for Hydro storage related fields

For the test to pass it require the sd/hydro_update branch on PowerSystemsTestData repo to be merged into master


src/parsers/pm2ps_parser.jl
```diff
@@ -211,6 +211,8 @@ function make_hydro_gen(gen_name, d, bus)
         reactivepower = d["qg"],
         tech = tech,
         op_cost = curtailcost,
+        storagecapacity = 0.0, #TODO: Implement better Solution for this
```
jd-lara: storage_capacity

src/parsers/power_system_table_data.jl
```diff
@@ -13,6 +13,7 @@ struct PowerSystemTableData
     gen::Union{DataFrames.DataFrame,Nothing}
     load::Union{DataFrames.DataFrame,Nothing}
     services::Union{DataFrames.DataFrame,Nothing}
+    storage::Union{DataFrames.DataFrame,Nothing}
```
jd-lara: Is this inflow?

src/models/generated/HydroDispatch.jl
```diff
@@ -10,6 +10,8 @@ This file is auto-generated. Do not edit.
         reactivepower::Float64
         tech::TechHydro
         op_cost::TwoPartCost
+        storagecapacity::Float64
```
jd-lara: we might need to add an inflow field as well here.

src/parsers/power_system_table_data.jl
sourabhdalvi: Thats for the `storage.csv` file in data/RTS_GMLC which currently holds :
- Max Volume GWh,
- Initial Volume GWh,
- Start Energy  GWh,
I think Inflow would be in  the data/forecasts

src/models/generated/HydroDispatch.jl
sourabhdalvi: ok, I can do that. Should `storage_capacity` and `inflow`fields be in the TechHydro ?

sourabhdalvi:
> We need to add the inflow field and also include in tech_hydro a conversion function from flow to power. We need to make sure the units are all consistent here. For instance, flow in m^3/s and storage en m^3.

I and @claytonpbarrows  talked about this yesterday and it seems like most of the inflow information that is used/available at NREL is in terms of energy. So Clayton suggested we keep everything in energy units.  So `storage_capacity` and `inflow` would both in terms of energy and that way its easier to update the parameter for inflow in PSI to keep track of stored energy over the length of the simulation.

claytonpbarrows:
> > We need to add the inflow field and also include in tech_hydro a conversion function from flow to power. We need to make sure the units are all consistent here. For instance, flow in m^3/s and storage en m^3.
> 
> I and @claytonpbarrows talked about this yesterday and it seems like most of the inflow information that is used/available at NREL is in terms of energy. So Clayton suggested we keep everything in energy units. So `storage_capacity` and `inflow` would both in terms of energy and that way its easier to update the parameter for inflow in PSI to keep track of stored energy over the length of the simulation.

Yes. I also agree that for some formulations/implementations it would be useful to keep track of the water itself (inflow, level, etc.). But, I think we can add that later.

Issue #407, opened 2020-01-11
Improve Hydro parsing from tabular data


claytonpbarrows:
- handle instance where storage.csv isn't present
 - enable assignment of Hydro device type from generator_mapping.yaml

jd-lara:
@claytonpbarrows can we close this now that #408 is closed?

PR #408, opened 2020-01-13
supporting HydroFix parsing in table data

claytonpbarrows:
- addresses #407

### DIFF
--- a/src/parsers/power_system_table_data.jl
+++ b/src/parsers/power_system_table_data.jl
@@ -712,8 +712,8 @@ function make_generator(data::PowerSystemTableData, gen, cost_colnames, bus)
 
     if gen_type == ThermalStandard
         generator = make_thermal_generator(data, gen, cost_colnames, bus)
-    elseif gen_type == HydroDispatch
-        generator = make_hydro_generator(data, gen, bus)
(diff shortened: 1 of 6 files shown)


src/parsers/power_system_table_data.jl
```diff
@@ -827,21 +827,22 @@ function make_hydro_generator(data::PowerSystemTableData, gen, bus)
...
+        initial_storage = nothing
+        inflow = nothing
+        for row in eachrow(storage_df)
+            row[Symbol("GEN UID")], gen.name
+            if row[Symbol("GEN UID")] == gen.name
+                storage_capacity = row[Symbol("Max Volume GWh")]
+                initial_storage = row[Symbol("Initial Volume GWh")]
+                inflow = row[Symbol("Inflow Limit GWh")]
+                break
+            end
```
daniel-thom: After the for loop you should throw an error if isnothing(storage_capacity).

daniel-thom: Is it possible to use our parsing scheme that allows custom table column names?

sourabhdalvi: I do agree with that

PR #409, opened 2020-01-13
Convert to standard parsing scheme

daniel-thom:
This addresses my comments in https://github.com/NREL/PowerSystems.jl/pull/408.

### DIFF
--- a/src/descriptors/power_system_inputs.json
+++ b/src/descriptors/power_system_inputs.json
@@ -510,11 +510,11 @@
     "storage": [
         {
             "name": "name",
-            "description": "Gen ID associated with storage"
+            "description": "Storage object name"
         },
         {
-            "name": "name",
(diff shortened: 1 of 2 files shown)


src/descriptors/power_system_inputs.json
```diff
@@ -510,11 +510,11 @@
     "storage": [
         {
             "name": "name",
-            "description": "Gen ID associated with storage"
+            "description": "Storage object name"
         },
         {
-            "name": "name",
-            "description": "Storage object name"
+            "name": "generator_name",
```
claytonpbarrows: looks like this will require a change in Tests

daniel-thom: It requires the companion change in PowerSystemsTestData.  We'll need to merge them to master at the same time.

PR #419, opened 2020-01-30
Jd/formatter code

jd-lara:


.github/workflows/formatter_code.jl
```diff
@@ -0,0 +1,15 @@
+main_paths = ["./src", "./test"]
+for main_path in main_paths
+    for (root, dir, files) in walkdir(main_path)
+        for f in files
+        @show file_path = abspath(root, f)
+        !occursin(".jl", f) && continue
```
daniel-thom: Do you want to exclude files that we copied from PowerModels?  This will cause merge conflicts when you pull their code in the future.

jd-lara: I think we have made enough changes to make that code suite our needs that there will be conflicts anyway.

daniel-thom: Objections worked out in offline conversation.

PR #420, opened 2020-02-01
Jd/formatter fix

jd-lara:
Fixes a wrong option in the formatter code


jd-lara:
> I suggest also setting whitespace_typedefs = true. Lines like this are becoming unreadable (for me).
> 
> `where {T<:Union{JuMP.VariableRef,PJ.ParameterRef},U<:PSY.Component}`

ok, I will update this

PR #421, opened 2020-02-03
Hydro Struct Renaming

sourabhdalvi:
Changing Struct name 
- `HydroFix` -> `HydroDispatch`
- `HydroDispatch` -> `HydroEnergyReservoir`


jd-lara:
Consider that this PR will introduce a breaking change for PSI.

sourabhdalvi:
> Consider that this PR will introduce a breaking change for PSI.

I only did what I was told and I do have an open PR on PSI to update the naming.

PR #427, opened 2020-02-07
Simplify access to enum values

### LINKED ISSUE #424: Export load type enum
When defining a load it is required to define its type. Currently it is required to specify its type as PowerSystems.ConstantPower for example, instead of ConstantPower.

daniel-thom:
This a proposal to fix #424.  I think this issue should be addressed in Julia itself.  There is some discussion of it online.  Other languages make this easier, IMHO.

Here are the ways to fix/mitigate this in order of simplicity.
1. Export all enum types.  Functions could accept parameters of type `LoadModel` (not `PowerSystems.LoadModel`).  However, you would have to type `PowerSystems.ConstantPower::LoadModel` in a notebook or REPL.

2. Export all enum types and values (manually or with some macro).  You could just type `ConstantPower`.  This would give Jose and Rodrigo what they want.  However, it would pollute the namespace.  All enum values would essentially be global variables and you could have collisions.  I think it is better to always to type the enum type with the value (ConstantPower::LoadModel).

3. Implement the workaround proposed here:  https://discourse.julialang.org/t/encapsulating-enum-access-via-dot-syntax/11785/2.  I think that this is the best solution.  In this PR I implemented it for one enum type.

Let me know what you guys think.


kdheepak:
I like option 3. We can also decrease developer effort and avoid visual noise in code by using a macro to imitate default syntax.

```julia
"""
Macro to wrap Enum in a baremodule for keeping top level scope clean
"""
macro scoped_enum(T, args...)
    blk = esc(:(
            baremodule $(Symbol("$(T)s"))
               using Base: @enum 
               @enum $T $(args...)
            end
        ))
    top = Expr(:toplevel, blk)
    return top
end
```

Usage:

```julia
julia> @scoped_enum Fruit Apple=1 Banana=2
Main.Fruits

julia> Fruits.Apple::Fruits.Fruit
Apple::Fruit = 1
```

daniel-thom:
I like the enum Dheepak proposed. We should put it in InfrastructureSystems.

daniel-thom:
Implemented the third option with support in IS.

PR #431, opened 2020-02-13
Linetype conversion

claytonpbarrows:
Adds a simple conversion between Line and MonitoredLine. I'm interested in thoughts on whether or not we should enable this.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -1154,3 +1154,69 @@ function _create_system_data_from_kwargs(; kwargs...)
         time_series_in_memory = time_series_in_memory,
     )
 end
+
+"""
+Converts a Line component to a MonitoredLine component and replaces the original in the
+system
+"""
+function convert_component!(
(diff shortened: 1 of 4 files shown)


src/base.jl
```diff
@@ -1154,3 +1154,69 @@ function _create_system_data_from_kwargs(; kwargs...)
...
+        line.arc,
+        line.r,
+        line.x,
+        line.b,
+        (from_to = line.rate, to_from = line.rate),
+        line.rate,
+        line.anglelimits,
+        line.services,
+        line.ext,
+        line._forecasts,
```
daniel-thom: Time series data is stored with the component UUID, so this won't work unless you use the same UUID for the new line.  You I don't think it would be bad to do that.  Just pass line.internal to the constructor.  Otherwise, you will need to iterate over the line's forecasts, removing and adding each.

PR #434, opened 2020-02-21
Add Area and LoadZone to Bus

daniel-thom:
This has dependencies on changes to InfrastructureSystems and PowerSystemsTestData.


src/parsers/pm2ps_parser.jl
```diff
@@ -87,27 +91,49 @@ function Base.convert(::Type{BusTypes.BusType}, x::MatpowerBusTypes.MatpowerBusT
...
+#        @info "There are no Areas in this file"
+#        return
+#    end
+#
+#    for (key, val) in data["areas"]
+#        area = Area(string(val["col_1"]))
+#        add_component!(sys, area; skip_validation = SKIP_PM_VALIDATION)
+#    end
+#end
+
```
daniel-thom: I want to call out that in the old code, if there were no areas defined in the file, we did not create any LoadZones.  Some matpower files don't define areas even when there are areas (and zones).  We will new create LoadZones based on whether a bus has reference to one, so should now have LoadZone objects that did not previously get created.

src/descriptors/power_system_structs.json
```diff
@@ -264,6 +264,67 @@
         ],
         "supertype": "DeviceParameter"
     },
+    {
+        "struct_name": "Area",
+        "docstring": "A geographic area.",
```
jd-lara: Not a geographic area necessarily. It could also be a control area.

src/parsers/pm2ps_parser.jl
jd-lara: This behavior is ok by me.

src/models/generated/Area.jl
```diff
@@ -0,0 +1,41 @@
+#=
+This file is auto-generated. Do not edit.
+=#
+"""
+    mutable struct Area <: AggregationTopology
+        name::String
+        internal::InfrastructureSystemsInternal
+    end
+
+A geographic area.
```
claytonpbarrows: collection of busses?

src/base.jl
```diff
@@ -561,6 +564,28 @@ function get_contributing_device_mapping(sys::System)
     return services
 end
 
+"""
+    get_buses(sys::System, aggregator::AggregationTopology)
```
claytonpbarrows: This is nice. The next step for users will be to collect the other components in an AggregationTopology. That doesn't seem straightforward to me. Something like:
```julia
[c for c in get_components(Generator, sys) if get_bus(c) in get_buses(sys, area)]
```
It gets more complicated for Branches. 
I'm wondering if we should add a method for this, or if you think it's reasonable for users to figure it out?

jd-lara: We need to implement this only for injection devices at minimum

daniel-thom: Added for StaticInjection devices.  If we need to do something similar for DynamicInjection devices then we should add an issue to track it.

src/base.jl
```diff
@@ -561,6 +564,81 @@ function get_contributing_device_mapping(sys::System)
...
+
+    return components
+end
+
+"""
+    get_aggregation_topology_mapping( ::Type{T}, sys::System) where {T<:AggregationTopology}
+
+Return a mapping of AggregationTopology name to vector of buses within it.
+"""
+function get_aggregation_topology_mapping(
```
daniel-thom: Note that I added this. This is more efficient than a user looping on top of get_buses per Area because this makes one pass over buses.

src/base.jl
```diff
@@ -561,6 +564,81 @@ function get_contributing_device_mapping(sys::System)
...
+
+Return a vector of components with buses in the AggregationTopology.
+"""
+function get_components_in_aggregation_topology(
+    ::Type{T},
+    sys::System,
+    aggregator::U,
+) where {T <: StaticInjection, U <: AggregationTopology}
+    buses = Set{String}((get_name(x) for x in get_buses(sys, aggregator)))
+    accessor_func = get_aggregation_topology_accessor(U)
```
daniel-thom: This isn't needed.  I'll fix.

Issue #433, opened 2020-02-19
rename pm2ps_dict() to System()


claytonpbarrows:
[This function name](https://github.com/NREL/PowerSystems.jl/blob/master/src/parsers/pm2ps_parser.jl#L15) is misleading, it should probably be `System()` or `pm2system()` or something like that...

daniel-thom:
Just `System` with the current function parameters is very generic (`Dict{String, Any)}`.  It could collide with a future function that also takes a dict.  We could create a type for the input data and pass that.

Example:
```
abstract type PowerSystemsInputDataSource end
struct PowerModelsData <: PowerSystemsInputDataSource end

function System(::PowerModelsData, data::Dict{String, Any}; kwargs...)
```

I'm also OK with `pm2System`.

jd-lara:
closed via #435

PR #435, opened 2020-02-22
Change pm2ps_parser to power_models_data

daniel-thom:
This will fix issue #433.  It changes PowerModels parsing to follow the model of PowerSystemTableData.

This model has one limitation:  it requires two steps from the user.
```
pm_data = PowerModelsData(file)
sys = System(pm_data)
```

There is alternate solution that we could consider for both PowerModelsData and PowerSystemTableData. What do you think of this?

```
function System(::Type{PowerModelsData}, file::Union{String, IO}; kwargs...)
function System(
    ::Type{PowerSystemTableData},
    directory::AbstractString,
    base_power::Float64,
    user_descriptor_file::AbstractString;
    kwargs...
)
```


src/parsers/standardfiles_parser.jl
```diff
@@ -14,18 +14,6 @@ sys = parse_standard_files(
 
 """
 function parse_standard_files(file::String; kwargs...)
```
daniel-thom: I'm not sure that this function is helpful.  It provides a second way of doing the same thing but doesn't add value (other than being a single step).

jd-lara: We can remove this function and leave Systems as the only entry point for Matpower data.

jd-lara:
> This will fix issue #433. It changes PowerModels parsing to follow the model of PowerSystemTableData.
> 
> This model has one limitation: it requires two steps from the user.
> 
> ```
> pm_data = PowerModelsData(file)
> sys = System(pm_data)
> ```
> 
> There is alternate solution that we could consider for both PowerModelsData and PowerSystemTableData. What do you think of this?
> 
> ```
> function System(::Type{PowerModelsData}, file::Union{String, IO}; kwargs...)
> function System(
>     ::Type{PowerSystemTableData},
>     directory::AbstractString,
>     base_power::Float64,
>     user_descriptor_file::AbstractString;
>     kwargs...
> )
> ```

I think that the current design is ok. I don't see too much value on the dispatch on a type. If you want to add it is ok but it seems that calling directly the file and detecting the extension is more straightforward.

claytonpbarrows:
I like the single line `System` functionality. I'd be fine with the suggestion from @jd-lara to detect file extension to enable this. I'd also be fine with requiring users pass in the type. However, if we go with the latter, I'd suggest adding the following to help avoid confusion about what is PowerModelsData:
```julia
abstract type MATPOWER <: PowerModelsData
abstract type RAW <: PowerModelsData
```

daniel-thom:
I changed my mind on my proposal above.  The PowerSystemTableData and System constructors are two complicated to combine. I think it's best to leave them independent.

PR #436, opened 2020-02-24
Jd/fix pf return

### LINKED ISSUE #359: solve power flow should return a status
currently `solve_power_flow!()` just updates the setpoints in `System` and is quiet about the execution status of the power flow. What happens when the power flow doesn't converge? I suggest that the method returns the convergence status to indicate success. This is similar to the `run_pf` routine in MATPOWER.

jd-lara:
closes #359. The test fails because of the updates in TestData.


src/utils/power_flow/power_flow.jl
```diff
@@ -142,6 +142,11 @@ solve_powerflow!(sys, nlsolve, method = :Newton)
 function solve_powerflow!(sys, nlsolve; args...)
     pf!, x0 = PowerSystems.make_pf(sys)
     res = nlsolve(pf!, x0; args...)
-    PowerSystems._write_pf_sol!(sys, res)
-    return
+    if res.f_converged
+        PowerSystems._write_pf_sol!(sys, res)
+        println("PowerFlow solve converged, the results have been stored in the system")
```
daniel-thom: Should probably use logging macros instead of println.

jd-lara: fixed via 120aaca

PR #448, opened 2020-04-01
Cb/tabledata improvements

claytonpbarrows:
- minor table data parser adjustments
- adds a system check for sum(generation) > sum(load)

### DIFF
--- a/docs/src/man/parsing.md
+++ b/docs/src/man/parsing.md
@@ -45,7 +45,7 @@ column names.
 For example, when parsing raw data for a generator the code expects a column
 called `name`. If the raw data instead defines that column as `GEN UID` then
 you can change the `custom_name` field under the `generator` category to
-`GEN UID` in your YAML file.
+`GEN UID` in your YAML file. 
 
 #### Per-unit conversion
 PowerSystems defines whether it expects a column value to be per-unit in
@@ -70,16 +70,16 @@ PowerSystems requires a metadata file that maps components to their time series
 data in order to be able to automatically construct forecasts from raw data
 files. The following fields are required for each time array:
 
-* Simulation:  User description of simulation
-* Category:  Type of component. Must map to PowerSystems abstract types (Bus,
+* simulation:  User description of simulation
+* category:  Type of component. Must map to PowerSystems abstract types (Bus,
   ElectricLoad, Generator, LoadZone, Reserve)
-* Object:  Name of component
-* Parameter:  Name of accessor function that can be called on the component to
+* component_name:  Name of component
+* label:  Name of accessor function that can be called on the component to
   retrieve the forecasted value.
-* Scaling Factor:  Controls normalization of the data. Use 1.0 for
+* scaling_factor:  Controls normalization of the data. Use 1.0 for
   pre-normalized data. Use 'Max' to divide the timeseries by the max value in the
   column. Use any float for a custom scaling factor.
-* Data File:  Path to the time series data file
+* data_file:  Path to the time series data file
 
 PowerSystems supports this metadata in either CSV or JSON formats. Refer to
 [RTS_GMLC](https://github.com/GridMod/RTS-GMLC/blob/master/RTS_Data/FormattedData/SIIP/timeseries_pointers.json)
--- a/docs/src/man/parsing.md
+++ b/docs/src/man/parsing.md
@@ -45,7 +45,7 @@ column names.
 For example, when parsing raw data for a generator the code expects a column
 called `name`. If the raw data instead defines that column as `GEN UID` then
 you can change the `custom_name` field under the `generator` category to
-`GEN UID` in your YAML file. 
+`GEN UID` in your YAML file.
 
 #### Per-unit conversion
 PowerSystems defines whether it expects a column value to be per-unit in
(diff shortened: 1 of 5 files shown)


src/utils/IO/system_checks.jl
```diff
@@ -61,3 +61,19 @@ function critical_components_check(sys::System)
...
+- `sys::System`: system
+"""
+function adequacy_check(sys::System)
+    total_gen = sum(get_rating.(get_components(Generator, sys))) * get_basepower(sys)
+    @debug "Total System Generation: $total_gen"
+    total_load =
+        sum(get_maxactivepower.(get_components(ElectricLoad, sys))) * get_basepower(sys)
+    @debug "Total System Load: $total_load"
+    total_load > total_gen &&
+    @error "System peak load ($total_load) exceeds total generation capability ($total_gen)."
```
jd-lara, line 78: This should be a warning if someone intends to parse a system and then add components to it.

src/parsers/power_system_table_data.jl
```diff
@@ -216,8 +216,12 @@ end
 
 """Return the dataframe for the category."""
 function get_dataframe(data::PowerSystemTableData, category::InputCategory)
-    @assert haskey(data.category_to_df, category)
-    return data.category_to_df[category]
+    if !haskey(data.category_to_df, category)
```
jd-lara, line 219: This function is type unstable.

jd-lara:
Also, tests are failing

src/utils/IO/system_checks.jl
daniel-thom, line 78: Should this throw an exception instead?

src/parsers/power_system_table_data.jl
```diff
@@ -216,8 +216,12 @@ end
...
+        @warn("Missing $category data.")
+        return nothing
```
daniel-thom, line 221: This implies that every call site needs to check whether the returned value is `nothing`.  In what scenario will there be a missing category?

src/utils/IO/system_checks.jl
claytonpbarrows: @daniel-thom and @jd-lara, I wasn't sure what to do here. For the example that @jd-lara mentioned, I didn't throw an exception. However, I think the severity is such that a warning is maybe insufficient...

jd-lara: I think that there are too many cases where it does make sense to have more load than generators. I strongly suggest we keep it a warning since there aren't strong warnings.

docs/src/man/parsing.md
```diff
@@ -70,16 +70,16 @@ PowerSystems requires a metadata file that maps components to their time series
...
-* label:  Name of accessor function that can be called on the component to
+* Object:  Name of component
+* Parameter:  Name of accessor function that can be called on the component to
   retrieve the forecasted value.
-* scaling_factor:  Controls normalization of the data. Use 1.0 for
+* Scaling Factor:  Controls normalization of the data. Use 1.0 for
   pre-normalized data. Use 'Max' to divide the timeseries by the max value in the
   column. Use any float for a custom scaling factor.
-* data_file:  Path to the time series data file
+* Data File:  Path to the time series data file
```
daniel-thom, line 82: The old values are the actual field names that must exist in the metadata file (ex:  data/RTS_GMLC/timeseries_pointers.json).  Are you sure that you want to change this?

src/parsers/power_system_table_data.jl
claytonpbarrows: I have an example where the tabular data file doesn't have a reserves.csv file. It's perfectly valid to not have reserves, and therefore not create `service` components. It wasn't an obvious problem to debug when I encountered it... in fact, it wasn't really a problem. But I do want to be notified about not having information that might be expected.

docs/src/man/parsing.md
claytonpbarrows: ... Ahh... the problem is that the timeseries_pointers.json field values and the timeseries_pointers.csv required column names differ.  We should either:
1. include both options in these instructions
2. allow for column name mapping in timeseries_pointers.csv similar to the other .csv column mapping.

daniel-thom: One set of names would be best.  Would we cause pain for anyone if we changed one set to match the other?

src/parsers/power_system_table_data.jl
daniel-thom: Could we create an empty dataframe instead of nothing?

docs/src/man/parsing.md
claytonpbarrows: I agree that one set of names would be best. I think the only person that this will cause pain is me to update the RTS-GMLC repo (and maybe some of the other RTS-GMLC users).

src/utils/IO/system_checks.jl
daniel-thom: That *should not* apply here.  `check!` should only be called on the system once it has been built, such as after parsing or deserialization from JSON.

daniel-thom: Regardless of warning vs error, we will just confuse people with messages if their data is actually correct.  Is it possible (and reasonably feasible) to determine when this acceptable?

docs/src/man/parsing.md
claytonpbarrows: resolved by. https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/70

src/utils/IO/system_checks.jl
```diff
@@ -61,3 +61,53 @@ function critical_components_check(sys::System)
     missing_critical_components &&
     throw(IS.InvalidValue("Critical Componeents are not present."))
 end
+
+"""
```
daniel-thom, line 65: If you want these docstrings to be proper you have to add the function signature to the top.

```
    adequacy_check(sys::System)

Checks the system for sum(generator ratings) >= sum(load ratings).

...
```

PR #454, opened 2020-04-15
tamu parser

claytonpbarrows:
- adds a parser for tamu cases
 - catches issues with branch endpoints at different voltage levels to create transformers

requires data from: https://github.com/NREL-SIIP/PowerSystemsTestData/pull/25

### DIFF
--- a/src/parsers/TAMU_data.jl
+++ b/src/parsers/TAMU_data.jl
@@ -1,0 +1,102 @@
+
+"""
+    TAMU_System(TAMU_FOLDER::AbstractString;
+                     kwargs...)
+
+Creates a system from a PSS/e .RAW (v33) load flow case, and an associated .csv with MW load
+time series data. The format is established by the
+[Texas A&M University Test Case Archive](https://electricgrids.engr.tamu.edu/electric-grid-test-cases/)
+
+The general format for data is
+folder:
+   [casename].raw
+   [casename]_load_time_series_MW.csv
+
+# Arguments
+- `directory::AbstractString`: directory containing RAW and CSV files
+
+# Examples
+```julia
+sys = parse_TAMU(
+    "./ACTIVSg25k",
+    configpath = "ACTIVSg25k_validation.json",
+    bus_name_formatter = x->string(x["name"]*"-"*string(x["index"])),
+    load_name_formatter = x->strip(join(x["source_id"], "_"))
+)
+```
+"""
+
+function TAMU_System(TAMU_FOLDER::AbstractString; kwargs...)
+    TAMU_CASE = basename(TAMU_FOLDER)
+    pm_data = PowerModelsData(joinpath(TAMU_FOLDER, TAMU_CASE * ".RAW"))
+
+    bus_name_formatter =
+        get(kwargs, :bus_name_formatter, x -> string(x["name"]) * "-" * string(x["index"]))
+    load_name_formatter =
+        get(kwargs, :load_name_formatter, x -> strip(join(x["source_id"], "_")))
+
+    # make system
+    sys = System(
+        pm_data,
+        bus_name_formatter = bus_name_formatter,
+        load_name_formatter = load_name_formatter;
+        kwargs...,
+    )
+
+    # add forecasts
+    header_row = 2
+
+    tamu_files = readdir(TAMU_FOLDER)
+    load_file = joinpath(
+        TAMU_FOLDER,
+        tamu_files[occursin.("_load_time_series_MW.csv", tamu_files)][1],
+    ) # currently only adding MW load forecasts
+    header = String.(split(open(readlines, load_file)[header_row], ","))
+    fixed_cols = ["Date", "Time", "Num Load", "Total MW Load", "Total Mvar Load"]
+    for l in header
+        l in fixed_cols && continue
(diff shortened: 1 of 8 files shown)


src/parsers/TAMU_data.jl
```diff
@@ -0,0 +1,108 @@
...
+sys = parse_TAMU(
+    "./ACTIVSg25k",
+    configpath = "ACTIVSg25k_validation.json",
+    bus_name_formatter = x->string(x["name"]*"-"*string(x["index"])),
+    load_name_formatter = x->strip(join(x["source_id"], "_"))
+)
+```
+"""
+
+function TAMU_System(TAMU_FOLDER::AbstractString; kwargs...)
```
daniel-thom, line 29: Even though it's a proper noun, I suggest that it's better to stick with naming conventions.  It's always predictable.  I would use `parse_tamu`, `TamuSystem`, `tamu_folder`.  At least `TAMUSystem`.

src/parsers/TAMU_data.jl
```diff
@@ -0,0 +1,108 @@
...
+    load_file = joinpath(
+        TAMU_FOLDER,
+        tamu_files[occursin.("_load_time_series_MW.csv", tamu_files)][1],
+    ) # currently only adding MW load forecasts
+
+    !isfile(load_file) && throw(DataFormatError("Cannot find $load_file"))
+
+    header = String.(split(open(readlines, load_file)[header_row], ","))
+    fixed_cols = ["Date", "Time", "Num Load", "Total MW Load", "Total Mvar Load"]
+    for l in header
```
daniel-thom, line 62: Many people use a rule where you're not allowed to use `l` as a variable name.  Looks too much like `1`.

src/parsers/TAMU_data.jl
```diff
@@ -0,0 +1,108 @@
...
+        l in fixed_cols && continue
+        lsplit = split(replace(string(l), "#" => ""), " ")
+        @assert length(lsplit) == 4
+        push!(fixed_cols, string("load_", join(lsplit[2:3], "_")))
+    end
+
+    loads = CSV.read(load_file, skipto = 3, header = fixed_cols)
+
+    function parse_datetime_ampm(ds::AbstractString, fmt::Dates.DateFormat)
+        m = match(r"(.*?)\s*(AM|PM)?$"i, ds)
```
daniel-thom, line 72: Usually `?` means 0 or 1 of the previous character.  So, I'm not sure I follow `.*?`.  Is that actually correct?

src/parsers/TAMU_data.jl
```diff
@@ -0,0 +1,108 @@
...
+            :Date,
+            :Time,
+            Symbol("Num Load"),
+            Symbol("Total MW Load"),
+            Symbol("Total Mvar Load"),
+        ]),
+    )
+        c = get_component(PowerLoad, sys, string(lname))
+        if !isnothing(c)
+            #@info "adding forecast for load $(lname)"
```
daniel-thom, line 96: stale info

src/parsers/TAMU_data.jl
```diff
@@ -0,0 +1,108 @@
...
+    end
+
+    dfmt = Dates.DateFormat("m/dd/yyy H:M:S")
+
+    loads[:timestamp] =
+        parse_datetime_ampm.(string.(loads[!, :Date], " ", loads[!, :Time]), dfmt)
+
+    for lname in setdiff(
+        Set(names(loads)),
+        Set([
```
daniel-thom, line 85: `setdiff` does not actually require that you create the intermediate sets.  Probably better if you don't.

src/parsers/TAMU_data.jl
```diff
@@ -0,0 +1,108 @@
...
+
+    !isfile(load_file) && throw(DataFormatError("Cannot find $load_file"))
+
+    header = String.(split(open(readlines, load_file)[header_row], ","))
+    fixed_cols = ["Date", "Time", "Num Load", "Total MW Load", "Total Mvar Load"]
+    for l in header
+        l in fixed_cols && continue
+        lsplit = split(replace(string(l), "#" => ""), " ")
+        @assert length(lsplit) == 4
+        push!(fixed_cols, string("load_", join(lsplit[2:3], "_")))
```
daniel-thom, line 66: Might be good to paste an example string that you're parsing.

claytonpbarrows:
> Usually `?` means 0 or 1 of the previous character.  So, I'm not sure I follow `.*?`.  Is that actually correct? 

I don't really understand regexs, but without the if I replace `.*?` with `.*` the command doesn't work.

src/parsers/TAMU_data.jl
claytonpbarrows: I went with `TamuSystem` and `tamu_folder`.

src/parsers/TAMU_data.jl
daniel-thom: I think this is what you want:
```
julia> m = match(r"(.*)\s(AM|PM)", s)
RegexMatch("1/1/2016,12:00:00 AM", 1="1/1/2016,12:00:00", 2="AM")

julia> m.captures[1]
"1/1/2016,12:00:00"

julia> m.captures[2]
"AM"
```

PR #457, opened 2020-04-22
Jd/updates for psi agc

jd-lara:
The CI from Travis is acting out a bit.

### DIFF
--- a/src/models/devices.jl
+++ b/src/models/devices.jl
@@ -173,7 +173,7 @@ function IS.convert_type(::Type{Device}, data::Any)
     error("This form of convert_type is not supported for Devices")
 end
 
-function has_forecast(d::T) where {T <: Component}
-    !hasfield(T, :forecasts) && return false
+function has_forecasts(d::T) where {T <: Component}
+    hasmethod(get_forecast, Tuple{T}) && return false
     return IS.has_forecasts(d)
 end
--- a/src/models/devices.jl
+++ b/src/models/devices.jl
@@ -173,7 +173,6 @@ function IS.convert_type(::Type{Device}, data::Any)
     error("This form of convert_type is not supported for Devices")
 end
 
-function has_forecasts(d::T) where {T <: Component}
-    hasmethod(get_forecast, Tuple{T}) && return false
+function has_forecasts(d::Component)
     return IS.has_forecasts(d)
 end
--- a/src/models/devices.jl
+++ b/src/models/devices.jl
@@ -173,7 +173,7 @@ function IS.convert_type(::Type{Device}, data::Any)
     error("This form of convert_type is not supported for Devices")
 end
 
-function has_forecast(d::T) where {T <: Device}
+function has_forecast(d::T) where {T <: Component}
     !hasfield(T, :forecasts) && return false
     return IS.has_forecasts(d)
 end
--- a/src/models/devices.jl
+++ b/src/models/devices.jl
@@ -175,5 +175,5 @@ end
 
 function has_forecast(d::T) where {T <: Device}
     !hasfield(T, :forecasts) && return false
-    return !isempty(d.forecasts)
+    return IS.has_forecasts(d)
 end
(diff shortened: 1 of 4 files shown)


src/models/devices.jl
```diff
@@ -172,3 +172,8 @@ end
 function IS.convert_type(::Type{Device}, data::Any)
     error("This form of convert_type is not supported for Devices")
 end
+
+function has_forecast(d::T) where {T <: Device}
```
daniel-thom, line 176: You should be able to use `IS.has_forecasts(d)`

src/utils/IO/system_checks.jl
```diff
@@ -59,7 +59,7 @@ function critical_components_check(sys::System)
         end
     end
     missing_critical_components &&
-    throw(IS.InvalidValue("Critical Componeents are not present."))
+        throw(IS.InvalidValue("Critical Componeents are not present."))
```
daniel-thom, line 62: Componeents is misspelled

src/models/devices.jl
```diff
@@ -172,3 +172,8 @@ end
...
+    !hasfield(T, :forecasts) && return false
```
daniel-thom, line 177: I think that you shouldn't do this and instead, rely completely on IS.has_forecasts.  That function relies the the `get_forecasts` method, so the user can use whatever field name they want.  It also has allows for legacy code that defined the field `_forecasts`, which some people may have copied and used.  

For consistency I also think it would be better to call the function the same name as the IS function since they do the same thing.

Finally, I'm not sure your choice of `Device` is correct.  Shouldn't it be `Component`?  LoadZone has forecasts.

jd-lara: My interpretation of what `IS.has_forecasts()` does, it requires that the component has the forecast field. Also, I think all `_forecast` has been removed. 

You are correct about using it for PSY.Component.

daniel-thom: No, it calls get_forecasts.  I know that we have changed all of our names.  I'm concerned about external users.
```
function has_forecasts(component::InfrastructureSystemsType)
    container = get_forecasts(component)
    return !isnothing(container) && !isempty(container)
end

```

daniel-thom: Anyway, we shouldn't require specific field names.  Accessor functions are a better way to go.

jd-lara: What is your plan for components that don't have a `get_forecast` method implemented?

jd-lara: `hasmethod(get_forecast, Tuple{T})` ?

jd-lara: see cadc7ac

daniel-thom: Sorry, I forgot to mention this catch-all function in IS.

```
"""
Return the internal forecast storage container or nothing, if the type doesn't store
forecasts.

Subtypes need to implement this method if they store forecasts.
"""
function get_forecasts(value::T) where {T <: InfrastructureSystemsType}
    # Older code relied on a field called _forecasts. Keep this until other packages comply
    # with the new requirement for implementing this method.
    return hasfield(T, :_forecasts) ? value._forecasts : nothing
end
```

jd-lara: we should remove that method now since it isn't required anymore.

jd-lara: https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/74

daniel-thom: The IS function `get_forecasts` provides default behavior so that it works even if a type doesn't contain forecasts.  It also allow `has_forecasts` to work without using introspection to find out if a method is implemented.  Do you disagree with this approach?  I think it works nicely.

jd-lara: The implementation of `get_forecasts()` in IS right now will return nothing if the methods isn't implemented. It isn't obvious but it's fine we can leave as it is I don't want to delay this anymore.

jd-lara: see 4ae3f83

PR #459, opened 2020-04-24
add the get_available components method

jd-lara:


src/base.jl
```diff
@@ -549,6 +549,15 @@ function get_components_by_name(
     return IS.get_components_by_name(T, sys.data, name)
 end
 
+
+"""
+    gets_components_availability. Requires type T to have the method get_available implemented
```
daniel-thom, line 554: Use of underscores in `gets_components_availability` implies...something?

PR #463, opened 2020-04-29
Fix serialization of DynamicInverter

daniel-thom:
This will be broken until https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/78 is merged and a new version is tagged.


test/common.jl
```diff
@@ -59,3 +59,126 @@ function get_branch(sys::System, other::Branch)
 
     error("Did not find branch with buses $(other.arc.from.name) ", "$(other.arc.to.name)")
 end
+
+function create_system_with_dynamic_inverter()
```
daniel-thom, line 63: @jd-lara  I moved your code so that I wouldn't have to duplicate all of it.  If you prefer that I didn't do this, I can revert it.

src/models/dynamic_inverter.jl
```diff
@@ -26,71 +26,71 @@ mutable struct DynamicInverter{
...
 
-        n_states = (
-            converter.n_states +
-            outer_control.n_states +
-            inner_control.n_states +
-            dc_source.n_states +
-            freq_estimator.n_states +
-            filter.n_states
-        )
+function DynamicInverter(
```
daniel-thom, line 31: The old inner constructor prevented me from being able to construct an object with a deserialized internal object.

test/common.jl
jd-lara: This is fine

src/models/dynamic_inverter.jl
jd-lara: We should make consistent with the constructor for DynamicGenerator

jd-lara:
@daniel-thom is this still a draft?

src/models/dynamic_inverter.jl
daniel-thom: Fixed

PR #465, opened 2020-04-30
Add methods for get_forecast_values

daniel-thom:
I did not add this option described in the story:

```
One more optional thing to consider is making the forecast type be default to Deterministic and leave it to the modeler/user to get a different type of forecast if present.
```

I think that doing this would mean that we change all forecast methods.  Would you want to do that?

### DIFF
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -177,7 +177,6 @@ export get_components_by_name
 export get_available_components
 export get_forecast_labels
 export get_forecast_initial_times
-export get_forecast_keys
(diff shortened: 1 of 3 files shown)


jd-lara:
> I did not add this option described in the story:
> 
> ```
> One more optional thing to consider is making the forecast type be default to Deterministic and leave it to the modeler/user to get a different type of forecast if present.
> ```
> 
> I think that doing this would mean that we change all forecast methods. Would you want to do that?

That's fine, we can delay that for later

PR #467, opened 2020-04-30
Change errors to warnings

daniel-thom:


src/utils/IO/system_checks.jl
```diff
@@ -113,7 +109,12 @@ Total sum of system generator ratings.
 - `sys::System`: system
 """
 function total_generation_rating(sys::System)
```
jd-lara, line 111: we should add storage here

daniel-thom: Should the name of this function change?

jd-lara: total_capacity_rating

claytonpbarrows: Ok, I support adding storage and changing to total_capacity_rating

PR #470, opened 2020-05-04
PSIMA-142:Adding ORDC Reserve

sourabhdalvi:
- Need to add test for this struct


src/PowerSystems.jl
```diff
@@ -126,6 +126,7 @@ export ReserveUp
 export ReserveDown
 export StaticReserve
 export VariableReserve
+export ORDC
```
jd-lara, line 129: Should choose a more descriptive name

sourabhdalvi: I know but `OperatingReserveDemandCurve` seem bit too long, so I would like some suggesting from you guys. May be just `ReserveDemandCurve` ?

jd-lara: I like `ReserveDemandCurve`

PR #471, opened 2020-05-05
Jd/pwl improvements

jd-lara:
some improvements in getter functions to manipulate PWL cost functions


claytonpbarrows:
Otherwise LGTM

PR #473, opened 2020-05-06
Regenerate structs with updated script

daniel-thom:


jd-lara:
@daniel-thom I approved but it seems that it will need the changes in IS. I will merge to master so that the IS tests pass and then re-trigger the tests

PR #475, opened 2020-05-08
modifying the pwl cost function to work with PSI

sourabhdalvi:


test/test_cost_functions.jl
```diff
@@ -1,7 +1,7 @@
 using PowerSystems
 cost = VariableCost([(1.0, 1.0), (2.0, 1.1), (3.0, 1.2)])
 slopes = get_slopes(cost)
-res = [1.0, 10.0, 10.0]
+res = [10.0, 10.0]
```
jd-lara, line 4: did you change it to return n - 1 slopes? Look at the docstring. I purposely made it for it to return n slopes. What's the reason for the change?

sourabhdalvi: Sorry about, I did change it back to normal

sourabhdalvi:
> please explain the update.

PR #476, opened 2020-05-08
Add field types to setter functions

daniel-thom:


jd-lara:
I will merge such that the IS tests pass

PR #477, opened 2020-05-08
Add connections between static and dynamic injectors

daniel-thom:


src/models/dynamic_generator.jl
```diff
@@ -5,9 +5,7 @@ mutable struct DynamicGenerator{
     TG <: TurbineGov,
     P <: PSS,
 } <: DynamicInjection
-    number::Int64
-    name::String
-    bus::Bus
+    static_injector::Union{Nothing, Generator}
     ω_ref::Float64
     V_ref::Float64
     P_ref::Float64
```
jd-lara, line 11: We can remove P_ref and V_ref also. Those need to match the injector active power and bus Voltage respectively.

test/common.jl
```diff
@@ -169,14 +173,9 @@ function create_system_with_dynamic_inverter()
...
-    for bus in nodes_OMIB
-        add_component!(sys, bus)
-    end
-    for lines in branch_OMIB
-        add_component!(sys, lines)
-    end
-    add_component!(sys, battery)
+    @test get_V_ref(test_inverter) == 1.045
+    @test get_P_ref(test_inverter) == 0.01375
+    @test get_Q_ref(test_inverter) == 0.0
```
daniel-thom, line 178: @jd-lara I think the old code had incorrect values that were passed to DynamicInverter.

jd-lara: It might be the case. I think that we never checked that the data in the static devices matched the refs in the dynamic ones.

PR #492, opened 2020-05-21
remove get_tech call

jd-lara:


src/models/generation.jl
```diff
@@ -14,6 +14,5 @@ function IS.get_limits(
     return (min = valid_range.min, max = valid_range.max, zero = 0.0)
 end
 
-get_rating(gen::Generator) = get_rating(get_tech(gen))
-get_active_power_limits_min(gen::Generator) = get_activepowerlimits(get_tech(gen)).min
-get_active_power_limits_max(gen::Generator) = get_activepowerlimits(get_tech(gen)).max
+get_active_power_limits_min(gen::Generator) = get_activepowerlimits(gen).min
```
daniel-thom, line 17: If this won't work for all generators then I don't think it should be done this way.  We should implement one method per subtype that supports it.

jd-lara: I agree, I made a "quick" fix and it opened a whole can of worms.

jd-lara:
On a second thought @daniel-thom I think we should drop these methods

PR #493, opened 2020-05-26
PSIMA-62:Adding Thermal PGLIB struct

sourabhdalvi:


src/descriptors/power_system_structs.json
```diff
@@ -1860,6 +1921,181 @@
...
+                "null_value": "nothing",
+                "data_type": "Union{Nothing, NamedTuple{(:up, :down), Tuple{Float64, Float64}}}",
+                "valid_range": {
+                    "min": 0,
+                    "max": null
+                },
+                "validation_action": "error"
+            },
+            {
+                "name": "starttimelimits",
```
daniel-thom, line 2028: Can you use `snake_case` for all new fields?  These are hard to read.

sourabhdalvi: yup

src/descriptors/power_system_structs.json
```diff
@@ -92,6 +92,67 @@
         ],
         "supertype": "OperationalCost"
     },
+    {
+        "struct_name": "PGLIBCost",
+        "docstring": "Data Structure Operational Cost Data in Three parts fixed, variable cost and start - stop costs.",
```
jd-lara, line 97: Docstring is outdated.

src/descriptors/power_system_structs.json
```diff
@@ -92,6 +92,67 @@
         ],
         "supertype": "OperationalCost"
     },
+    {
+        "struct_name": "PGLIBCost",
```
jd-lara, line 96: This should MultiStageCost or something along those lines to generalize better

sourabhdalvi: I named it MultiStartCost & ThermalMultiStart, I know these are probably not the best names I could come up with but open to any suggestions

src/models/generated/MultiStartCost.jl
```diff
@@ -0,0 +1,91 @@
...
+    MultiStartCost(variable, no_load, fixed, startup, shutdn, forecasts, )
+end
+
+# Constructor for demo purposes; non-functional.
+function MultiStartCost(::Nothing)
+    MultiStartCost(;
+        variable=VariableCost((0.0, 0.0)),
+        no_load=0.0,
+        fixed=0.0,
+        startup=(hot = Inf, warm = Inf,cold = Inf),
```
jd-lara, line 57: Inf as default?

src/models/generated/ThermalMultiStart.jl
```diff
@@ -0,0 +1,237 @@
...
+        bus::Bus
+        activepower::Float64
+        reactivepower::Float64
+        rating::Float64
+        primemover::PrimeMovers.PrimeMover
+        fuel::ThermalFuels.ThermalFuel
+        activepowerlimits::Min_Max
+        reactivepowerlimits::Union{Nothing, Min_Max}
+        ramplimits::Union{Nothing, NamedTuple{(:up, :down), Tuple{Float64, Float64}}}
+        power_trajectory::Union{Nothing, NamedTuple{(:startup, :shutdown), Tuple{Float64, Float64}}}
```
jd-lara, line 18: Can you please add a comment about what this field is?

src/models/generated/ThermalMultiStart.jl
```diff
@@ -0,0 +1,237 @@
...
+end
+
+function ThermalMultiStart(; name, available, status, bus, activepower, reactivepower, rating, primemover, fuel, activepowerlimits, reactivepowerlimits, ramplimits, power_trajectory, timelimits, start_time_limits, start_types, op_cost, services=Device[], participation_factor=0.0, time_at_status=INFINITE_TIME, must_run=false, dynamic_injector=nothing, ext=Dict{String, Any}(), forecasts=InfrastructureSystems.Forecasts(), )
+    ThermalMultiStart(name, available, status, bus, activepower, reactivepower, rating, primemover, fuel, activepowerlimits, reactivepowerlimits, ramplimits, power_trajectory, timelimits, start_time_limits, start_types, op_cost, services, participation_factor, time_at_status, must_run, dynamic_injector, ext, forecasts, )
+end
+
+# Constructor for demo purposes; non-functional.
+function ThermalMultiStart(::Nothing)
```
jd-lara, line 110: missing `basepower::Float64` field

src/models/generated/MultiStartCost.jl
daniel-thom: I don't think anyone has fixed the issue with JSON-serializing `Inf` and `NaN`.

sourabhdalvi: I didn't want it to be zero but a large number, as there are cases where a device only has 1 or 2 types of starts.

sourabhdalvi: Thing I'll do is add a const `START_COST` and use it instead, which would be a reasonably large number

src/models/generated/ThermalMultiStart.jl
sourabhdalvi: will do

src/models/generated/MultiStartCost.jl
jd-lara: we should then add an individual field for each and make a constant for "high_cost" I am not sure to want Inf to be the value since it might not be interpretable by the solvers.

src/parsers/power_system_table_data.jl
```diff
@@ -819,6 +821,112 @@ function make_thermal_generator(data::PowerSystemTableData, gen, cost_colnames,
...
 
+function make_thermal_generator_multistart(
+    data::PowerSystemTableData,
+    gen,
+    cost_colnames,
+    bus,
+)
+    fuel_cost = gen.fuel_price / 1000
+
+    var_cost = [(getfield(gen, hr), getfield(gen, mw)) for (hr, mw) in cost_colnames]
```
daniel-thom, line 832: Is the calculation of var_cost a direct copy/paste of the similar code in make_thermal_generator?  If so, could you make it a common function?

sourabhdalvi: Sure

src/parsers/power_system_table_data.jl
```diff
@@ -760,6 +762,12 @@ function make_thermal_generator(data::PowerSystemTableData, gen, cost_colnames,
         var_cost = [(0.0, var_cost[1][2]), (1.0, var_cost[1][2])]
         fixed = 0.0
     end
+    return var_cost, fixed
+end
+
+function make_thermal_generator(data::PowerSystemTableData, gen, cost_colnames, bus)
+    fuel_cost = gen.fuel_price / 1000
```
jd-lara, line 769: Why division over 1000. Also, don't divide over an integer.

src/parsers/power_system_table_data.jl
```diff
@@ -819,6 +827,78 @@ function make_thermal_generator(data::PowerSystemTableData, gen, cost_colnames,
...
 end
 
+function make_thermal_generator_multistart(
+    data::PowerSystemTableData,
+    gen,
+    cost_colnames,
+    bus,
+)
+    var_cost, fixed = calculate_variable_cost(data, gen, cost_colnames)
+    fuel_cost = gen.fuel_price / 1000
```
jd-lara, line 837: same here

src/parsers/power_system_table_data.jl
daniel-thom: You are computing fuel_cost twice. Perhaps return it from `calculate_variable_cost`?

PR #495, opened 2020-05-27
Include machine base power in generator's structs

rodrigomha:
The following PR provides an additional field for generators, namely `machine_basepower`, that stores the machine base power in MVA.

This field is provided in both .m and .raw data files for each generator, and is parsed by PowerModels as `"mbase"` in the dictionary, and saved in the NamedTuple as `base_mva`.  The change proposed here simply store this value in an additional field in each generator struct.

This value usually defaults to 100.0 MVA in many datasets since it is not relevant for steady-state analysis. However, for dynamic simulations, active and reactive powers must be treated in machine/device per unit, and values in system per unit must be transformed to properly work in dynamic simulations.

The new field is usually include after `op_cost` or after `timelimits`, depending on the specific generator struct.

Tests seems to be passing locally, with the exception of InfrastructureSystems one: `test_generate_structs.jl`.


jd-lara:
@claytonpbarrows we need to decide to keep the base MVA of the machine in MVA units or normalized by the system's base.

src/models/generated/RenewableDispatch.jl
```diff
@@ -53,6 +55,8 @@ mutable struct RenewableDispatch <: RenewableGen
     powerfactor::Float64
     "Operation Cost of Generation [`TwoPartCost`](@ref)"
     op_cost::TwoPartCost
+    "Base power of the unit in MVA"
+    machine_basepower::Float64
```
jd-lara, line 59: I am not sure we should call it machine_basepower in this case since it could also be an Inverter

src/descriptors/power_system_structs.json
```diff
@@ -1319,6 +1319,17 @@
                 "data_type": "TwoPartCost",
                 "comment": "Operation Cost of Generation [`TwoPartCost`](@ref)"
             },
+            {
+                "name": "machine_basepower",
```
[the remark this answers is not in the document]
rodrigomha: Changed to `basepower` in 5ef7ef2

src/descriptors/power_system_structs.json
```diff
@@ -1319,6 +1319,17 @@
                 "data_type": "TwoPartCost",
                 "comment": "Operation Cost of Generation [`TwoPartCost`](@ref)"
             },
+            {
```
daniel-thom, line 1322: Should this value default to 100.0 so that users are not required to provide it?  (for all devices)

jd-lara: There shouldn't be a default for this. It only makes sense for very little systems, but the data in all formats has this column

rodrigomha: It is possible. I didn't include it as a default since the parameter is available in the datasets. However, if you check those columns in .m files or in .raw files, you most likely see everything at 100.0.

It's up to the discussion. The alternative is default to 100.0, and in the parsing, if it's available, do the update. Considering the parsing available, this will be updated all the time. The key difference will be when constructing it directly via the struct itself.

daniel-thom: One of my thoughts is that we are continually breaking our users.  Maybe this is OK until we tag 1.0.

jd-lara: @daniel-thom agreed, this is why we have time until September to tag 1.0 hopefully having developed most the code for operations + dynamics will allow covering most of the requirements.

PR #500, opened 2020-06-01
Update Dynamic Structs

rodrigomha:
The following PR provides an update on the dynamic structures for better integration to the dynamic models.

The main changes are:

- Add `V_ref` to the AVR models and OuterControl models. This is a parameter that defaults to 1.0, and must be updated when initializing the dynamic model via the solution of the static Power Flow.
- Add `P_ref` to the TurbineGov models and OuterControl models. Also defaults to 1.0.
- Remove `MVABase` from machine. This parameter should be the same as the base power provided in the static generator structs.

Note that `V_ref` and `P_ref` are not necessarily the solution of the static power flow model. These parameters will depend on the specific dynamic models used, and must be updated in order to match solutions of the static PF with the initial condition of the dynamic model. The user should not (but could) interfere directly.

### DIFF
--- a/src/descriptors/power_system_structs.json
+++ b/src/descriptors/power_system_structs.json
@@ -2333,10 +2333,9 @@
         "docstring": "Parameters of a AVR that returns a fixed voltage to the rotor winding",
         "fields": [
             {
-                "name": "V_ref",
-                "comment": "Reference Voltage Set-point",
+                "name": "Vf",
+                "comment": "Fixed voltage field applied to the rotor winding",
                 "null_value": 0,
-                "default": "1.0",
                 "data_type": "Float64",
                 "valid_range": {
                     "min": 0,
--- a/src/descriptors/power_system_structs.json
+++ b/src/descriptors/power_system_structs.json
@@ -2333,9 +2333,10 @@
         "docstring": "Parameters of a AVR that returns a fixed voltage to the rotor winding",
         "fields": [
             {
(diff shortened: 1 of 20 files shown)


src/descriptors/power_system_structs.json
```diff
@@ -2333,9 +2333,10 @@
         "docstring": "Parameters of a AVR that returns a fixed voltage to the rotor winding",
         "fields": [
             {
-                "name": "Emf",
-                "comment": "Fixed voltage to the rotor winding",
+                "name": "V_ref",
```
jd-lara, line 2336: I am not sure we should move to use V_ref for the fix AVR since it is meant to represent a fixed emf even if the voltage at the terminals can't be kept. I think this is a different model that can't guarantee to keep the voltage at a PV bus at the set level.

jd-lara:
The issue with the fixed AVR is that V_ref is set to a default of 1.0 and also has a null value so the method for null is overwriting the default constructor. I see no reason for this AVR to have a default value.

src/descriptors/power_system_structs.json
rodrigomha: Yes. I agree with that. Will revert the change, but modify the name to Vf, that is the standard name used in LITS.

rodrigomha: Addressed via abdb0fa

PR #501, opened 2020-06-02
Fix serialization for AGC and RegulationDevice

daniel-thom:
This will break deserialization from previous versions because we were including an Area inside a Bus JSON object instead of its UUID.


src/models/regulation_device.jl
```diff
@@ -5,7 +5,7 @@ mutable struct RegulationDevice{T <: StaticInjection} <: Device
     reserve_limit::Float64
     inertia::Float64
 
-    function RegulationDevice(
+    function RegulationDevice{T}(
```
jd-lara, line 8: We should maintain a method that isn't parametrized to create regulation devices easily

PR #503, opened 2020-06-03
Fix serialization of area and load_zone in buses

daniel-thom:
This will definitely break deserialization from previous versions because we were storing areas and load zones as objects instead of UUIDs in the JSON.  We were also allowing addition of buses if attached areas and load zones were not part of the system.


jd-lara:
I am not too concerned with back-compatibility of serialization until 1.0

PR #504, opened 2020-06-04
Use copy_forecasts function to fix convert_component!

daniel-thom:
This existing function does not work if the component has forecasts.

This is dependent on IS changes in https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/97

Let me know if you think `copy_forecasts` should be an exported PSY function.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -856,6 +856,26 @@ function add_forecast!(
 end
 
 """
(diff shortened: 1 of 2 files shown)


daniel-thom:
This is updated with new behaviors from our discussion today.

PR #505, opened 2020-06-04
Add validation descriptors for RegulationDevice

daniel-thom:
This depends on changes in IS in https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/98

### DIFF
--- a/src/descriptors/power_system_structs.json
+++ b/src/descriptors/power_system_structs.json
@@ -1,5034 +1,5077 @@
-[
+{
+  "auto_generated_structs": [
     {
-        "struct_name": "TwoPartCost",
-        "docstring": "Data Structure Operational Cost Data in two parts: fixed and variable cost.",
-        "fields": [
-            {
-                "name": "variable",
-                "comment": "variable cost",
-                "null_value": "VariableCost((0.0, 0.0))",
-                "data_type": "VariableCost"
-            },
-            {
-                "name": "fixed",
-                "comment": "fixed cost",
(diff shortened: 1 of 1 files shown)


src/models/generated/FullMachine.jl
```diff
@@ -99,14 +99,14 @@ function FullMachine(::Nothing)
...
-        L_1d=1.0,
-        L_1q=1.0,
+        L_d=1,
+        L_q=1,
+        L_ad=2,
+        L_aq=2,
+        L_f1d=1,
+        L_ff=2,
+        L_1d=1,
+        L_1q=1,
```
daniel-thom, line 109: This is an error.  I used `jq` to format the JSON and it converted 2.0 to 2.  I'll fix it

src/descriptors/power_system_structs.json
```diff
@@ -1,5034 +1,5077 @@
...
+          "name": "states",
+          "comment": "",
+          "internal_default": "[:δ, :ω]",
+          "data_type": "Vector{Symbol}"
+        },
+        {
+          "name": "n_states",
+          "comment": "",
+          "internal_default": 2,
+          "data_type": "Int64"
```
jd-lara, line 5047: max should be 1.0

jd-lara, line 5063: I am not sure if we can implement (not urgent) that but it should be less than rating in the internal ThermalStandard

jd-lara:
I will correct the values in my branch

PR #507, opened 2020-06-07
[DNMY] Add GENROU/SAL models

rodrigomha:
The following PR includes both GENROU and GENSAL data structs used in PSS/E. The main differences with the data provided in the Reference Manual of PSS/E are:
- Machine models does not include the angle and rotor speed. Those are computed in the shaft.
- The armature resistance is not provided as a parameter of the GENROU/SAL models in the reference manual. Model descriptions in Paszek book: Synchronous Generators and Excitation Systems Operating in a Power System include the armature resistance, so for completeness we decided to include it. We can default it to zero already in the struct definition or when doing the parsing. We are not sure how PSS/E provides the armature resistance (or if it's actually modeled).

We decided to include states doc-strings. This is a feature in progress, and eventually will be added to all dynamic structures.

### DIFF
--- a/src/models/generated/GENROU.jl
+++ b/src/models/generated/GENROU.jl
@@ -36,7 +36,7 @@ Parameters of 4-states round-rotor synchronous machine: GENROU model
 - `Xq_p::Float64`: Transient reactance after EMF in q-axis per unit, validation range: (0, nothing)
 - `Xd_pp::Float64`: Sub-Transient reactance after EMF in d-axis per unit. Note: Xd_pp = Xq_pp, validation range: (0, nothing)
 - `Xl::Float64`: Stator leakage reactance, validation range: (0, nothing)
-- `S1_0::Float64`: Saturation factor at 1 pu flux: S(1.0) = B(|ψ_pp|-A)^2 -> S(1.2) = B(1.2-A)^2, validation range: (0, nothing)
+- `S1_0::Float64`: Saturation factor at 1 pu flux: S(1.0) = B(|ψ_pp|-A)^2 -> S(1.0) = B(1.0-A)^2, validation range: (0, nothing)
 - `S1_2::Float64`: Saturation factor at 1.2 pu flux: S(1.2) = B(|ψ_pp|-A)^2 -> S(1.2) = B(1.2-A)^2, validation range: (0, nothing)
 - `ext::Dict{String, Any}`
 - `states::Vector{Symbol}`: The states are:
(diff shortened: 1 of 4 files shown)


src/models/generated/GENROU.jl
```diff
@@ -0,0 +1,185 @@
+#=
+This file is auto-generated. Do not edit.
+=#
+"""
+    mutable struct GENROU <: Machine
```
jd-lara, line 5: The names GenSal, GenRou we should include them in the docstrings as the name in PSS/e but we need to give them a proper name (SalientPoleExponentialSaturation or something like that) and add a reference to Sauer Pai.

src/models/generated/GENROU.jl
```diff
@@ -0,0 +1,185 @@
...
+    Xq::Float64
+    "Transient reactance after EMF in d-axis per unit"
+    Xd_p::Float64
+    "Transient reactance after EMF in q-axis per unit"
+    Xq_p::Float64
+    "Sub-Transient reactance after EMF in d-axis per unit. Note: Xd_pp = Xq_pp"
+    Xd_pp::Float64
+    "Stator leakage reactance"
+    Xl::Float64
+    "Saturation factor at 1 pu flux"
```
jd-lara, line 73: We need to document better the S1 and s0 factors

rodrigomha:
> I haven't looked into the other structs but can you please confirm that the nomenclature is consistent with the other machines?

We are consistent with the other machines. However, the order of parameters in the struct is not consistent. We decided to go with this order given that is the same of the one provided in the PPS/E manual.

src/models/generated/GENROU.jl
rodrigomha: I'm not sure exactly if I agree in this one. GENROU, GENSAL and those naming are in a way the standard approach of naming those. I understand that here we are only referring to the machine itself and not the generator, but going with this naming could be confusing. Let's discuss more on this.

src/models/generated/GENROU.jl
rodrigomha: See changes in 78eca89

rodrigomha: We should store S in a tuple and saturation_coefficients (A,B) also in a tuple. We should include in supplemental_constructors how to obtain those coefficients.

rodrigomha:
Ton of changes to be done. I'll close the PR and open a new one.

PR #508, opened 2020-06-09
Add Salient Pole and Round Rotor machine dynamic models

rodrigomha:
The following PR provides the two struct used for standard Round Rotor and Salient Pole (GENROU and GENSAL) models.

The saturation coefficients are computed using a specific function, that solves a quadratic equation directly.

### DIFF
--- a/src/models/dynamic_generator_components.jl
+++ b/src/models/dynamic_generator_components.jl
@@ -4,3 +4,32 @@ abstract type Machine <: DynamicGeneratorComponent end
 abstract type PSS <: DynamicGeneratorComponent end
 abstract type Shaft <: DynamicGeneratorComponent end
 abstract type TurbineGov <: DynamicGeneratorComponent end
+
+"""
+Obtain coefficients (A, B) of the function Se = B(x - A)^2 for
+Se(1.2) = B(1.2 - A)^2 and Se(1.0) = B(1.2 - A)^2 as:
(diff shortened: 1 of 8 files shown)


src/models/dynamic_generator_components.jl
```diff
@@ -4,3 +4,32 @@ abstract type Machine <: DynamicGeneratorComponent end
...
+(1.2 - A)^2 Se(1.0) = Se(1.2) * (1.0 - A)^2 or expanding:
+(Se(1.2) - Se(1.0)) A^2 + (2.4 Se(1.0) - 2 Se(1.2)) A + (Se(1.2) - 1.44 Se(1.0)) = 0
+and uses the negative solution of the quadratic equation 
+"""
+function get_quadratic_saturation(Se::Tuple{Float64, Float64})
+    if Se[1] == 0.0 && Se[2] == 0.0
+        return (0.0, 0.0)
+    end
+    if Se[2] <= Se[1]
+        print("Se(1.2) <= Se(1.0). Ignoring saturation")
```
rodrigomha, line 21: This should be a warning. But we still not sure if we should check this on other form. We should probably check before this case.

jd-lara:
```suggestion
        throw(IS.ConflictingInputsError("Se(1.2) <= Se(1.0). Saturation Data is incosistent"))
```

jd-lara: @daniel-thom opinions here?

PR #509, opened 2020-06-09
Power flow fix

jd-lara:
This PR partially resolves some performance issues with the PF solution using Finite Differences.


src/utils/power_flow/power_flow.jl
```diff
@@ -99,8 +99,25 @@ function _write_pf_sol!(sys::System, nl_result)
...
 
+# TODO: Apply actions according to load type
+function _get_load_data(sys::System, b::Bus)
+    activepower = 0.0
+    reactivepower = 0.0
+    for l in get_components(ElectricLoad, sys)
+        if !isa(l, FixedAdmittance) && (l.bus == b)
+            activepower += get_activepower(l)
+            reactivepower += get_reactivepower(l)
+        elseif isa(l, FixedAdmittance) && (l.bus == b)
```
daniel-thom, line 110: This calls `isa` on l twice and compares the buses twice.  You could easily refactor the code to call them each only once.

src/utils/power_flow/power_flow.jl
```diff
@@ -139,12 +156,121 @@ solve_powerflow!(sys, nlsolve, method = :newton)
...
+            if gen.bus == b
+                P_GEN_BUS[ix] += get_activepower(gen)
+                Q_GEN_BUS[ix] += get_reactivepower(gen)
+            end
+        end
+
+        P_LOAD_BUS[ix], Q_LOAD_BUS[ix] = _get_load_data(system, b)
+
+        if b.bustype == BusTypes.REF
+            injection_components = get_components(StaticInjection, system, d -> d.bus == b)
```
daniel-thom, line 206: A potential optimization is to create a `set` of buses attached to StaticInjection devices ahead of time.  This will iterate over the same components many times.

src/utils/power_flow/power_flow.jl
```diff
@@ -139,12 +156,121 @@ solve_powerflow!(sys, nlsolve, method = :newton)
...
+        elseif b.bustype == BusTypes.PV
+            Vm[ix] = get_voltage(b)::Float64
+            x0[state_variable_count] = Q_GEN_BUS[ix]
+            x0[state_variable_count + 1] = bus_angle
+            state_variable_count += 2
+        elseif b.bustype == BusTypes.PQ
+            x0[state_variable_count] = bus_voltage
+            x0[state_variable_count + 1] = bus_angle
+            state_variable_count += 2
+        end
```
daniel-thom, line 223: else, `error()`

src/utils/power_flow/power_flow.jl
jd-lara: There are other improvements missing that we can add in a second round. This will speed up the process of setting up the problem which is not currently a big bottle neck.

PR #514, opened 2020-06-15
Dynamic Data Parsing

jd-lara:


src/parsers/psse_dynamic_mapping.yml
```diff
@@ -0,0 +1,37 @@
+Machine:
+- {
+  GENROU: '(RoundRotorQuadratic, RoundRotorShaft)',
```
claytonpbarrows, line 3: suggest:
GENROU:[
machine: 'RoundRotorQuadratic',
shaft: 'roundRotorShaft']

rodrigomha:
The PR is ready for review.

We should check:
- Naming used for functions. Not sure if `parse_dyn_system` and `add_dyn_injectors!` are proper naming.
- Test used to check dyn parse. We can probably expand later testing when adding more dyr examples.
- Overall performance.

PR #515, opened 2020-06-16
Add new AVR models

rodrigomha:
The following PR will add new AVR struct models.

A key discussion must be done for parameters that implicitly define different different differential equations (and hence a different model) depending on if some time constants are set to zero or not.

To do:
- [x] Add ESST1A struct (ESAC1 and ESST1A are the only ones required for the Australian case)
- [x] Add EXAC1 struct
- [x] Add GAST struct
- [x] Set-up tests


src/models/generated/includes.jl
```diff
@@ -95,6 +103,7 @@ export get_L_d
 export get_L_f1d
 export get_L_ff
 export get_L_q
+export get_PSS_inputs
```
jd-lara, line 106: why PSS_inputs?

src/models/generated/ST1A.jl
```diff
@@ -0,0 +1,233 @@
...
+        Tf=0,
+        K_lr=0,
+        I_lr=0,
+        V_ref=0,
+        ext=Dict{String, Any}(),
+    )
+end
+
+"""Get ST1A UEL_inputs."""
+get_UEL_inputs(value::ST1A) = value.UEL_inputs
```
jd-lara, line 146: do we need these here or are they inputs from an external component?

rodrigomha: UEL inputs and PSS inputs are the flags used to determine on which the UEL and PSS signals enter in the AVR. This are flags of the AVR model, and hence must be parameters (or flags) of the AVR.

I'm thinking of renaming to UEL_flags and PSS_flags.

rodrigomha: See ST1A in here:
https://www.powerworld.com/files/Block-Diagrams-17.pdf

jd-lara: ok agreed this needs to be a parameter.

src/models/generated/AC1A.jl
```diff
@@ -0,0 +1,225 @@
...
+# Arguments
+- `Tr::Float64`: Regulator input filter time constant in s, validation range: (0, nothing)
+- `Tb::Float64`: Regulator denominator (lag) time constant in s, validation range: (0, nothing)
+- `Tc::Float64`: Regulator numerator (lead) time constant in s, validation range: (0, nothing)
+- `Ka::Float64`: Regulator output gain, validation range: (0, nothing)
+- `Ta::Float64`: Regulator output time constant in s, validation range: (0, nothing)
+- `Va_lim::Tuple{Float64, Float64}`: Limits for regulator output (Va_min, Va_max)
+- `Te::Float64`: Exciter field time constant in s, validation range: (&quot;eps()&quot;, nothing)
+- `Kf::Float64`: Rate feedback excitation system stabilizer gain, validation range: (0, nothing)
+- `Tf::Float64`: Rate feedback time constant, validation range: (&quot;eps()&quot;, nothing)
```
rodrigomha, line 42: @jd-lara @daniel-thom 
Any idea on how to fix this comment to not output `&quot`. In this case the validation range has to be strictly greater than zero (i.e. `eps()`).

src/descriptors/power_system_structs.json
```diff
@@ -2978,9 +2978,571 @@
...
+        {
+          "name": "Vr_lim",
+          "comment": "Limits for regulator output (Vr_min, Vr_max)",
+          "null_value": "(0.0, 0.0)",
+          "data_type": "Tuple{Float64, Float64}"
+        },
+        {
+          "name": "Te",
+          "comment": "Exciter field time constant in s",
+          "null_value": "eps()",
```
rodrigomha, line 3227: @jd-lara @daniel-thom 
Should I go with `null_value = 0` even though the validation range is `(eps(), infinity)`?. Not sure here what is the exactly rule to set-up `null_value`.

rodrigomha:
@daniel-thom Hey Dan, I'm having some issues with codecov, since any time I add a new struct the `set_value!` functions are not being used.
Any ideas here on how to test those functions automatically?

daniel-thom:
> @daniel-thom Hey Dan, I'm having some issues with codecov, since any time I add a new struct the `set_value!` functions are not being used.
> Any ideas here on how to test those functions automatically?

@rodrigomha Yes, I've been wanting to do this.  We could automatically construct an instance of every auto-generated struct that has a `demo` constructor, call each getter function, and call each setter function with the listed `null` value.  This would be a small amount of code in IS.

src/descriptors/power_system_structs.json
daniel-thom: This is for the demo constructor like
```
tThermalGen = ThermalStandard(nothing)
```
An invalid value of 0 is appropriate.

src/models/generated/AC1A.jl
daniel-thom: This is a problem in the IS Mustache template definition.  The `valid_range` and `validation_action` variables need to be enclosed in three curly braces instead of two.

Current:
```
- `{{name}}::{{{data_type}}}`{{#comment}}: {{{comment}}}{{/comment}}{{#valid_range}}, validation range: {{valid_range}}{{/valid_range}}{{#validation_action}}, action if invalid: {{validation_action}}{{/validation_action}}
```
Fix:
```- `{{name}}::{{{data_type}}}`{{#comment}}: {{{comment}}}{{/comment}}{{#valid_range}}, validation range: {{{valid_range}}}{{/valid_range}}{{#validation_action}}, action if invalid: {{{validation_action}}}{{/validation_action}}
```

Please test this locally and then we can fix push a fix in IS.

rodrigomha: This work locally and create a documentation of the form:
`- `R::Float64`: Speed droop parameter, validation range: ("eps()", nothing)`

I will not do the change now since it is failing the `test_generate_structs.jl` that it seems to compare the docstring, and will probably fail the test here in Travis.

daniel-thom: Unfortunately, we will have to make the change in IS, create a new tag, update to that tag in PSY, regenerate the structs, commit that change..

test/test_powersystemconstructors.jl
```diff
@@ -122,17 +122,43 @@ end
...
+                _func!(obj, aux)
+                @test func(obj) == aux
+            elseif typeof(val) == Bool
+                aux = !val
+                _func!(obj, aux)
+                @test func(obj) == aux
+            end
         end
     end
 end
```
rodrigomha, line 164: @jd-lara @daniel-thom What do you think of this attempt for testing `set_value!` functions?

jd-lara: LGTM, I will wait until Dan takes a look as well

test/test_powersystemconstructors.jl
```diff
@@ -122,17 +122,43 @@ end
     # If that isn't appropriate for this type, add it to types_to_skip below.
 
     types_to_skip = (System,)
+    types_to_skip_set = (ThermalMultiStart,)
+    func_to_skip_set = (PowerSystems.set_dynamic_injector!,)
```
daniel-thom, line 126: Just curious, why is this not allowed for ThermalMultiStart but is allowed for other generators.  I'm assuming that the value for this field from the `demo` constructor will always be `nothing`.

test/test_powersystemconstructors.jl
```diff
@@ -122,17 +122,43 @@ end
...
+                _func! = getfield(
+                    InfrastructureSystems,
+                    Symbol("set_" * string(field_name) * "!"),
+                )
             else
                 func = getfield(PowerSystems, Symbol("get_" * string(field_name)))
+                _func! = getfield(PowerSystems, Symbol("set_" * string(field_name) * "!"))
             end
             val = func(obj)
+            if !((ps_type in types_to_skip_set) & (_func! in func_to_skip_set))
```
daniel-thom, line 142: Recommend a logical `and` (`&&`) instead of binary.

test/test_powersystemconstructors.jl
```diff
@@ -122,17 +122,43 @@ end
...
+                _func!(obj, val)
+            end
             @test val isa field_type
+            #Test set function for different cases
+            if typeof(val) == Float64 || typeof(val) == Int64
```
daniel-thom, line 147: Thanks for adding this.

test/test_powersystemconstructors.jl
rodrigomha: Not clear for me too. The error is basically that set_dynamic_injector in ThermalMultiStart is not implemented (but it is defined). The value is indeed nothing from the getter function. Here is the stack trace:
```
set_dynamic_injector! is not implemented for ThermalMultiStart
  Stacktrace:
   [1] error(::String) at ./error.jl:33
   [2] set_dynamic_injector!(::ThermalMultiStart, ::Nothing) at /Users/rhenriquez/.julia/dev/PowerSystems/src/models/injection.jl:12
```

jd-lara: This is an issue with the fact that multistart doesn't have the proper field and should be added. But that should be done in a separate PR.

daniel-thom: The problem is that ThermalMultStart is not listed in the set_dynamic_injector! function in src/models/supplemental_accessors.jl.  You could add it to the Union in that function.  Maybe we should consider using ThermalGen and RenewableGen instead of concrete types.  I used concretes because I wasn't sure if every instance would have a dynamic generator.  I'll defer to you guys on this.

Your test found a bug!

jd-lara: Agreed that should be based on the Abstract

daniel-thom:
Why not fix the issue before merging?

PR #517, opened 2020-06-17
update the addition of services 

jd-lara:
This PR allows the user to create a custom device without requiring the implementation of services.

### DIFF
--- a/src/models/static_models.jl
+++ b/src/models/static_models.jl
@@ -1,6 +1,10 @@
 abstract type StaticInjection <: Device end
 function supports_services(::T) where {T <: Device}
-    return hasfield(T, :services)
+    return true
+end
+
+function get_services(::Device)
+    return Vector{Service}()
 end
 
 get_dynamic_injector(d::StaticInjection) = nothing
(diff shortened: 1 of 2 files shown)


src/models/static_models.jl
```diff
@@ -1,3 +1,6 @@
 abstract type StaticInjection <: Device end
-supports_services(d::Device) = true
+function supports_services(::T) where {T <: Device}
```
daniel-thom, line 2: I advise against this.  It imposes a field name requirement, which we've avoided doing.  The previous code mandated that every subtype of device implement a `get_services` method which means the field names could be anything.

Your solution is the simplest and perhaps the most pragmatic in the short term.  Here are some alternatives that may be better in the long run:
1. Leave the old `supports_services` method in place.  If any new device does not have services then it must implement a version that returns false.
2. Change the old `supports_services` method to call `error("not implemented")`.  Require that every subtype of Device be covered by an appropriate method.  This may be straightforward if there is a clean breakdown by abstract type.
3. Use your solution as a catch-all but implement `support_services` for types that we know support them and document that others should do the same.  That may not work out well in practice.

PR #519, opened 2020-06-18
Add set dynamic injector for MultiThermalStart

rodrigomha:
Forgot to fix this in the previous PR. 

This PR also clean the tests for `set_dynamic_injector!`.


daniel-thom:
@rodrigomha @jd-lara I think the proper fix is to change the common set_dynamic_injector function in the manner I described in the previous PR.  It's important that we reach consensus here.  

I chose to make a common function instead of relying on auto-generated functions in order to provide protections.  This function gets called automatically in `add_component!` and `remove_component`.  We want to make sure that PSY calls the function and not the user.

This strategy doesn't seem to be going well.  Should we reconsider?

https://github.com/NREL-SIIP/PowerSystems.jl/blob/master/src/models/supplemental_accessors.jl#L14

PR #521, opened 2020-06-22
Add Jacobian to power flow solver

rodrigomha:
The following PR add the Jacobian function to the `solve_powerflow!` function.

I modified the function to include now both `nlsolve` and `OnceDifferentiable` and an extra boolean for testing. Sadly, the sparse Jacobian implementation is slower than the finite differences one. For the 2000 bus case, it takes around 20 seconds to run using finite diff, while 250 seconds with Jacobian in my computer. I'm still not sure what is causing the problem, but at least is converging (that is a huge improvement). Any ideas here?

I also modified the **signs** of the variables of P and Q from the slack bus and Q from the PV bus. For some reason, doing this change helped to accelerate the solution of finite diff. **Without this change, the Jacobian implementation was not converging.**

For testing purposes, I removed the `write_pf_sol!` for now in the loop. So you can re-run without worrying on system updates (I haven't updated the tests yet).

You can use the following code to test the results:
```
#checkout to this branch
using PowerSystems
using NLsolve
case_dyr = "ACTIVSg2000.m" #use your location of the file
sys = System(case_dyr)
#Test Finite Diff
@time solve_powerflow!(sys, nlsolve, OnceDifferentiable, true)
#Test Jacobian implementation
@time solve_powerflow!(sys, nlsolve, OnceDifferentiable, false)
```


rodrigomha:
0be836e fixes the computational slowdowns. We are now faster than PM (sort of).

There is some room to improvements on the sparsity pattern of the Jacobian to do and some refactoring on the function itself.

src/utils/power_flow/power_flow.jl
```diff
@@ -255,27 +487,177 @@ function solve_powerflow!(system::System, nlsolve; args...)
...
+                b = bus_types[ix_t]
+
+                if b == BusTypes.REF
+                    # State variables are Active and Reactive Power Generated
+                    # F[2*i-1] := p[i] = p_flow[i] + p_load[i] - x[2*i-1]
+                    # F[2*i] := q[i] = q_flow[i] + q_load[i] - x[2*i]
+                    # x does not appears in p_flow and q_flow
+                    if ix_f == ix_t
+                        J[F_ix_f_r, X_ix_t_fst] = -1.0
+                        J[F_ix_f_r, X_ix_t_snd] = 0.0
```
jd-lara, line 526: If these are always cero they don't need an update.

src/utils/power_flow/power_flow.jl
```diff
@@ -255,27 +487,177 @@ function solve_powerflow!(system::System, nlsolve; args...)
...
+                        J[F_ix_f_i, X_ix_t_snd] = -1.0
+                    end
+                elseif b == BusTypes.PV
+                    # State variables are Reactive Power Generated and Voltage Angle
+                    # F[2*i-1] := p[i] = p_flow[i] + p_load[i] - p_gen[i]
+                    # F[2*i] := q[i] = q_flow[i] + q_load[i] - x[2*i]
+                    # x[2*i] (associated with q_gen) does not appear in q_flow
+                    # x[2*i] (associated with q_gen) does not appear in the active power balance
+                    if ix_f == ix_t
+                        J[F_ix_f_r, X_ix_t_fst] = 0.0
```
jd-lara, line 537: same here

src/utils/power_flow/power_flow.jl
```diff
@@ -255,27 +487,177 @@ function solve_powerflow!(system::System, nlsolve; args...)
...
+                        J[F_ix_f_i, X_ix_t_snd] =
+                            Vm[ix_f] * sum(
+                                Vm[k] * (
+                                    real(Yb[ix_f, k]) * cos(θ[ix_f] - θ[k]) -
+                                    imag(Yb[ix_f, k]) * -sin(θ[ix_f] - θ[k])
+                                ) for k in neighbors[ix_f] if k != ix_f
+                            )
+                    else
+                        #q_gen[ix_t] does not appear in other equations
+                        J[F_ix_f_r, X_ix_t_fst] = 0.0
```
jd-lara, line 558: same here

src/utils/power_flow/power_flow.jl
```diff
@@ -172,6 +371,38 @@ function solve_powerflow!(system::System, nlsolve; args...)
...
+    #Create Jacobian structure
+    J0_I = Int64[]
+    J0_J = Int64[]
+    J0_V = Float64[]
+
+    for ix in a
+        F_ix_r = 2 * ix - 1
+        F_ix_i = 2 * ix
+
+        for j in neighbors[ix]
```
jd-lara, line 383: We have the capability of taking into account BusTypes here to fill properly and avoid filling unnecessary "holes" of the Jacobian

PR #523, opened 2020-06-26
minor table data parser improvements

claytonpbarrows:


src/parsers/power_system_table_data.jl
```diff
@@ -758,35 +758,18 @@ function calculate_variable_cost(data::PowerSystemTableData, gen, cost_colnames)
         for i in 2:length(var_cost)
             var_cost[i] = (var_cost[i - 1][1] + var_cost[i][1], var_cost[i][2])
         end
+    elseif length(unique(var_cost)) == 1
```
daniel-thom, line 761: Whenever I see this code I wonder why the length of `unique(var_cost)` is checked but then it indexes into `var_cost`.  Is this always correct?  Does it merit a comment on what is happening?

claytonpbarrows: the issue is that if var_cost looks like `[(0.5, 200), (0.5, 200), (0.5, 200)]` , we're going to create a PWL curve with multiple segments, when it's principally a straight line. This check avoids the case where there is only one straight line represented by the data, but we could do a better job in removing extra points that might exist when there are multiple distinct PWL slopes. 
Historically, the existence of extra points in the PWL curve when they weren't needed caused us problems in PSI. But with recent PWL changes, I'm not sure if that is still true. @jd-lara do you know off the tp of your head?

jd-lara: Additional cost terms like that example will always cause problems because it creates the potential for degeneracy. i.e., the system can choose to operate at any intermediate point along the straight line.

claytonpbarrows: > Additional cost terms like that example will always cause problems because it creates the potential for degeneracy. i.e., the system can choose to operate at any intermediate point along the straight line.

I think in this case it would just create multiple PWL variables and constraints where only one is sufficient.

daniel-thom: Could you convert `var_cost` to `unique(var_cost)` at the beginning of the function?

claytonpbarrows: yes, as long as that preserves order?

daniel-thom: Yes.

```
?unique
Return an array containing only the unique elements of collection itr, as determined by isequal, in the order
  that the first of each set of equivalent elements originally appears. The element type of the input is
  preserved.
```

PR #524, opened 2020-06-26
Cb/ptdf monitored line

claytonpbarrows:
this PR resolves an issue where we were making PTDF calculations based on branch types. The calculations failed when a `MonitoredLine` was present since the calculation only checked for `Line`. I've added an abstract type for `ACLine` and `ACTransformer` for these types of checks.

### DIFF
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -23,7 +23,6 @@ export Branch
 export StaticInjection
 export ACBranch
 export ACLine
-export ACTransformer
 export Line
 export MonitoredLine
 export DCBranch
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -22,6 +22,8 @@ export Device
 export Branch
 export StaticInjection
 export ACBranch
+export ACLine
+export ACTransformer
 export Line
 export MonitoredLine
 export DCBranch
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -22,7 +22,6 @@ export Device
 export Branch
 export StaticInjection
 export ACBranch
-export ACLine
 export Line
 export MonitoredLine
 export DCBranch
--- a/src/models/supplemental_accessors.jl
+++ b/src/models/supplemental_accessors.jl
@@ -14,7 +14,7 @@ _remove_aggregration_topology!(bus::Bus, ::Area) = bus.area = nothing
 """
     calculates the admittance of AC branches
 """
-get_admittance(b::T) where {T <: ACBranch} = 1 / get_x(b)
+get_admittance(b::ACBranch) =  1 / get_x(b)
 get_admittance(b::TapTransformer) = 1 / (get_x(b) * get_tap(b))
 function get_admittance(b::PhaseShiftingTransformer)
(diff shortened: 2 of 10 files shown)


src/PowerSystems.jl
```diff
@@ -22,6 +22,8 @@ export Device
 export Branch
 export StaticInjection
 export ACBranch
+export ACLine
+export ACTransformer
```
jd-lara, line 26: Is there such thing as a DC Transformer? .... ACBranch covers both cases already.

claytonpbarrows: but if you want all transformers... what should the abstract type be?
 - `Transformer`?

src/utils/network_calculations/ptdf_calculations.jl
```diff
@@ -37,7 +37,7 @@ function _buildptdf(branches, nodes, dist_slack::Array{Float64} = [0.1])
         elseif isa(b, TapTransformer)
             inv_X[ix, ix] = 1 / (get_x(b) * get_tap(b))
 
-        elseif isa(b, Line)
+        elseif isa(b, ACLine)
```
jd-lara, line 40: A cleaner solution is to implement `get_admittance` 

```Julia
get_admittance(dev::ACBranch) =  1 / get_x(b)
get_admittance(dev::TapTransformer) =  1 / (get_x(b) * get_tap(b))
function get_admittance(dev::PhaseShiftingTransformer) 
     y = 1 / (get_r(b) + get_x(b) * 1im)		         
     y_a = y / (get_tap(b) * exp(get_α(b) * 1im * (π / 180)))
    return 1 / imag(y_a)
  end
```

src/PowerSystems.jl
jd-lara: That can be implemented but I don't see the benefit from the modeling perspective al transformers that aren't Tap or PhaseShifters are the same as an ACBranch

src/models/supplemental_accessors.jl
```diff
@@ -10,3 +10,14 @@ set_area!(bus::Bus, area::Area) = bus.area = area
 """
 _remove_aggregration_topology!(bus::Bus, ::LoadZone) = bus.load_zone = nothing
 _remove_aggregration_topology!(bus::Bus, ::Area) = bus.area = nothing
+
+"""
+    calculates the admittance of AC branches
+"""
+get_admittance(b::T) where {T <: ACBranch} =  1 / get_x(b)
```
jd-lara:
```suggestion
get_admittance(b::ACBranch) =  1 / get_x(b)
```

src/PowerSystems.jl
```diff
@@ -22,6 +22,7 @@ export Device
 export Branch
 export StaticInjection
 export ACBranch
+export ACLine
```
jd-lara, line 25: I still don't see why what's the value-added of ACLine is there is Line. I feel this is a duplicate.

claytonpbarrows: Line is a concrete type. We need something that collects both `MonitoredLine` and `Line`.

claytonpbarrows: I suppose now that we're just using the accessor functions to calculate admittance, we could get rid of it.

PR #527, opened 2020-07-05
more table data improvements

claytonpbarrows:


src/parsers/power_system_table_data.jl
```diff
@@ -534,6 +537,40 @@ function load_csv_parser!(sys::System, data::PowerSystemTableData)
...
+    load_csv_parser!(sys::System, data::PowerSystemTableData)
+
+Add loads to the System from the raw load data.
+
+"""
+function load_csv_parser!(sys::System, data::PowerSystemTableData)
+    for rawload in iterate_rows(data, LOAD::InputCategory)
+        bus = get_bus(sys, rawload.bus_id)
+        if isnothing(bus)
+            throw(DataFormatError("could not find $(rawload.bus_id)"))
```
daniel-thom, line 549:
```suggestion
            throw(DataFormatError("could not find bus_number=$(rawload.bus_id) for load=$(rawload.name)"))
```

src/base.jl
```diff
@@ -460,7 +460,17 @@ function IS.add_forecast!(
...
+            key = IS.ForecastKey(
+                typeof(forecast),
+                get_initial_time(forecast),
+                IS.get_label(forecast),
+            )
+            if haskey(forecasts.data, key)
+                @warn "forecast $key is already stored for $(get_name(component_)), not adding $(get_name(component)) Area forecast"
+            else
+                IS.add_forecast!(data, component_, forecast, ts_data)
+            end
```
daniel-thom, line 473: I'm OK with this, but it seems a bit cumbersome, particularly if someone needs to do the same thing elsewhere.  As an alternative, we could add an optional flag to `IS.add_forecast` called `skip_if_present` to avoid the exception.  What do you think?

src/parsers/power_system_table_data.jl
```diff
@@ -492,10 +493,12 @@ function gen_csv_parser!(sys::System, data::PowerSystemTableData)
...
 
 """
-Add loads to the System from the raw data.
+    busload_csv_parser!(sys::System, data::PowerSystemTableData)
+
+Add loads to the System from the raw bus data.
 
 """
-function load_csv_parser!(sys::System, data::PowerSystemTableData)
+function busload_csv_parser!(sys::System, data::PowerSystemTableData)
```
daniel-thom, line 501: I'm not sure why I haven't noticed this sooner...I do think this function has changed a lot over time.  Can't we create the PowerLoad directly in `bus_csv_parser!`?  We are running through redundant loops.

src/parsers/power_system_table_data.jl
```diff
@@ -266,13 +266,14 @@ function System(
 
     loadzone_csv_parser!(sys, data)
     bus_csv_parser!(sys, data)
-    load_csv_parser!(sys, data)
+    busload_csv_parser!(sys, data)
```
daniel-thom, line 269: It might just be me, but I don't find it obvious as to why you create PowerLoads implicitly from buses and explicitly from raw data.  Is any of this redundant?  Does it merit a comment here?

claytonpbarrows: I don't know what the right approach here is. In this instance, I've added the explicit load calculation because I want to do both. I have a dataset that has load's defined as implicit loads at each bus, and a seperate loads.csv defining auxillary loads. This may be an edge case that we don't want to support directly. I think the common cases that we should support are:
 1. bus.csv contains the implicit load
 2. bus.csv doesn't contain implicit loads, all loads are contained in load.csv

daniel-thom: How about this?

You could look for loads in load.csv.  If you find some, don't create them implicitly from buses.  Allow an override flag to always create implicit loads.

src/base.jl
claytonpbarrows: addressed here: https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/103

claytonpbarrows:
https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/104 addresses the issue about conflicting forecasts. This PR now depends on IS#master.

claytonpbarrows:
Closing in favor of #534

PR #529, opened 2020-07-08
Add Impedance source in ext when parsing RAW files

rodrigomha:
The following PR adds in the ext of the parsed generators the value of R_source and X_source. These values are necessary to run dynamic simulations.


src/parsers/power_models_data.jl
```diff
@@ -362,6 +362,12 @@ function make_thermal_gen(gen_name::AbstractString, d::Dict, bus::Bus, sys_mbase
     op_cost =
         ThreePartCost(; variable = cost, fixed = fixed, startup = startup, shutdn = shutdn)
 
+    ext = Dict{String, Any}()
+    if haskey(d, "r_source")
```
jd-lara, line 366: Collect these into a single entry.

PR #531, opened 2020-07-10
Cb/per unit tabledata

claytonpbarrows:
- requires https://github.com/NREL-SIIP/PowerSystemsTestData/pull/29


src/parsers/power_system_table_data.jl
```diff
@@ -3,7 +3,7 @@ const POWER_SYSTEM_DESCRIPTOR_FILE =
     joinpath(dirname(pathof(PowerSystems)), "descriptors", "power_system_inputs.json")
 
 struct PowerSystemTableData
-    base_power::Float64
+    basepower::Float64
```
daniel-thom, line 6: Is changing `base_power` to `basepower` a merge problem?  Jose just changed all of these?

claytonpbarrows: it's a merge problem.

PR #533, opened 2020-07-11
Jd/add dynamic branch

jd-lara:


src/base.jl
```diff
@@ -1329,6 +1329,10 @@ function handle_component_addition!(sys::System, component::RegulationDevice)
     remove_component!(sys, component.device)
 end
 
+function handle_component_addition!(sys::System, component::DynamicBranch)
```
daniel-thom, line 1332: Do you also want to add a `check_component_addition` method that checks if the `Branch` is part of the system?  As it stands if it is not part of the system then an exception will be thrown in `remove_component`, which will seem confusing to users.

PR #534, opened 2020-07-11
Cb/tabledata improvements

claytonpbarrows:
this PR makes some incremental improvments to tabledata parsing and should be included in lieu of #527


src/descriptors/power_system_inputs.json
```diff
@@ -239,7 +239,8 @@
         {
             "name": "tap",
             "unit": "%",
-            "description": "Transformer winding ratio"
+            "description": "Transformer winding ratio",
+            "default_value": 0.0
```
jd-lara, line 243: Default 1.0

PR #538, opened 2020-07-14
Fix powerflow to update in device base

rodrigomha:
The following PR fix the power flow solver to always work in system per unit, and then revert it back to the original one. It also fixes the issue with updating the solution in `system_base`, instead of `device_base`. I added a `@info` for the results exported in a DataFrame, to be clear that those results are exported in `system_base`.

### DIFF
--- a/src/utils/power_flow/power_flow.jl
+++ b/src/utils/power_flow/power_flow.jl
@@ -174,7 +174,7 @@ end
 Return power flow results in dictionary of dataframes.
 """
 function _write_results(sys::System, nl_result)
-    @info "Results are exported in system base"
+    @info("Results are exported in system base")
     result = round.(nl_result.zero; digits = 7)
     buses = sort(collect(get_components(Bus, sys)), by = x -> get_number(x))
     N_BUS = length(buses)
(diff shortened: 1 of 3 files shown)


src/utils/power_flow/power_flow.jl
```diff
@@ -171,6 +174,7 @@ end
 Return power flow results in dictionary of dataframes.
 """
 function _write_results(sys::System, nl_result)
+    @info("Results are exported in system base")
```
jd-lara, line 177: Should we export results in MW?

src/utils/power_flow/power_flow.jl
```diff
@@ -292,14 +296,20 @@ solve_powerflow!(sys, finite_diff = true)
 
 """
 function solve_powerflow!(system::System; finite_diff = false, kwargs...)
-    #finite_diff = get(kwargs, :finite_diff, false)
+    #Save per-unit flag
+    flag = deepcopy(system.units_settings.unit_system)
```
jd-lara, line 300: flag doesn't really depict what we had. We should leave it as settings_cache.

PR #540, opened 2020-07-15
add set base system

jd-lara:
Helper function to change system unit base

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -353,7 +353,7 @@ function set_units_base_system!(system::System, settings::String)
     return
 end
 
-function set_units_base_system!(system::System, settings::SystemUnitsSettings)
+function set_units_base_system!(system::System, settings::IS.SystemUnitsSettings)
     system.units_settings.unit_system = settings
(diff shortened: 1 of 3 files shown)


src/base.jl
```diff
@@ -344,6 +344,20 @@ function set_unit_system!(component::Component, settings::SystemUnitsSettings)
     return
 end
 
+"""
+Sets the units base for the getter functions on the devices. It modifies the behavior of all getter functions
+"""
+function set_units_base_system!(system::System, settings::String)
+    system.units_settings.unit_system = UNIT_SYSTEM_MAPPING[settings]
```
daniel-thom, line 351: Should we log this?  and the one below?

jd-lara: Should I add an @info statement or @debug?

daniel-thom: I'm imagining someone filing an issue about getting incorrect units and us wanting to know whether they set this.  Having it in the log would solve that problem.  info unless it will be called a lot, which should not be the case.

PR #542, opened 2020-07-16
sort buses un Ybus constructor

### LINKED ISSUE #447: Check binfo return value from getrf!(B)
It could be an idea to check the binfo return value of getrf!(B). If negative, it means an illegal input value. If positive, it means there are zeros that cause the matrix to be singular indicating islanding or some other problem with the network.

It could be nice to check for this and provide some additional feedback to the user.

I notice there is a #TODO in the code to check for islands, this could also be useful here.

From http://www.netlib.org/lapack/explore-html/dd/d9a/group__double_g_ecomputational_ga0019443faea08275ca60a734d0593e60.html

```
INFO is INTEGER
          = 0:  successful exit
          < 0:  if INFO = -i, the i-th argument had an illegal value
          > 0:  if INFO = i, U(i,i) is exactly zero. The factorization
                has been completed, but the factor U is exactly
                singular, and division by zero will occur if it is used
                to solve a system of equations.
```

jd-lara:
closes #447


src/utils/network_calculations/ptdf_calculations.jl
```diff
@@ -5,6 +5,19 @@ struct PTDF{Ax, L <: NTuple{2, Dict}} <: PowerNetworkMatrix{Float64}
     lookup::L
 end
 
+function _binfo_check(binfo)
+    if binfo != 0
+        if binfo < 0
+            error("Illegal Agument in Inputs")
```
claytonpbarrows, line 11: typo: "Agument"

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -141,39 +121,75 @@ function _buildybus(branches, nodes, fixed_admittances)
...
-
     for fa in fixed_admittances
         _ybus!(ybus, fa, num_bus)
     end
+    return SparseArrays.dropzeros!(ybus)
+end
 
-    return ybus
-
+function _goderya(ybus::SparseArrays.SparseMatrixCSC{ComplexF64, Int64})
```
rodrigomha, line 130: Any reference here? I'm guessing a Docstring here providing a reference will be useful. I'm not familiar with his method.

jd-lara: No need, not meant to be exported.

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -141,39 +121,75 @@ function _buildybus(branches, nodes, fixed_admittances)
...
+    for n in 1:(node_count - 1)
+        I, _, _ = SparseArrays.findnz(T_)
+        if length(I) == max_I
+            @info "The System has no islands"
+            break
+        elseif length(I) < max_I
+            temp = T_ * T
+            I_temp, _, _ = SparseArrays.findnz(temp)
+            if all(I_temp == I)
+                error("The system contains islands")
```
rodrigomha, line 145: Why do we want to give errors if the system contains islands? It is possible to solve Power Flows in islanded systems right? It is definitely a problem for Dynamic Simulations though, but we can use that in some way to stop the simulation. Any issues in PSI?

jd-lara: If the system has islands the Jacobian is singular.

rodrigomha: I see. Better not support islanded systems for now.

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -141,39 +121,75 @@ function _buildybus(branches, nodes, fixed_admittances)
...
+function _goderya(ybus::SparseArrays.SparseMatrixCSC{ComplexF64, Int64})
+    node_count = size(ybus)[1]
+    max_I = node_count^2
+    I, J, val = SparseArrays.findnz(ybus)
+    T = SparseArrays.sparse(I, J, ones(length(val)))
+    T_ = T * T
+    for n in 1:(node_count - 1)
+        I, _, _ = SparseArrays.findnz(T_)
+        if length(I) == max_I
+            @info "The System has no islands"
```
claytonpbarrows, line 139: This just checks for disconnected leaf nodes, right? If that's the case, that's a good check, but the error message should be "The System has no disconnected buses". If I'm wrong, you can disregard.

jd-lara: Goderya's algorithm searches for islands generally. See the tests.

PR #543, opened 2020-07-16
Fix serialization dynamic branch

daniel-thom:
Depends on https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/107

### DIFF
--- a/src/models/dynamic_branch.jl
+++ b/src/models/dynamic_branch.jl
@@ -3,14 +3,18 @@ mutable struct DynamicBranch <: ACBranch
     n_states::Int64
     states::Vector{Symbol}
     internal::IS.InfrastructureSystemsInternal
-    function DynamicBranch(branch::T) where {T <: ACBranch}
+
+    function DynamicBranch(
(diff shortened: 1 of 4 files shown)


src/models/dynamic_branch.jl
```diff
@@ -3,14 +3,18 @@ mutable struct DynamicBranch <: ACBranch
     n_states::Int64
     states::Vector{Symbol}
     internal::IS.InfrastructureSystemsInternal
-    function DynamicBranch(branch::T) where {T <: ACBranch}
+
+    function DynamicBranch(
+        branch::T;
+        internal = IS.InfrastructureSystemsInternal(),
```
daniel-thom, line 9: @jd-lara  In order for deserialization to work we have to be able to create the object with an existing InfrastructureSystemsInternal.  If you create an inner constructor then you must allow for that.

jd-lara: Sorry, forgot about this. This needs documentation

PR #544, opened 2020-07-16
add max getter functions

jd-lara:


src/models/components.jl
```diff
@@ -36,3 +36,7 @@ function get_value(c::T, value::V) where {T <: Component, V}
     @warn("conversion not implemented for $(V) in component $(T)")
     return value::V
 end
+
+function has_forecasts(d::Component)
```
daniel-thom, line 40: This already exists in src/models/devices.jl

jd-lara: Is the same function I just moved to the file for components.

src/models/supplemental_accessors.jl
```diff
@@ -23,3 +23,15 @@ function get_series_susceptance(b::PhaseShiftingTransformer)
...
+
+function get_max_active_power(d::Device)
+    return get_active_power_limits(d).max
+end
+
+function get_max_reactive_power(d::Device)
+    return get_reactive_power_limits(d).max
+end
+
+function get_max_active_power(d::RenewableGen)
```
daniel-thom, line 35: Consider putting this in the same file where RenewableGen is defined.  Consider also implementing default methods for get_rating and get_power_factor that call `error("not implemented")`.  Or add a comment next to RenewableGen that says those methods are required.

jd-lara: I've made the updates in c5c1d2f and 8139779

src/models/generation.jl
```diff
@@ -2,6 +2,11 @@ abstract type Generator <: StaticInjection end
 const Generators = Array{<:Generator, 1}
 
 abstract type HydroGen <: Generator end
+
+"""
+Abstract Type to represent renewable generation devices. Requires the implementation of get_rating
+    and get_power_factor methods
```
daniel-thom, line 8: Is there a Julia rule that says to indent here?

src/models/generation.jl
```diff
@@ -13,3 +18,13 @@ function IS.get_limits(
     # e.g. "valid_range": {"min":-1.571, "max":1.571}.
     return (min = valid_range.min, max = valid_range.max, zero = 0.0)
 end
+
+"""
+Return the max active power for the Renewable Generation calutades as the rating * powe_factor
```
daniel-thom, line 23: Two typos:  calutades, powe_factor

src/models/generation.jl
```diff
@@ -13,3 +18,13 @@ function IS.get_limits(
...
+"""
+function get_max_active_power(d::T) where T <: RenewableGen
+    if !hasmethod(get_rating, T) || !hasmethod(get_power_factor, T)
```
daniel-thom, line 26: I think you're better off letting Julia fail on its own (certainly more performant).

If you disagree, I think your throw statement is wrong.  I think you meant `get_max_active_power` instead of `get_max_reactive_power`.  Second, are you sure that syntax is correct?

jd-lara: I think there shouldn't be too much of a performance hit to run these checks. I will correct the typo

daniel-thom: I think you have syntax errors.  Do the tests pass?

jd-lara: 3dd11b4

src/models/supplemental_accessors.jl
```diff
@@ -23,3 +23,17 @@ function get_series_susceptance(b::PhaseShiftingTransformer)
 end
 
 get_series_admittance(b::ACBranch) = 1 / (get_r(b) + get_x(b) * 1im)
+
+function get_max_active_power(d::T) where {T <: Device}
+    if !hasmethod(get_active_power_limits, Tuple{T})
+        throw(MethodError(get_max_active_power, d))
```
daniel-thom, line 29: Now you've compelled me to test it.  It's not the `MethodError` you're looking for.

```
ERROR: MethodError: no method matching iterate(::TestRenDevice)
Closest candidates are:
  iterate(::Core.SimpleVector) at essentials.jl:603
  iterate(::Core.SimpleVector, ::Any) at essentials.jl:603
  iterate(::ExponentialBackOff) at error.jl:253
```

I think you want something like `throw(MethodError(get_max_active_power, tuple(d)))`.  This still might be confusing to users because it doesn't tell them what they are missing.

If you really want these checks, which do have a minor performance cost, I suggest something like `throw(ArgumentError("Missing required methods for running get_max_active_power on $T"))`.  Or some other exception that accepts a string.

daniel-thom: Keep in mind that if you remove the checks, Julia will tell the user exactly what they are missing.

jd-lara: I changed the implementation.

PR #547, opened 2020-07-20
[WIP] Add Supplemental Reserves

andrewrosemberg:
Adds `SupplementalStaticReserve`. It will allow the user to add a reserve in PowerSimulations.jl that can be contributed both by committed and non-committed devices.


src/models/reserves.jl
```diff
@@ -1,6 +1,8 @@
 abstract type ReserveDirection end
 abstract type ReserveUp <: ReserveDirection end
 abstract type ReserveDown <: ReserveDirection end
+abstract type ReserveSupplementaryUp  <: ReserveDirection end
+abstract type ReserveSupplementaryDown  <: ReserveDirection end
```
nickrobinson251, line 5: I see there's this comment below

> ```julia
> # Note that src/models/services_struct_types.jl and the testset
> # "Test struct type collections" in test/test_services.jl must be modified if any types are
> # added here.
> ```

As far as I can tell `src/models/services_struct_types.jl` no longer exists?
But the testset does: https://github.com/NREL-SIIP/PowerSystems.jl/blob/1b5beef78adbe7ec4285cddc3809696071c4177f/test/test_services.jl#L229-L242

Please can you update that test?

Also it would be good if we can update the comment to be correct about what changes to `src/` are required (and ideally move it above the code, as it's very easy to miss right now)

daniel-thom: This is a stale test and comment and I think both can be deleted.  The test is no longer doing anything useful.  I can take of this separately or if you want to do it here, that's OK with me.  Let me know.

andrewrosemberg: Thanks for the answer @daniel-thom.
As soon as we have more feedback on the [PSI PR](https://github.com/NREL-SIIP/PowerSimulations.jl/pull/484) about how is the best way to model our new reserve, I can make this change here.

daniel-thom: Thanks for making this cleaner.  We haven't always been great about fixing these sorts of things after revising implementations.

jd-lara:
@andrewrosemberg can you please fix the tests to make the final review. The PR looks generally ok and could be included in the following release.

src/models/generated/SupplementalStaticReserve.jl
```diff
@@ -0,0 +1,80 @@
+#=
+This file is not auto-generated. Do edit if needed.
+=#
+"""
+    mutable struct SupplementalStaticReserve{T <: ReserveDirection} <: Reserve{T}
```
jd-lara, line 5: Do we still need to add this additional reserve object? It seems that the expected implementation is not different from the current static reserves.

andrewrosemberg: It's exactly the same you are right, but I added this additional reserve so we can have both in the same system. Is it already possible to have multiple static reserves with different formulations?

andrewrosemberg:
Closing it in favor of a future non-spinning PR

PR #550, opened 2020-07-21
Cb/get rating bug

claytonpbarrows:


src/parsers/power_system_table_data.jl
```diff
@@ -764,8 +764,16 @@ function calculate_variable_cost(data::PowerSystemTableData, gen, cost_colnames,
             (var_cost[2][1] / (var_cost[2][2] - var_cost[1][2]) * var_cost[1][2]),
         )
         var_cost[1] = (var_cost[1][1] - fixed, var_cost[1][2])
-        for i in 2:length(var_cost)
-            var_cost[i] = (var_cost[i - 1][1] + var_cost[i][1], var_cost[i][2])
+        for i in 1:length(var_cost)
+            if i == 1
+                var_cost[i] =
+                    (var_cost[i][1], var_cost[i][2] * gen.base_mva / data.base_power)
```
daniel-thom, line 770: Would it be better to change the calculation on line 766 instead?

claytonpbarrows: ... my oversight

PR #554, opened 2020-07-22
Add StaticGroupReserve

andrewrosemberg:
Adds a `StaticGroupReserve` type related to reserve requirements that group multiple reserve types (as discussed in this [issue](https://github.com/NREL-SIIP/PowerSimulations.jl/issues/479)).

Plus this PR generalizes `ServiceContributingDevices` to allow for "contributing services".


src/base.jl
```diff
@@ -666,32 +666,35 @@ end
 
 struct ServiceContributingDevices
     service::Service
-    contributing_devices::Vector{Device}
+    contributing_devices::Vector{Component}
```
jd-lara, line 669: Why components here? This could enable weird behavior like having a bus or a cost function be part of a service.

jd-lara: I answered myself this below.

andrewrosemberg:
> Looks like the objective is to model a group of reserves. Only a few changes might be needed, probably we want to have specific methods for this object to limit StaticReserveGroups to only have other Reserves as contributing devices.

yes, exactly. The same way that `add_service!` [function](https://github.com/NREL-SIIP/PowerSystems.jl/blob/3ac11e010f25e074f2c18c55f90bdb97b23c4709/src/base.jl#L413) tests o see if the `contributing_devices` is filled with components of type `Device`, we could add a check at the constructor (or somewhere else if you prefer)

src/models/generated/GroupReserve.jl
```diff
@@ -0,0 +1,78 @@
...
+"""
+mutable struct StaticGroupReserve <: Service
+    name::String
+    available::Bool
+    "the static value of required reserves in system p.u."
+    requirement::Float64
+    ext::Dict{String, Any}
+    "power system internal reference, do not modify"
+    internal::InfrastructureSystemsInternal
+    contributing_services::Vector{<:PSY.Service}
```
andrewrosemberg, line 32: > > Looks like the objective is to model a group of reserves. Only a few changes might be needed, probably we want to have specific methods for this object to limit StaticReserveGroups to only have other Reserves as contributing devices.
> 
> yes, exactly. The same way that `add_service!` [function](https://github.com/NREL-SIIP/PowerSystems.jl/blob/3ac11e010f25e074f2c18c55f90bdb97b23c4709/src/base.jl#L413) tests o see if the `contributing_devices` is filled with components of type `Device`, we could add a check at the constructor (or somewhere else if you prefer)

Actually, this is already enforced since the constructor only accepts `Vector{<:PSY.Service}`.

src/PowerSystems.jl
```diff
@@ -150,6 +150,7 @@ export VariableReserve
 export AGC
 export ReserveDemandCurve
 export Transfer
+export StaticGroupReserve
```
claytonpbarrows, line 153: How about `StaticReserveGroup`?

src/models/generated/GroupReserve.jl
```diff
@@ -0,0 +1,84 @@
+#=
+This file was not auto-generated.
```
claytonpbarrows, line 2: This code should get auto-generated if you make appropriate changes to [power_system_structs.json](https://github.com/NREL-SIIP/PowerSystems.jl/blob/master/src/descriptors/power_system_structs.json) and execute the following line of code:

```julia
```

jd-lara: You don't need to generate this function. The methods should work automatically.

andrewrosemberg: uhmm but the proposed `get_contributing_device_mapping` calls `_get_contributing_devices` which for normal service does the following:
```
function _get_contributing_devices(sys::System, service::T) where {T <: Service}
    uuid = IS.get_uuid(service)
    devices = ServiceContributingDevices(service, Vector{Device}())
    for device in get_components(Device, sys)
        for _service in get_services(device)
            if IS.get_uuid(_service) == uuid
                push!(devices.contributing_devices, device)
                break
            end
        end
    end
    return devices
end
```
but since the contributing services are inside my struct I have to treat in differently.

jd-lara:
@andrewrosemberg if we finish this PR it can be merged for the next release

src/PowerSystems.jl
andrewrosemberg: Even Better! I will change to that

andrewrosemberg: done!

src/models/generated/GroupReserve.jl
andrewrosemberg: should I move this function to base.jl?

andrewrosemberg: done

jd-lara: Yes, in the same place the `current get_contributing_devices`

andrewrosemberg:
> @andrewrosemberg if we finish this PR it can be merged for the next release

Just finished the required changes. Shall I start to rebase? Is there a preference on how to do this (I'm not an expert on rebasing and I'm not sure If I need to pay attention to any special thing)

jd-lara:
> > @andrewrosemberg if we finish this PR it can be merged for the next release
> 
> Just finished the required changes. Shall I start to rebase? Is there a preference on how to do this (I'm not an expert on rebasing and I'm not sure If I need to pay attention to any special thing)

I have found this tutorial useful https://www.atlassian.com/git/tutorials/merging-vs-rebasing especially the part on how not to do it.

andrewrosemberg:
@jd-lara I believe I have done the rebase correctly. Please point out if I have done something wrong!

andrewrosemberg:
Currently fixing test errors

src/base.jl
```diff
@@ -667,32 +667,45 @@ end
...
-                    push!(devices.contributing_devices, device)
-                    break
-                end
-            end
-
-            key = ServiceContributingDevicesKey((typeof(service), get_name(service)))
-            services[key] = devices
-        end
+        key = ServiceContributingDevicesKey((typeof(service), get_name(service)))
+        services[key] = _get_contributing_devices(sys, service)
```
daniel-thom, line 708: It seems like this won't work if the service is a StaticReserveGroup.  The first _get_contributing_devices method returns a ServiceContributingDevices whereas the second method return a Vector.  This dictionary requires the former.

andrewrosemberg: Yes, sorry about that! there was a bug on `get_contributing_devices`.

andrewrosemberg: I will do some thinking tonight about the way I went with those functions and answer you in the morning.

andrewrosemberg: Now it should be fixed. However, since, on the new release, you guys are calling 'PSY.get_contributing_device_mapping(sys)' inside the 'construct_service' function, all this change to 'ServiceContributingDevices' is no longer needed haha I can just call 'get_contributing_services'. 
The whole thing was just to not change the 'construct_services' API.

andrewrosemberg: @daniel-thom and @jd-lara I will refactor the PR accordingly tomorrow morning to remove all the useless stuff.

src/descriptors/power_system_structs.json
```diff
@@ -2450,6 +2450,55 @@
...
+        {
+          "name": "ext",
+          "data_type": "Dict{String, Any}",
+          "null_value": "Dict{String, Any}()",
+          "default": "Dict{String, Any}()"
+        },
+        {
+          "name": "contributing_services",
+          "comment": "Services that contribute for this requirement constraint",
+          "data_type": "Vector{<:Service}",
```
andrewrosemberg, line 2489: This doesn't work for some reason! It doesn't create a `Vector{<:Service}` type in the generated file, so I had to modify it by hand (which breaks the generated struct test). Any idea on how to proceed?

daniel-thom: What does it generate?  Also, are you sure that you don't want `Vector{Service}`?

src/base.jl
```diff
@@ -437,6 +437,23 @@ function add_service!(sys::System, service::Service, contributing_devices; kwarg
...
+- `sys::System`: system
+- `service::StaticReserveGroup`: service to add
+"""
+function add_service!(sys::System, service::StaticReserveGroup; kwargs...)
+    if sys.runchecks && !validate_struct(sys, service)
+        throw(InvalidValue("Invalid value for $service"))
+    end
+
+    set_unit_system!(service, sys.units_settings)
+    # Since this isn't atomic, order is important. Add to system before adding to devices.
```
daniel-thom, line 453: I think you can delete this comment since it doesn't apply here.

src/base.jl
```diff
@@ -437,6 +437,23 @@ function add_service!(sys::System, service::Service, contributing_devices; kwarg
...
+
+# Arguments
+- `sys::System`: system
+- `service::StaticReserveGroup`: service to add
+"""
+function add_service!(sys::System, service::StaticReserveGroup; kwargs...)
+    if sys.runchecks && !validate_struct(sys, service)
+        throw(InvalidValue("Invalid value for $service"))
+    end
+
```
daniel-thom, line 451: From reading your test it looks like the user workflow is to

1. Add services to the system.
2. Add service group to the system.

That being the case, I think that we should verify that the individual services are already part of the system.  Otherwise, someone could get the system into a strange state.

Would you mind doing this?  It should be something like
```
for _service in get_contributing_services(service)
    if !is_attached(_service, sys)
        throw(ArgumentError("service $(get_name(_service)) is not part of the system"))
    end
end

That raises another question.  What would you like to happen if a user calls `remove_component!(sys, individual_service)` when that service is part of a group?  My inclination is that we should block that and force the user to remove the group before removing a service.  I'll offer to help covering these corner cases.

File: src/descriptors/power_system_structs.json (hunk 6 above)
Comment (andrewrosemberg) [reply src/descriptors/power_system_structs.json:2489]: > Does it print something like Vector{&lt;:Service})? 

Yes!

> Are you sure that you don't want Vector{Service}? 

I have tried that but, for some reason, it doesn't accept subtypes of `Service` in the `set_contributing_services` function.

**Comment by daniel-thom** [comment]
We might need to do some work on our serialization process for this.  As it stands, if you serialize a system to JSON (which will happen in PowerSimulations if you ask it to), and then deserialize it, I expect that the service group will have individual services that are equal in content to the real ones but won't be references to the real ones.  This may or may not matter, but I think we will want them to be the same objects.  I'll do this for you tomorrow.  The process isn't exactly trivial to comprehend and replicate.

File: src/base.jl (hunk 8 above)
Comment (andrewrosemberg) [reply src/base.jl:451]: This is a great point! After the refactoring, I will add this!

> That raises another question.  What would you like to happen if a user calls `remove_component!(sys, individual_service)` when that service is part of a group?  My inclination is that we should block that and force the user to remove the group before removing a service.  I'll offer to help covering these corner cases.

Indeed, I will come back to this and talk it over with you, as this will be the more difficult part of the change.

File: src/descriptors/power_system_structs.json (hunk 6 above)
Comment (daniel-thom) [reply src/descriptors/power_system_structs.json:2489]: I proposed a fix to the auto-generation issue here:  https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/112/files.

That said, you could fix the issue you saw in your test by explicitly declaring an array of type Service. 

```
array = Vector{Service}()
push!(array, service)
```

Either way works for me.

File: src/base.jl (hunk 5 above)
Comment (andrewrosemberg) [reply src/base.jl:717]: done

File: src/descriptors/power_system_structs.json (hunk 6 above)
Comment (andrewrosemberg) [reply src/descriptors/power_system_structs.json]: My bad, that's true! I fixed it with your suggestion!

File: src/base.jl (hunk 8 above)
Comment (andrewrosemberg) [reply src/base.jl:457]: done!

File: src/base.jl [hunk 9]
```diff
@@ -557,6 +579,24 @@ function remove_component!(sys::System, component::T) where {T <: Component}
...
+            throw(ArgumentError("service $(get_name(_service)) is part of a StaticReserveGroup"))
+            return
+        end
+    end
+    check_component_removal(sys, service)
+    IS.remove_component!(sys.data, service)
+    handle_component_removal!(sys, service)
+    clear_units!(service)
+end
+
```
andrewrosemberg, line 598: > That raises another question.  What would you like to happen if a user calls `remove_component!(sys, individual_service)` when that service is part of a group?  My inclination is that we should block that and force the user to remove the group before removing a service.  I'll offer to help covering these corner cases.

@daniel-thom Is this a good solution?

andrewrosemberg:
> This looks good with a couple requested changes.

For some reason, github still thinks there are open change requests

src/base.jl
daniel-thom: You can actually make it a little simpler.  The way we set this up is that there is one main `remove_component!` method for a `Component`.  That method calls `check_component_removal` before doing the removal and `handle_component_removal` afterwards.  The default methods for both are no-ops.  The best way for you to solve this problem is to implement your checking logic in a new `check_component_removal` method that takes a Service.

Two minor things:
- `get_components` returns an iterable, so you can remove the call to `collect` and avoid some memory allocations.
- You may want to check if T is a StaticReserveGroup and then return early.

andrewrosemberg: Done!

andrewrosemberg: Please point it out if I did something not desired.

jd-lara:
@andrewrosemberg the PR from @daniel-thom is merged in IS. As soon as I tag these tests should pass.

andrewrosemberg:
The refactoring to follow @daniel-thom suggestions are now completed. Just waiting for the PR on serialization to finish this last part.

andrewrosemberg:
Format-check is the only left error. Do you guys have any suggestions on how to fix this?

daniel-thom:
> 
> 
> Format-check is the only left error. Do you guys have any suggestions on how to fix this?

I keep a file on my system to handle this.  You'll need to add `JuliaFormatter` to your global Julia package space.

```
cat format.jl 
using JuliaFormatter
include(".github/workflows/formatter_code.jl"
```
And then from the root of the repo:
```
julia format.jl
```

daniel-thom:
@andrewrosemberg This may not be clear.  I fixed the Mustache error you saw by removing the types from the setter functions.  Jose merged it in the last hour and then updated PowerSystems.  This means that you'll need to
1. Pull the latest InfrastructureSystems changes.
2. Pull the latest PowerSystems changes to your local master branch.
3. Rebase the master branch onto your branch.
4. Rerun the `IS.generate_structs()` function

jd-lara:
@andrewrosemberg I will make a release with bug fixes now, merge this afterward and make a minor release once @daniel-thom has included serialization. Since this is all new features will be non-breaking.

andrewrosemberg:
I don't understand the current error in tests.  Does anyone have any suggestions, I have just rebased.
```
ERROR: Package PowerSystems errored during testing
Stacktrace:
 [1] pkgerror(::String) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Pkg/src/Types.jl:52
 [2] test(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}; coverage::Bool, julia_args::Cmd, test_args::Cmd, test_fn::Nothing) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Pkg/src/Operations.jl:1566
 [3] test(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}; coverage::Bool, test_fn::Nothing, julia_args::Cmd, test_args::Cmd, kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Pkg/src/API.jl:328
 [4] #test#61 at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Pkg/src/API.jl:67 [inlined]
 [5] test(; name::Nothing, uuid::Nothing, version::Nothing, url::Nothing, rev::Nothing, path::Nothing, mode::Pkg.Types.PackageMode, subdir::Nothing, kwargs::Base.Iterators.Pairs{Symbol,Bool,Tuple{Symbol},NamedTuple{(:coverage,),Tuple{Bool}}}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Pkg/src/API.jl:80
 [6] top-level scope at none:1
```

jd-lara:
> Does anyone have any suggestions? I have just rebased. Apparently it thinks that some generated file is not generated.
> 
> ```
> Test generated structs: Test Failed at /Users/andrewrosemberg/Workspace/PowerSystems.jl/test/test_generate_structs.jl:5
>   Expression: IS.test_generated_structs(descriptor_file, existing_dir)
> Stacktrace:
>  [1] top-level scope at /Users/andrewrosemberg/Workspace/PowerSystems.jl/test/test_generate_structs.jl:5
>  [2] top-level scope at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Test/src/Test.jl:1113
>  [3] top-level scope at /Users/andrewrosemberg/Workspace/PowerSystems.jl/test/test_generate_structs.jl:2
> ```

This means that the generated files do not match the files in the generated folder. You might need to update IS and rebase again.

src/models/generated/AC1A.jl
```diff
@@ -189,46 +189,46 @@ get_states_types(value::AC1A) = value.states_types
 get_internal(value::AC1A) = value.internal
 
 """Set AC1A Tr."""
-set_Tr!(value::AC1A, val) = value.Tr = val
+set_Tr!(value::AC1A, val::Float64) = value.Tr = val
```
daniel-thom, line 192: Something went wrong with your rebase.  Commit ID b9fadd466 from Jose removed the types from these parameters.  Commit ID 2e996e0e0 from you put them back.

That doesn't seem related to your error message but there could be a common cause.  Can you determine what happened?

andrewrosemberg: @jd-lara and @daniel-thom, you are both right, sorry! my manifest was pinned on IS 0.13 for some reason. Fixing it now

andrewrosemberg: done tests passing locally now! Will wait for the ci.

daniel-thom:
@andrewrosemberg Looks like your branch needs another rebase.

andrewrosemberg:
rebase to tag 0.25 just pushed.

daniel-thom:
@andrewrosemberg I made a PR for serialization here:  https://github.com/NREL-SIIP/PowerSystems.jl/pull/571

PR #557, opened 2020-07-22
add special case for RenewableDispatch

jd-lara:


src/models/supplemental_accessors.jl
```diff
@@ -24,10 +24,26 @@ end
...
     return get_reactive_power_limits(d).max
 end
+
+"""
+Return the max reactive power for the Renewable Generation calculated as the rating * power_factor if
+reactive_power_limits is nothing
+"""
+function get_max_reactive_power(d::RenewableDispatch)
+    reactive_power_limits = get_reactive_power_limits(d)
+    isnothing(reactive_power_limits) && return get_max_reactive_power(d)
```
daniel-thom, line 47: Isn't this an infinite loop because of recursion?

PR #561, opened 2020-07-28
Dynamic data parsing

rodrigomha:
Clean PR for Dynamic Data Parsing.

We should check:

- Naming used for functions. Not sure if `parse_dyn_system` and `add_dyn_injectors!` are proper for naming.
- Test used to check dyn parse. We can probably expand later testing when adding more dyr examples.
- Overall performance.


rodrigomha:
New changes added a new function for System, that now accepts two strings, that must be a .raw file and a .dyr file.

src/parsers/psse_dynamic_data.jl
```diff
@@ -0,0 +1,267 @@
...
+    start = 1
+    parsed_values = Dict{Int, Dict}()
+    models = Vector{String}()
+    while start < length(dyr_text)
+        val = findnext('/', dyr_text, start)
+        if isnothing(val)
+            break
+        end
+        line = replace(dyr_text[start:(val - 1)], "\'" => "")
+        val_array = strip.(split(line))
```
daniel-thom, line 17: It would be good to `@assert` on the required size of this array and print the actual if it does trip.

src/parsers/psse_dynamic_data.jl
```diff
@@ -0,0 +1,267 @@
...
+        active_power = get_active_power(g),
+        reactive_power = get_reactive_power(g),
+        R_th = r,
+        X_th = x,
+    )
+end
+
+function _assign_missing_components!(bus_dict, component_table)
+    for va in values(bus_dict)
+        !isassigned(va, component_table["AVR"]) &&
```
daniel-thom, line 78: I think we've agreed in the coding standards to implement these as if statements.

src/parsers/psse_dynamic_data.jl
```diff
@@ -0,0 +1,267 @@
...
+    #dic will contain the dictionary index by bus.
+    #Each entry will be a dictionary, with id as keys, that contains the vector of components
+    dic = Dict{Int64, Any}()
+    component_table =
+        Dict("Machine" => 1, "Shaft" => 2, "AVR" => 3, "TurbineGov" => 4, "PSS" => 5)
+    for (bus_num, bus_data) in data
+        bus_dict = Dict{Int64, Any}()
+        for (componentID, componentValues) in bus_data
+            #Fill array of 5 components per generator
+            temp = get!(bus_dict, componentID[2], Vector{Any}(undef, 5))
```
daniel-thom, line 130: What happens with `temp`?  It doesn't appear to be used for anything.

src/parsers/psse_dynamic_data.jl
```diff
@@ -0,0 +1,267 @@
...
+                if x == 0.0
+                    @warn "No series reactance found. Setting it to 1e-6"
+                    x = 1e-6
+                end
+                set_Xd_p!(machine, x)
+                if get_H(shaft) == 0.0
+                    @info "Machine at bus $(_num), id $(_id) has zero inertia. Modeling it as Voltage Source"
+                    s = _make_source(g, r, x)
+                    #Remove ThermalGenerator
+                    remove_component!(typeof(g), sys, _name)
```
daniel-thom, line 255: You could call `remove_component!(sys, g)` instead.  Is this safe in Julia?  You are removing an object from a dictionary while iterating over that dictionary.  Calling `collect(get_components())` would guarantee safety.  Or build a vector of generators to remove.

src/parsers/psse_dynamic_data.jl
```diff
@@ -0,0 +1,267 @@
...
+            if typeof(machine) == BaseMachine
+                if x == 0.0
+                    @warn "No series reactance found. Setting it to 1e-6"
+                    x = 1e-6
+                end
+                set_Xd_p!(machine, x)
+                if get_H(shaft) == 0.0
+                    @info "Machine at bus $(_num), id $(_id) has zero inertia. Modeling it as Voltage Source"
+                    s = _make_source(g, r, x)
+                    #Remove ThermalGenerator
```
daniel-thom, line 254: Nit:  I recommend leaving comments like these out.  The code already speaks for itself.

src/parsers/psse_dynamic_data.jl
rodrigomha: `temp` is the value that is stored in `bus_dict` with key `componentID[2]`. Basically, modifying `temp`, modifies the vector stored in `bus_dict`. The nice thing also, is that I can create the vector if that key is not defined yet.

src/parsers/psse_dynamic_data.jl
rodrigomha: In `.dyr` files this is not possible to check. Depending on the component the size of the array is different (for example in GENCLS the length will be 5, and in GENROU will be 18). However, the first three strings are the same (bus, model and id).

PR #563, opened 2020-07-30
Add more defaults to ThermalStandard

raphaelsaavedra:
Add more default values to the `ThermalStandard` fields. I added the same values as the null ones.

Please let me know if editing the JSON is all I need to do for this change.

### DIFF
--- a/src/descriptors/power_system_structs.json
+++ b/src/descriptors/power_system_structs.json
@@ -1903,8 +1903,7 @@
         {
           "null_value": "false",
           "name": "status",
-          "data_type": "Bool",
-          "default": "false"
+          "data_type": "Bool"
         },
         {
           "name": "bus",
@@ -1976,8 +1975,7 @@
             "max": null
           },
           "validation_action": "error",
-          "needs_conversion": true,
-          "default": "nothing"
+          "needs_conversion": true
         },
         {
           "name": "time_limits",
@@ -1994,8 +1992,7 @@
         {
           "name": "operation_cost",
           "null_value": "ThreePartCost(nothing)",
-          "data_type": "ThreePartCost",
-          "default": "ThreePartCost(nothing)"
+          "data_type": "ThreePartCost"
         },
         {
           "name": "base_power",
--- a/src/descriptors/power_system_structs.json
+++ b/src/descriptors/power_system_structs.json
@@ -1903,7 +1903,8 @@
         {
           "null_value": "false",
           "name": "status",
-          "data_type": "Bool"
+          "data_type": "Bool",
+          "default": "false"
         },
         {
           "name": "bus",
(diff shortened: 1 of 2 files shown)


src/descriptors/power_system_structs.json
```diff
@@ -1983,12 +1988,14 @@
...
-          "validation_action": "error"
+          "validation_action": "error",
+          "default": "nothing"
         },
         {
           "name": "operation_cost",
           "null_value": "ThreePartCost(nothing)",
-          "data_type": "ThreePartCost"
+          "data_type": "ThreePartCost",
+          "default": "ThreePartCost(nothing)"
```
nickrobinson251, line 1998: does this introduce functionality that relies on the "non-functional" demo constructor
https://github.com/NREL-SIIP/PowerSystems.jl/blob/228ef8dd5765640dee30f7c4e9da924223ea30ab/src/models/generated/ThreePartCost.jl#L47

Would it be preferable to add defaults to `ThreePartCost`?

More generally, i wonder if all `Foo(::Nothing)` methods could be replaced by `Foo()` and appropraite defaults (which would them also be useful to the keyword constructors that override some but not all defaults)

jd-lara:
Not all the fields should have defaults, only those for which there is a reasonable value.

src/descriptors/power_system_structs.json
```diff
@@ -1936,19 +1937,22 @@
...
-          "data_type": "PrimeMovers.PrimeMover"
+          "data_type": "PrimeMovers.PrimeMover",
+          "default": "PrimeMovers.OT"
         },
         {
           "name": "fuel",
           "comment": "prime_mover Fuel according to EIA 923",
           "null_value": "ThermalFuels.OTHER",
-          "data_type": "ThermalFuels.ThermalFuel"
+          "data_type": "ThermalFuels.ThermalFuel",
```
jd-lara, line 1954: This is a reasonable default.

src/descriptors/power_system_structs.json
```diff
@@ -1936,19 +1937,22 @@
...
-          "needs_conversion": true
+          "needs_conversion": true,
+          "default": "0.0"
         },
         {
           "name": "prime_mover",
           "comment": "prime_mover Technology according to EIA 923",
           "null_value": "PrimeMovers.OT",
-          "data_type": "PrimeMovers.PrimeMover"
+          "data_type": "PrimeMovers.PrimeMover",
```
jd-lara, line 1947: This is a reasonable default

src/descriptors/power_system_structs.json
```diff
@@ -1903,7 +1903,8 @@
         {
           "null_value": "false",
           "name": "status",
-          "data_type": "Bool"
+          "data_type": "Bool",
```
jd-lara, line 1906: This should be defined properly by the user

src/descriptors/power_system_structs.json
```diff
@@ -1972,7 +1976,8 @@
             "max": null
           },
           "validation_action": "error",
-          "needs_conversion": true
+          "needs_conversion": true,
```
jd-lara, line 1979: Same here, these fields should have reasonable values

src/descriptors/power_system_structs.json
jd-lara: The same issue is here, there are reasonable estimates for this. Even if it is zero. Having a `nothing` implies that the information doesn't exist or there is no reasonable estimates.

jd-lara:
@raphaelsaavedra you need to run 
```
``` 

to generate the new ThermalStandard code

src/descriptors/power_system_structs.json
raphaelsaavedra: I have removed the default for this field as requested by @jd-lara

src/descriptors/power_system_structs.json
raphaelsaavedra: Removed

src/descriptors/power_system_structs.json
raphaelsaavedra: Removed

PR #564, opened 2020-07-31
fixing vom cost omission in table data parsing

claytonpbarrows:


src/parsers/TAMU_data.jl
```diff
@@ -65,7 +65,7 @@ function TamuSystem(tamu_folder::AbstractString; kwargs...)
         push!(fixed_cols, "load_" * join(lsplit[2:3], "_"))
     end
 
-    loads = CSV.read(load_file, skipto = 3, header = fixed_cols)
+    loads = DataFrames.DataFrame(CSV.File(load_file, skipto = 3, header = fixed_cols))
```
nickrobinson251, line 68: can this use `copycols=false`? I believe that would be more performant (if that's a consideration, and it's safe to avoid the copies here)

```suggestion
    loads = DataFrames.DataFrame(CSV.File(load_file, skipto = 3, header = fixed_cols), copycols = false)
```

claytonpbarrows: Thanks!

PR #566, opened 2020-08-03
Update .travis.yml

jd-lara:


.travis.yml
```diff
@@ -5,7 +5,7 @@ os:
   - osx
   - windows
 julia:
-  - 1.4
+  - 1.5
```
nickrobinson251, line 8: Any reason not to use just  `1`, to get the latest stable 1.x release?

jd-lara: we normally transition these after checking nothing broke locally. We got traumatized by the changes between 1.0 and 1.2 that broke stuff for us.

PR #571, opened 2020-08-05
Fix serialization for StaticReserveGroup

daniel-thom:
I added documentation in this PR to make it easier for future contributors.  I think that documentation needs more content.  I intend this to be a starting point, not something final.

### DIFF
--- a/docs/src/developer/adding_types.md
+++ b/docs/src/developer/adding_types.md
@@ -1,0 +1,123 @@
+# Adding PowerSystems Types
+
+This page describes how developers should add types to PowerSystems. Refer to
+this
+[guide](https://nrel-siip.github.io/InfrastructureSystems.jl/latest/man/guide/#Component-structs)
+first.
+
+## Type Hierarchy
+All structs that correlate to power system components must be subtypes of the
+``Component`` abstract type. Browse its type hierachy to choose an appropriate
+supertype for your new struct.
+
+## Auto-generating Structs
+Most PowerSystems structs are auto-generated from the JSON descriptor file
+``src/descriptors/power_system_structs.json``. You can add your new struct
+here or write it manually.
+
+If all you need is the basic struct definition and getter/setter functions then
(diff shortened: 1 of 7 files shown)


jd-lara:
The PR looks good to me, it needs a check that this didn't break deserialization of v0.25 files.

daniel-thom:
> 
> 
> The PR looks good to me, it needs a check that this didn't break deserialization of v0.25 files.

I think it's safe but would be happy to do more verification.  Is the next step on you or me?

jd-lara:
> > The PR looks good to me, it needs a check that this didn't break deserialization of v0.25 files.
> 
> I think it's safe but would be happy to do more verification. Is the next step on you or me?

Just checking if the serialization broke is enough. Just need to make sure if I can tag v0.25.1

docs/src/developer/adding_types.md
```diff
@@ -0,0 +1,123 @@
+# Adding PowerSystems Types
```
daniel-thom, line 1: Here's a link with rendered HTML:  https://github.com/daniel-thom/PowerSystems.jl/blob/static-reserve-group-serialization/docs/src/developer/adding_types.md

src/models/devices.jl
```diff
@@ -1,9 +1,4 @@
-"""
-Add a service to a device.
-
-Throws ArgumentError if the service is already attached to the device.
-"""
-function add_service!(device::Device, service::Service)
+function add_service_internal!(device::Device, service::Service)
```
daniel-thom, line 1: @jd-lara Please challenge this change if you think it could be problematic.  I made it because I think the former implementation was a bit unsafe.  A user could add a service to a device without proper checks.  The public interface is `add_service!(sys, services, contributing_devices)`

jd-lara: I think this is fine, it might break some scripts but not PSI code.

PR #574, opened 2020-08-07
regulation device patch

### LINKED ISSUE #573: UnitsInfo lost in forwarding of RegulationDevice
When forwarding a device from regulation device, the units' info is lost.

jd-lara:
closes #573 but the serialization broke. Any hints @daniel-thom ?

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -1457,7 +1457,6 @@ end
 function handle_component_addition!(sys::System, component::RegulationDevice)
     copy_forecasts!(component.device, component)
     remove_component!(sys, component.device)
-    set_unit_system!(component.device, component.internal.units_info)
     return
 end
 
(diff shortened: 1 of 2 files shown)


daniel-thom:
I suspect that you need to define a new `set_unit_system!` method for `RegulationDevice` that actually sets the underlying device.

The code you fixed doesn't run during deserialization.  Remember this patch:  https://github.com/NREL-SIIP/PowerSystems.jl/blob/master/src/base.jl#L1257.

The current code likely also wouldn't work if the user called set_unit_system! on an existing `RegulationDevice`.

src/base.jl
```diff
@@ -1457,6 +1457,7 @@ end
 function handle_component_addition!(sys::System, component::RegulationDevice)
     copy_forecasts!(component.device, component)
     remove_component!(sys, component.device)
+    set_unit_system!(component.device, component.internal.units_info)
```
daniel-thom, line 1460: I don't think this is necessary.  It is already handled by same call made in `add_component!`.

PR #582, opened 2020-08-16
Psse parse madness

jd-lara:
@rodrigomha we should add a test that parses the 2000 Bus system.


rodrigomha:
Yes. I open a PR in Test Data to fix the 2000 bus system data. Can you check it out?

jd-lara:
> Yes. I open a PR in Test Data to fix the 2000 bus system data. Can you check it out?

Merged

src/descriptors/power_system_structs.json
```diff
@@ -3020,118 +3044,136 @@
       "supertype": "AVR"
     },
     {
-      "struct_name": "AVRTypeII",
-      "docstring": "Parameters of an Automatic Voltage Regulator Type II -  Typical static exciter model",
+      "struct_name": "ESDC2A",
+      "docstring": "Is used to represent field-controlled dc commutator exciters with continuously acting voltage regulators having power supplies derived from the generator or auxiliaries bus.. \nParameters of IEEE Std 421.5 Type DC2A Excitacion System. This model corresponds to ESDC2A in PSSE and PSLF",
```
daniel-thom, line 3048:
```suggestion
      "docstring": "Is used to represent field-controlled dc commutator exciters with continuously acting voltage regulators having power supplies derived from the generator or auxiliaries bus.\nParameters of IEEE Std 421.5 Type DC2A Excitacion System. This model corresponds to ESDC2A in PSSE and PSLF",
```

daniel-thom: Something to consider: we could detect long lines in the code generation script and add line breaks.  This might add consistency.  You could still add them in places where you want itemized lists.

src/descriptors/power_system_structs.json
```diff
@@ -4139,75 +4217,2672 @@
...
+          "data_type": "Float64",
+          "valid_range": {
+            "min": 0,
+            "max": 10
+          },
+          "validation_action": "warn"
+        },
+        {
+          "name": "Tk",
+          "comment": "Voltage Regulator lead time xonstant",
```
daniel-thom, line 4696:
```suggestion
          "comment": "Voltage Regulator lead time constant",
```

src/utils/conversion.jl
```diff
@@ -0,0 +1,7 @@
+"Convert Tuple to Min Max Named Tuple"
+Base.convert(::Type{Min_Max}, input::Tuple{Float64, Float64}) =
+    (min = input[1], max = input[2])
+
+"Convert Tuple to Up Down Named Tuple"
+Base.convert(::Type{Up_Down}, input::Tuple{Float64, Float64}) =
+    (up = input[1], down = input[2])
```
daniel-thom, line 7: Are these necessary?  I don't see a call site. You can do this instead:
```
a = (3.4, 5.6)
Min_Max(a)
(min = 3.4, max = 5.6)
```

daniel-thom:
The amount of code (especially structs and functions) tied to dynamics is becoming quite large.  Should we consider breaking them up?  We could put them in another module within PowerSystems.  Example:  `PowerSystems.Dynamics.AVRFixed`

src/descriptors/power_system_structs.json
jd-lara: I think we tried that and the issue was it breaks JSON syntax

jd-lara:
> The amount of code (especially structs and functions) tied to dynamics is becoming quite large. Should we consider breaking them up? We could put them in another module within PowerSystems. Example: `PowerSystems.Dynamics.AVRFixed`

I don't think that is an issue. The whole point of PSY is precisely to contain all these structs and allow interfacing with the other structs.

src/utils/conversion.jl
jd-lara: The code is used in the parsing of the dynamics. These are defined and automatically used by Julia when this happens

```Julia
struct foo
    a::Min_Max
end

foo((1.0, 2.0))
```

src/descriptors/power_system_structs.json
daniel-thom: I don't follow.  The JSON strings would not have new line characters.  The script would detect long lines and add new line characters in the generated Julia comments.

PR #587, opened 2020-08-18
Add costs to reserves

andrewrosemberg:
Adds operating costs to reserves (Static and variable)

### DIFF
--- a/src/descriptors/power_system_structs.json
+++ b/src/descriptors/power_system_structs.json
@@ -2750,6 +2750,13 @@
           "default": "InfrastructureSystems.Forecasts()"
         },
         {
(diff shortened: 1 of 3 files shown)


andrewrosemberg:
> @andrewrosemberg looks like this PR only adds costs to Static reserves. Can you add it to Variable too, please?

@jd-lara this seems to be a bug in Github haha (first time I see such a thing) if you go to the appropriate files you will see the changes.

andrewrosemberg:
@jd-lara sorry about that. I added the cost to `VariableReserve` in my last PR instead to `VariableReserveNonSinning`. So did the opposite on this one :)

PR #591, opened 2020-08-19
change errors to warn

### LINKED ISSUE #583: changing the upper value of lines and transformers' resistance, reactance and susceptance.
The current default upper limits are defined to avoid outliers and to avoid cases where ac-opf fails. However, we want to be able to accept values out of the defined range accompanied by a warning error related to possible issues with ac run.

jd-lara:
closes #583

### DIFF
--- a/src/descriptors/power_system_structs.json
+++ b/src/descriptors/power_system_structs.json
@@ -385,7 +385,7 @@
             "min": 0,
             "max": 4
           },
-          "validation_action": "error"
+          "validation_action": "warn"
         },
         {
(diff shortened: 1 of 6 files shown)


daniel-thom:
Who set these originally?  If it was @claytonpbarrows then he should be the approver.  If it was you then I approve.

jd-lara:
> Who set these originally? If it was @claytonpbarrows then he should be the approver. If it was you then I approve.

It was shared between me and Clayton earlier in PSY. I switched the reviewer

PR #593, opened 2020-08-20
Simplify System constructor

daniel-thom:
This is a breaking change for any script that constructs a System from vectors of components.  This should be easier for users to understand and use.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -27,24 +27,21 @@ A power system defined by fields for base_power, components, and forecasts.
 
 ```julia
 System(base_power)
-System(components, base_power)
-System(buses, generators, loads, branches, storage, base_power, services, annex; kwargs...)
-System(buses, generators, loads, base_power; kwargs...)
+System(base_power, buses, components...)
+System(base_power, buses, generators, loads, branches, storage, services; annex, kwargs...)
+System(base_power, buses, generators, loads; kwargs...)
 System(file; kwargs...)
 System(; buses, generators, loads, branches, storage, base_power, services, annex, kwargs...)
 System(; kwargs...)
 ```
 
 # Arguments
-- `buses::Vector{Bus}`: an array of buses
-- `generators::Vector{Generator}`: an array of generators of (possibly) different types
-- `loads::Vector{ElectricLoad}`: an array of load specifications that includes timing of the loads
-- `branches::Union{Nothing, Vector{Branch}}`: an array of branches; may be `nothing`
-- `storage::Union{Nothing, Vector{Storage}}`: an array of storage devices; may be `nothing`
 - `base_power::Float64`: the base power value for the system
-- `services::Union{Nothing, Vector{<:Service}}`: an array of services; may be `nothing`
+- `buses::Vector{Bus}`: an array of buses
+- `components...`: Each element must be an iterable containing subtypes of Component.
 
 # Keyword arguments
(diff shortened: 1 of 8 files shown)


src/base.jl
```diff
@@ -101,130 +98,44 @@ function System(data, base_power; kwargs...)
...
     load_zones = isnothing(annex) ? nothing : get(annex, :LoadZone, nothing)
     if !isnothing(load_zones)
         for lz in load_zones
-            try
-                add_component!(sys, lz)
-            catch e
-                if isa(e, IS.InvalidRange)
-                    error_detected = true
-                else
-                    rethrow()
```
daniel-thom, line 155: I think this exception handling was out-dated and redundant.

test/test_powersystemconstructors.jl
```diff
@@ -121,7 +95,7 @@ end
     # with nullable values. Consider adding them so that this "demo-constructor" works.
     # If that isn't appropriate for this type, add it to types_to_skip below.
 
-    types_to_skip = (System, TestDevice, TestRenDevice)
```
daniel-thom, line 124: This was a bug that manifested if you `include` this test file without first including the test where these types are defined (which happens to be later alphabetically).  There should not be a need to skip these types.

src/base.jl
```diff
@@ -27,24 +27,21 @@ A power system defined by fields for base_power, components, and forecasts.
...
-- `loads::Vector{ElectricLoad}`: an array of load specifications that includes timing of the loads
-- `branches::Union{Nothing, Vector{Branch}}`: an array of branches; may be `nothing`
-- `storage::Union{Nothing, Vector{Storage}}`: an array of storage devices; may be `nothing`
 - `base_power::Float64`: the base power value for the system
-- `services::Union{Nothing, Vector{<:Service}}`: an array of services; may be `nothing`
+- `buses::Vector{Bus}`: an array of buses
+- `components...`: Each element must be an iterable containing subtypes of Component.
 
 # Keyword arguments
+- `annex::Dict`: Defines LoadZones if key :LoadZone is present
```
claytonpbarrows, line 44: should we change this to `ext` for consistency with our components?

claytonpbarrows: > should we change this to `ext` for consistency with our components?

if so, we can do so in a separate PR

jd-lara: We should also remove the comment since now we have a topology object for this

daniel-thom: I don't think I saw it being used.  Not if any other script is using it.  Could we remove the option altogether? The user can pass LoadZone object in `components...`.

jd-lara: I think we should keep the option of adding ext.

daniel-thom: We already have `ext` for `System`. Two questions:
1. Do you want to allow the user to pass it here?  If you say yes then we should probably make sure all of the constructors support that.  I could do that here.
2. I think it would be better to remove LoadZone handling here.  That's contingent on whether we can break backwards compatibility. Thoughts?

jd-lara: I don't think anyone is creating LoadZones like that

daniel-thom:
> LGTM. Do we need to update the test data in PSI?

Yes.  The code in test/test_utils/get_test_data.jl will need to be updated.

src/base.jl
claytonpbarrows: I think it's ok to break bw compatibility for this.

PR #595, opened 2020-08-25
Added struct for Zero Order Battery Model (ZeroOrderBESS.jl)

ciaranrob:
Created a struct for a zero order battery energy storage system to capture DC-side dynamical behavior. Includes representation of battery, DC/DC converter and DC/DC controller

### DIFF
--- a/src/descriptors/power_system_structs.json
+++ b/src/descriptors/power_system_structs.json
@@ -9056,6 +9056,152 @@
       "supertype": "DCSource"
     },
     {
+      "struct_name": "ZeroOrderBESS",
+      "docstring": "Parameters for the DC-side with a Battery Energy Storage System",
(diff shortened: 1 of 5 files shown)


src/descriptors/power_system_structs.json
```diff
@@ -9055,6 +9055,152 @@
       ],
       "supertype": "DCSource"
     },
+    {
+      "struct_name": "ZeroOrderBESS",
+      "docstring": "Parameters for the DC-side with a Battery Energy Storage System",
```
jd-lara, line 9060: Can you provide more detail about the model in the docstring, please. If possible a reference to where the model came from a paper or standard.

ciaranrob:
Requested changes implemented

PR #596, opened 2020-08-26
cost function updates

jd-lara:
@andrewrosemberg this PR adds an additional cost function to support market bids. Please review to check this has all the fields you need. See the file https://github.com/NREL-SIIP/PowerSystems.jl/blob/c8f277c3c217e0caf403e7be4b7e870d1eb43dfe/src/models/generated/MarketBidCost.jl

### DIFF
--- a/src/descriptors/power_system_structs.json
+++ b/src/descriptors/power_system_structs.json
@@ -129,9 +129,9 @@
         {
           "name": "ancillary_services",
           "comment": "Bids for the ancillary services",
-          "null_value": "Dict{String, Float}()",
+          "null_value": "Dict{String, Float64}()",
           "data_type": "Dict{String, Float64}",
-          "default": "Dict{String, Float}()"
+          "default": "Dict{String, Float64}()"
         },
         {
           "name": "forecasts",
@@ -2214,7 +2214,7 @@
           "name": "power_trajectory",
           "comment": "Power trajectory the unit will take during the start-up and shut-down ramp process",
           "null_value": "nothing",
-          "data_type": "Union{Nothing, NamedTuple{(:start_up, :shut_down), Tuple{Float64, Float64}}}",
+          "data_type": "Union{Nothing, NamedTuple{(:startup, :shutdown), Tuple{Float64, Float64}}}",
           "valid_range": {
             "min": 0,
             "max": null
--- a/src/descriptors/power_system_structs.json
+++ b/src/descriptors/power_system_structs.json
@@ -113,7 +113,7 @@
           "name": "start_up",
           "comment": "start-up cost at different stages of the thermal cycle. Warm is also refered as intermediate in some markets",
           "null_value": "(hot = START_COST, warm = START_COST, cold = START_COST)",
-          "data_type": "NamedTuple{(:hot, :warm :cold), NTuple{3, Float64}}"
+          "data_type": "NamedTuple{(:hot, :warm, :cold), NTuple{3, Float64}}"
         },
         {
           "name": "shut_down",
--- a/src/descriptors/power_system_structs.json
+++ b/src/descriptors/power_system_structs.json
@@ -129,9 +129,9 @@
         {
           "name": "ancillary_services",
           "comment": "Bids for the ancillary services",
-          "null_value": "Dict{String, Float64}()",
-          "data_type": "Dict{String, Float64}",
-          "default": "Dict{String, Float64}()"
+          "null_value": "IdDict{Any,Float64}()",
+          "data_type": "IdDict{Any,Float64}",
+          "default": "IdDict{Any,Float64}()"
         },
         {
           "name": "forecasts",
--- a/src/descriptors/power_system_structs.json
+++ b/src/descriptors/power_system_structs.json
@@ -55,7 +55,7 @@
           "data_type": "Float64"
         },
         {
-          "name": "startup",
+          "name": "start_up",
           "comment": "startup cost",
           "null_value": "0.0",
           "data_type": "Float64",
@@ -66,7 +66,7 @@
           "validation_action": "warn"
         },
         {
-          "name": "shutdn",
+          "name": "shut_down",
           "comment": "shutdown cost",
           "null_value": "0.0",
           "data_type": "Float64",
--- a/src/descriptors/power_system_structs.json
+++ b/src/descriptors/power_system_structs.json
@@ -98,22 +98,22 @@
       "docstring": "Data Structure Operational Cost to reflect market bids of energy and ancilliary services.\nCompatible with most US Market bidding mechanisms",
       "fields": [
         {
-          "name": "energy_bid",
-          "comment": "variable cost",
+          "name": "variable",
+          "comment": "variable cost representing the energy bid",
           "null_value": "VariableCost((0.0, 0.0))",
           "data_type": "VariableCost"
         },
         {
-          "name": "no_load_cost",
+          "name": "no_load",
           "comment": "no load cost",
           "null_value": "0.0",
           "data_type": "Float64"
         },
         {
           "name": "start_up",
-          "comment": "start-up cost",
-          "null_value": "(hot = START_COST, int = START_COST, cold = START_COST)",
-          "data_type": "NamedTuple{(:hot, :int, :cold), NTuple{3, Float64}}"
+          "comment": "start-up cost at different stages of the thermal cycle. Warm is also refered as intermediate in some markets",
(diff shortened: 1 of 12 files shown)


src/models/generated/MarketBidCost.jl
```diff
@@ -0,0 +1,92 @@
+#=
+This file is auto-generated. Do not edit.
+=#
+"""
+    mutable struct MarketBidCost <: OperationalCost
+        energy_bid::VariableCost
+        no_load_cost::Float64
+        start_up::NamedTuple{(:hot, :int, :cold), NTuple{3, Float64}}
```
claytonpbarrows, line 8: `int`? I'd prefer `warm` if it's appropriate, or `intermediate` ...

src/descriptors/power_system_structs.json
```diff
@@ -94,36 +94,30 @@
...
           "data_type": "Float64"
         },
         {
-          "name": "fixed",
-          "comment": "fixed cost",
-          "null_value": "0.0",
-          "data_type": "Float64"
+          "name": "start_up",
+          "comment": "start-up cost",
+          "null_value": "(hot = START_COST, int = START_COST, cold = START_COST)",
```
claytonpbarrows, line 115: could we also add a more substantial comment too describe what the intended usage is?

andrewrosemberg:
Is there a reason why `operation_cost::ThreePartCost` in [ThermalStandard](https://github.com/NREL-SIIP/PowerSystems.jl/blob/d4ffbcc1c6ce97e799e2cf079786f54db00a881b/src/models/generated/ThermalStandard.jl#L54) can not change to this? If not do we have to use multi-start?

jd-lara:
We have to add another fix to allow abstract costs.

src/models/generated/MarketBidCost.jl
raphaelsaavedra: +1 for changing this name, `warm` seems better

src/descriptors/power_system_structs.json
```diff
@@ -94,36 +94,30 @@
...
         {
-          "name": "variable",
+          "name": "energy_bid",
           "comment": "variable cost",
           "null_value": "VariableCost((0.0, 0.0))",
           "data_type": "VariableCost"
         },
         {
-          "name": "no_load",
+          "name": "no_load_cost",
```
raphaelsaavedra, line 107:
```suggestion
          "name": "no_load",
```
I'd prefer to keep it as `no_load` for consistency with the other names that don't have `_cost` (e.g. `start_up`, `shut_down`, `fixed`, ...)

src/descriptors/power_system_structs.json
```diff
@@ -94,36 +94,30 @@
...
     },
     {
-      "struct_name": "MultiStartCost",
-      "docstring": "Data Structure Operational Cost Data which includes fixed, variable cost, multiple start up cost and stop costs.",
+      "struct_name": "MarketBidCost",
+      "docstring": "Data Structure Operational Cost to reflect market bids of energy and ancilliary services.\nCompatible with most US Market bidding mechanisms",
       "fields": [
         {
-          "name": "variable",
+          "name": "energy_bid",
```
raphaelsaavedra, line 101:
```suggestion
          "name": "variable",
```
I would also not change the name here as effectively it's the same as `variable` in `ThreePartCost` or other types. Better to be consistent and have documentation to make it clear if we judge necessary.

src/descriptors/power_system_structs.json
```diff
@@ -132,6 +126,13 @@
           },
           "validation_action": "warn"
         },
+        {
+          "name": "ancillary_services",
+          "comment": "Bids for the ancillary services",
+          "null_value": "Dict{String, Float64}()",
+          "data_type": "Dict{String, Float64}",
```
raphaelsaavedra, line 133: What is the idea behind this dict exactly? Is it something like

```
"regulation" => C^{reg}
"spinning" => C^{spin}
...
```

or maybe having the strings as the names of the services?

jd-lara: the strings should match the names of the services. We will probably want something that can do this automatically when running add service but this will come later as we refine the interface.

raphaelsaavedra: In that case, is there any reason not to have the dict as a `Dict{Service, Float64}`?

src/models/generated/MarketBidCost.jl
jd-lara: I had warm but the forms ask for intermediate as the name of the field. But if this is preferred I will fix it.

src/descriptors/power_system_structs.json
daniel-thom: The service structs are mutable, so they should not be keys of a dictionary.  You could use `IdDict`, however.

If (1) you need to be able to find the services objects and (2) the services can be any subtype of `Service` then storing only the name is technically not sufficient for looking up a service with `get_component`.  You need the concrete type as well.  That's why we have `ServiceContributingDevicesKey`.  I may not have known about `IdDict` when I implemented that.

If these objects will be stored in the `System` and you do include the actual service objects then some customization will be needed for serialization.  Keys of JSON objects must be strings.  We would have to substitute their UUIDs.

jd-lara: @daniel-thom if the services are already stored in the same device, this should be simpler. The search would go in the other direction using the service in the device to find the bid.

daniel-thom: OK, that makes sense.  Are you still vulnerable to a name collision?  Services of different concrete types could have the same name.  All code that adds to this dict could assert uniqueness as a guarantee.

jd-lara: Given that dicts don't allow for duplicate keys it will require a sequence of user mistakes to end up with such a situation.

daniel-thom: What would be the user mistakes?  In your comment above you said that we will eventually have code that automatically builds this dictionary.  Users are allowed to have duplicate names across different concrete types.  So, they could end up with a collision and not know it.

jd-lara: The error will entail creating the cost function with the names of the services first and then adding to the device 2 services with the exact same name after. 

To avoid this one option is to do `add_service!()` and pass the bid and recorded in the cost function. The other option is that users add the key to the dict when making the cost function but we won't be able to guarantee there is a service with that name.

daniel-thom: Even if this is all very unlikely I think we should keep the design consistent.  The design allows a device to have services of differing types.  Therefore, it allows a device to have services with the same name as long as the types are different.  We could add restrictions to this, of course.

If this dictionary will be an internally-managed field (never created or accessed by the user) then you could use the UUID instead of the name.

jd-lara: As discussed, we will go with `IdDicts` such that the key to the bid is the service in questions. Please note that this will require creating the component and the service BEFORE been able to add the bids. As a result, by default, the bids dict will always start empty.

jd-lara:
@daniel-thom further tests require merging #597 and the new serialization.

daniel-thom:
> 
> 
> @daniel-thom further tests require merging #597 and the new serialization.

Sounds good.  What is your preferred sequencing here?  I expect to submit the serialization PR on Monday and it will likely take 1-3 days for reviews & modifications.

jd-lara:
> > @daniel-thom further tests require merging #597 and the new serialization.
> 
> Sounds good. What is your preferred sequencing here? I expect to submit the serialization PR on Monday and it will likely take 1-3 days for reviews & modifications.

I think we can merge this PR, then #597 since it is still in draft (cc. @claytonpbarrows) and then fix the serialization.

daniel-thom:
I suspect that you need a helper function to retrieve the `Float64` stored as values in `ancillary_services` so that users don't have to access that dictionary.  That will allow you to change the implementation in the future.  Something like

```
get_service_bid(market_bid::MarketBid, service::Service) = market_bid.ancillary_services[service]
```

but with error checking.

jd-lara:
> I suspect that you need a helper function to retrieve the `Float64` stored as values in `ancillary_services` so that users don't have to access that dictionary. That will allow you to change the implementation in the future. Something like
> 
> ```
> get_service_bid(market_bid::MarketBid, service::Service) = market_bid.ancillary_services[service]
> ```
> 
> but with error checking.

Yes, I will add it while doing https://github.com/NREL-SIIP/PowerSimulations.jl/pull/516 so I can add interfaces useful to PSI modeling.

PR #597, opened 2020-08-26
adding PHES

claytonpbarrows:


src/descriptors/power_system_structs.json
```diff
@@ -1435,9 +1435,9 @@
         },
         {
           "name": "operation_cost",
-          "null_value": "TwoPartCost(nothing)",
-          "data_type": "TwoPartCost",
-          "comment": "Operation Cost of Generation [`TwoPartCost`](@ref)"
+          "null_value": "ThreePartCost(nothing)",
```
jd-lara, line 1438: We should make this a Union of types or abstract on OperationalCost? So we can support also the MarketBid cost.

src/descriptors/power_system_structs.json
```diff
@@ -1673,6 +1673,255 @@
...
+          "validation_action": "warn"
+        },
+        {
+          "name": "prime_mover",
+          "comment": "Prime mover technology according to EIA 923",
+          "null_value": "PrimeMovers.HY",
+          "data_type": "PrimeMovers.PrimeMover"
+        },
+        {
+          "name": "active_power_limits",
```
jd-lara, line 1737: Do we need to distinguish active power limits by the direction a la storage?

test/test_printing.jl
```diff
@@ -24,7 +24,8 @@ function are_type_and_fields_in_output(obj::T) where {T <: Component}
 
     for (fieldname, fieldtype) in zip(fields, fieldtypes(T))
         if isnothing(getfield(obj, fieldname)) ||
-           fieldtype <: IS.InfrastructureSystemsInternal
+           fieldtype <: IS.InfrastructureSystemsInternal ||
+           fieldtype <: PowerSystems.OperationalCost
```
claytonpbarrows, line 28: @daniel-thom I added this because I couldn't figure out how to extend the conditional in line 39 to catch all subtypes. Let me know if you have other ideas.

daniel-thom: There are likely two problems:
1. Sometimes the field type is `OperationCost` which won't show up.
2. Sometimes the field type is `Union{Nothing, OperationCost}`

Try changing line 34 to
```
val = getfield(obj, fieldname)
expected = string(typeof(val))
```
Assignment of `val` could be made at the top of the for loop.

src/parsers/power_system_table_data.jl
```diff
@@ -729,9 +730,9 @@ function make_generator(data::PowerSystemTableData, gen, cost_colnames, bus)
         generator = make_thermal_generator(data, gen, cost_colnames, bus)
     elseif gen_type == ThermalMultiStart
         generator = make_thermal_generator_multistart(data, gen, cost_colnames, bus)
-    elseif gen_type <: HydroGen
+    elseif !isnothing(gen_type) && gen_type <: HydroGen
```
daniel-thom, line 733: This would be simpler (and make future additions easier) if you handled `isnothing(gen_type)` explicitly.

PR #599, opened 2020-09-01
Refactor serialization

daniel-thom:
This is tied to https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/120.  Tests will not pass until that is merged.

This new code will not support systems serialized with previous versions.

- Replace JSON2 with JSON3.
- Serialize to standard types with custom methods and then to JSON.
- Add kwarg-only constructors to all component structs.
- Add Arcs to the System.
- Change dynamic injectors so that they no longer contain references to static injectors.
- Change argument order in copy_forecasts! and export it.
- Change Component supertype to align with new InfrastructureSystems.
- Make all non-component structs subtypes of InfrastructureSystemsType.
- Remove forecasts and internal from OperationalCost structs.


src/models/dynamic_generator.jl
```diff
@@ -97,32 +84,72 @@ function DynamicGenerator(
 end
 
 function DynamicGenerator(;
-    static_injector::Generator,
+    name::Union{String, Generator},
```
daniel-thom, line 87: @jd-lara We need to discuss the workflows here.  Does the user pass the generator, its name, or either?

src/models/generated/ESAC6A.jl
```diff
@@ -43,7 +43,7 @@ Parameters of IEEE Std 421.5 Type AC6A Excitacion System. ESAC6A in PSSE and PSL
 - `Tc::Float64`: Regulator numerator (lead) time constant in s, validation range: `(0, 20)`, action if invalid: `warn`
 - `Va_lim::NamedTuple{(:min, :max), Tuple{Float64, Float64}}`: Limits for regulator output `(Va_min, Va_max)`
 - `Vr_lim::NamedTuple{(:min, :max), Tuple{Float64, Float64}}`: Limits for exciter field voltage `(Vr_min, Vr_max)`
-- `Te::Float64`: Exciter field time constant, validation range: `("eps()", 2.0)`, action if invalid: `error`
+- `Te::Float64`: Exciter field time constant, validation range: `("eps()", 2)`, action if invalid: `error`
```
daniel-thom, line 46: There are many unintentional changes of floats to integers in the auto-generated code.  I'll look into this.

Project.toml
```diff
@@ -12,6 +12,7 @@ InfrastructureSystems = "2cd47ed4-ca9b-11e9-27f2-ab636a7671f1"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
 JSON2 = "2535ab7d-5cd8-5a07-80ac-9b1792aadce3"
```
nickrobinson251, line 14: can JSON and/or JSON2 be removed now?

src/models/OuterControl.jl
```diff
@@ -41,8 +41,38 @@ function OuterControl(
...
-function OuterControl(; active_power, reactive_power, ext = Dict{String, Any}())
-    OuterControl(active_power, reactive_power, ext)
+function OuterControl(;
+    active_power,
+    reactive_power,
+    ext = Dict{String, Any}(),
+    states = nothing,
+    n_states = nothing,
+)
+    if isnothing(states)
```
nickrobinson251, line 51: helps inference  https://github.com/JuliaLang/julia/pull/36444
```suggestion
    if states === nothing
```
same applies to multiple places in the PR :)

src/models/OuterControl.jl
```diff
@@ -41,8 +41,38 @@ function OuterControl(
...
+    for (field_name, field_type) in zip(fieldnames(OuterControl), fieldtypes(OuterControl))
+        val = data[string(field_name)]
+        if field_name == :active_power
+            vals[field_name] = deserialize(T.parameters[1], val)
+        elseif field_name == :reactive_power
+            vals[field_name] = deserialize(T.parameters[2], val)
+        elseif field_name == :states
+            vals[field_name] = [Symbol(x) for x in val]
+        else
+            vals[field_name] = deserialize(field_type, val)
```
nickrobinson251, line 71:
```suggestion
        vals[field_name] = if field_name === :active_power
            deserialize(T.parameters[1], val)
        elseif field_name === :reactive_power
            deserialize(T.parameters[2], val)
        elseif field_name === :states
            [Symbol(x) for x in val]
        else
            deserialize(field_type, val)
```
similar cahnges apply to multiple places in the PR :)

src/models/serialization.jl
```diff
@@ -1,98 +1,115 @@
...
-    elseif data.cost[1] isa Array
-        variable = Vector{Tuple{Float64, Float64}}()
-        for array in data.cost
-            push!(variable, Tuple{Float64, Float64}(array))
-        end
-    else
-        @assert data.cost isa Tuple || data.cost isa Array
-        variable = Tuple{Float64, Float64}(data.cost)
+should_encode_as_uuid(val) = mapreduce(x -> val isa x, |, _ENCODE_AS_UUID_B)
+should_encode_as_uuid(::Type{T}) where {T} = mapreduce(x -> T <: x, |, _ENCODE_AS_UUID_A)
```
nickrobinson251, line 14: is this the same as `any` or `all`? e.g.`all(x -> T <: x,  _ENCODE_AS_UUID_A)`

src/parsers/enums.jl
```diff
@@ -10,8 +10,18 @@
     STORAGE
 end
 
+ENUMS = (
```
nickrobinson251, line 13:
```suggestion
const ENUMS = (
```

Project.toml
daniel-thom: Yes, thanks.  I removed them in InfrastructureSystems but not here.

src/models/OuterControl.jl
daniel-thom: I have no problem changing to this.  However, if you're not supposed to use a Julia standard library call then maybe it shouldn't exist?  Or come with a recommendation on when to use it?

src/parsers/enums.jl
daniel-thom: Thanks

src/models/serialization.jl
daniel-thom: Yes, that would be better.  I'll change it.

src/models/generated/ESAC6A.jl
daniel-thom: @jd-lara JSON3 is the cause of this.  Let me know if this is problematic for you.  We could switch back to JSON with minimal effort.  The only reason I chose JSON3 is that it was up to 4 times faster at reading our data.  We could use JSON for reading our struct descriptors and JSON3 for everything else.  The cost is an extra project dependency.  I will note that there is no correctness issue here.

These examples illustrate the problem:
```
julia> JSON3.read(JSON3.write(Dict("a" => 1.0)), Dict)
Dict{String,Any} with 1 entry:
  "a" => 1
```
```
julia> JSON.parse(JSON.json(Dict("a" => 1.0)))
Dict{String,Any} with 1 entry:
  "a" => 1.0
```

jd-lara: I don't think this is a big deal, Julia handles the conversion. It's better to keep the speed. We can switch back to JSON.jl whenever they decide to release 1.0.

jd-lara (approved):
I will approve but the changes are too broad to do a comprehensive review. Based on the tests and the demonstrations made so far I will say let's go ahead and merge. We need to test this with the other packages too.

PR #602, opened 2020-09-03
add Hybrid System

jd-lara:
This PR adds a basic struct for Hybrid Devices. Still doesn't address the addition to the system. 

Pending items: 

- [ ] Check bus in all underlying devices is the same as the `HybridSystems`
- [ ] Don't add the underlying devices to the System.

### DIFF
--- a/src/models/HybridSystem.jl
+++ b/src/models/HybridSystem.jl
@@ -1,0 +1,172 @@
+"""
+Representation of a Hybrid System that collects renewable generation, load, thermal generation
+and storage
+"""
+mutable struct HybridSystem{
+    T <: ThermalGen,
+    L <: ElectricLoad,
+    S <: Storage,
+    R <: RenewableGen,
+} <: StaticInjection
+    name::String
+    available::Bool
+    status::Bool
+    bus::Bus
+    active_power::Float64
+    reactive_power::Float64
+    thermal_unit::T
+    electric_load::L
+    storage::S
+    renewable_unit::R
+    # PCC Data
+    "Thermal limited MVA Power Output of the unit. <= Capacity"
+    pcc_impedance::ComplexF64
+    rating::Float64
+    input_active_power_limits::NamedTuple{(:min, :max), Tuple{Float64, Float64}}
+    output_active_power_limits::NamedTuple{(:min, :max), Tuple{Float64, Float64}}
+    reactive_power_limits::Union{Nothing, Min_Max}
+    base_power::Float64
+    operation_cost::OperationalCost
+    "corresponding dynamic injection device"
+    services::Vector{Service}
+    dynamic_injector::Union{Nothing, DynamicInjection}
+    ext::Dict{String, Any}
+    "internal forecast storage"
+    forecasts::InfrastructureSystems.Forecasts
+    "power system internal reference, do not modify"
+    internal::InfrastructureSystemsInternal
+end
+
+function HybridSystem(;
+    name,
+    available,
+    status,
+    bus,
+    active_power,
+    reactive_power,
+    thermal_unit::T,
+    electric_load::L,
+    storage::S,
+    renewable_unit::R,
+    pcc_impedance,
+    rating,
+    input_active_power_limits,
+    output_active_power_limits,
+    reactive_power_limits,
+    base_power,
+    operation_cost,
+    services = Service[],
+    dynamic_injector = nothing,
+    ext = Dict{String, Any}(),
+    forecasts = InfrastructureSystems.Forecasts(),
+    internal = InfrastructureSystemsInternal(),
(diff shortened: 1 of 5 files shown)


src/models/HybridSystem.jl
```diff
@@ -0,0 +1,172 @@
...
+    bus::Bus
+    active_power::Float64
+    reactive_power::Float64
+    thermal_unit::T
+    electric_load::L
+    storage::S
+    renewable_unit::R
+    # PCC Data
+    "Thermal limited MVA Power Output of the unit. <= Capacity"
+    pcc_impedance::ComplexF64
```
claytonpbarrows, line 23: I'm torn on the use of `pcc`. It's a fairly standard abbreviation, but it's still an abbreviation.  Also, is the `pcc_` even neede, could it be `impedance`.

src/models/HybridSystem.jl
```diff
@@ -0,0 +1,172 @@
...
+    rating::Float64
+    input_active_power_limits::NamedTuple{(:min, :max), Tuple{Float64, Float64}}
+    output_active_power_limits::NamedTuple{(:min, :max), Tuple{Float64, Float64}}
+    reactive_power_limits::Union{Nothing, Min_Max}
+    base_power::Float64
```
claytonpbarrows, line 28: How will this relate to the `base_power` of the components in the `thermal_unit` and other fields? Will this `base_power` only apply to the local fields while the sub-components will have their own?

claytonpbarrows (commented):
Generally this LGTM.
One potential requirement that I think is not satisfied is the ability to specify/limit power transfers behind the meter. It's also not clear to me what data exists to make such a specification.

src/models/HybridSystem.jl
jd-lara: This would be the base power of the PCC interconnection either given by the transformer or the output inverter. On the practical side, it is needed to scale the PCC input/output limits and rating.

src/models/HybridSystem.jl
jd-lara: I am fine with the change in name, just thought this would make it more clear

src/models/HybridSystem.jl
```diff
@@ -0,0 +1,172 @@
...
+    available::Bool
+    status::Bool
+    bus::Bus
+    active_power::Float64
+    reactive_power::Float64
+    thermal_unit::T
+    electric_load::L
+    storage::S
+    renewable_unit::R
+    # PCC Data
```
yinongsun, line 21: If I understand this correctly, the following couple parameters are to specify inverter (PCC here) properties? This seems to me is modeled as a DC-coupled system (i.e., shared inverter for the whole system). If we want to model AC-coupled systems, we'll need inverter properties for each component in the hybrid system. Would this representation be able to model it?

src/models/HybridSystem.jl
```diff
@@ -0,0 +1,172 @@
...
+    "Thermal limited MVA Power Output of the unit. <= Capacity"
+    pcc_impedance::ComplexF64
+    rating::Float64
```
yinongsun, line 24: A question here: so is this `rating` parameter for each component rating/size, or also for inverter rating? I'm just wondering which parameters specify the e.g ILR of the inverter.

src/models/HybridSystem.jl
jd-lara: If you need to model independent interconnections then the information is stored inside of each component that composes the hybrid system

src/models/HybridSystem.jl
jd-lara: Rating works as a proxy to limit the interconnection's max current. This can be implemented as 
```P^2 + Q^2 <= rating^2``` or ```V*I < rating``` depending on the modeling needs.

File: src/models/HybridSystem.jl [hunk 5]
```diff
@@ -0,0 +1,172 @@
...
+    operation_cost::OperationalCost
+    "corresponding dynamic injection device"
+    services::Vector{Service}
+    dynamic_injector::Union{Nothing, DynamicInjection}
+    ext::Dict{String, Any}
+    "internal forecast storage"
+    forecasts::InfrastructureSystems.Forecasts
+    "power system internal reference, do not modify"
+    internal::InfrastructureSystemsInternal
+end
```
claytonpbarrows, line 38: Per conversation with @yinongsun  and @bethanyfrew , we think that all of the PCC parameters and costs should be `Nothing`able. This allows the user to explicitly control the DC coupled vs AC coupled nature of the system along with the associated operational cost.

jd-lara: ok, I will probably come back to this after merging 1.0

src/models/HybridSystem.jl
jd-lara: I switched to interconnection

PR #605, opened 2020-09-04
remove setters for states and n_states in dynamic components

jd-lara:


daniel-thom (approved):

rodrigomha (approved):
Sounds good, although I don't know why is breaking.

PR #611, opened 2020-09-21
Add add_time_series! method that accepts multiple components

daniel-thom:
- Remove add_time_series! method that creates Deterministic and adds in
one step. Users now need to perform this individually.
- Support new NormalizationFactor type.

This removes add_time_series! methods that constructed Deterministic
and added it to the system in one step.

This is dependent on https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/125, and so tests will fail.


jd-lara (approved):
LGTM, needs the formatter run also

PR #612, opened 2020-09-28
Time series updates

daniel-thom:
This updates PowerSystems to with the pre-10 branch of InfrastructureSystems.  It also requires the pre-10 branch of PowerSystemsTestData.

A number of things are commented out.  Leaving this as a draft until we workout what to do with them.

All tests pass except for 
- Some in test_read_time_series.jl.  These failures are due to an interface change in IS.read_time_series.  We currently don't have a way to convert PSY-supported time series CSV files into TimeArray objects.  Leaving this unresolved until we determine whether we want to restore the prior functionality.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -415,7 +415,7 @@ function add_service!(
 end
 
 """
-Adds time series data from a metadata file or metadata descriptors.
+Add time series data from a metadata file or metadata descriptors.
 
 # Arguments
 - `sys::System`: system
@@ -423,11 +423,7 @@ Adds time series data from a metadata file or metadata descriptors.
   that includes an array of IS.TimeSeriesFileMetadata instances or a vector.
 - `resolution::DateTime.Period=nothing`: skip time series that don't match this resolution.
 """
-function add_time_series_from_file_metadata!(
-    sys::System,
-    metadata_file::AbstractString;
-    resolution = nothing,
-)
+function add_time_series!(sys::System, metadata_file::AbstractString; resolution = nothing)
     return IS.add_time_series_from_file_metadata!(
         sys.data,
         Component,
@@ -437,14 +433,14 @@ function add_time_series_from_file_metadata!(
 end
 
 """
-Adds time series data from a metadata file or metadata descriptors.
+Add time series data from a metadata file or metadata descriptors.
 
 # Arguments
 - `sys::System`: system
 - `timeseries_metadata::Vector{IS.TimeSeriesFileMetadata}`: metadata for timeseries
 - `resolution::DateTime.Period=nothing`: skip time series that don't match this resolution.
 """
-function add_time_series_from_file_metadata!(
+function add_time_series!(
     sys::System,
     file_metadata::Vector{IS.TimeSeriesFileMetadata};
     resolution = nothing,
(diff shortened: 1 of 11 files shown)


src/base.jl
```diff
@@ -423,8 +423,17 @@ Adds time series data from a metadata file or metadata descriptors.
   that includes an array of IS.TimeSeriesFileMetadata instances or a vector.
 - `resolution::DateTime.Period=nothing`: skip time series that don't match this resolution.
 """
-function add_time_series!(sys::System, metadata_file::AbstractString; resolution = nothing)
-    return IS.add_time_series!(Component, sys.data, metadata_file; resolution = resolution)
+function add_time_series_from_file_metadata!(
```
jd-lara, line 426: Can we keep the previous name? I think we can leave the details underneath

src/base.jl
```diff
@@ -938,34 +895,29 @@ Return a TimeSeriesData for a subset of the time series range stored for these p
 function get_time_series(
     ::Type{T},
     component::Component,
-    initial_time::Dates.DateTime,
-    label::AbstractString,
+    name::AbstractString,
```
jd-lara, line 898: The get_time_series method in IS allows passing an initial time and horizon here.

src/base.jl
```diff
@@ -974,11 +926,19 @@ Return a TimeSeries.TimeArray for the requested time series parameters.
...
-    initial_time::Dates.DateTime,
-    label::AbstractString,
-    horizon::Union{Nothing, Int} = nothing,
+    name::AbstractString;
+    start_time::Union{Nothing, Dates.DateTime} = nothing,
+    len::Union{Nothing, Int} = nothing,
+    count::Int = 1,
 ) where {T <: TimeSeriesData}
-    return IS.get_time_series_array(T, component, initial_time, label, horizon)
+    return IS.get_time_series_array(
```
jd-lara, line 934: Do we need to implement this for Deterministic?

src/base.jl
```diff
@@ -1008,11 +976,19 @@ Return an Array of values for the requested time series parameters.
...
-    initial_time::Dates.DateTime,
-    label::AbstractString,
-    horizon::Union{Nothing, Int} = nothing,
+    name::AbstractString;
+    start_time::Union{Nothing, Dates.DateTime} = nothing,
+    len::Union{Nothing, Int} = nothing,
+    count::Int = 1,
 ) where {T <: TimeSeriesData}
-    return IS.get_time_series_values(T, component, initial_time, label, horizon)
+    return IS.get_time_series_values(
```
jd-lara, line 984: Same here, do we have this for Deterministic?

jd-lara (approved):
The changes look good. Which methods are we missing here for Deterministic?

src/base.jl
daniel-thom: Agree. Must fix.

src/base.jl
daniel-thom: I think this supports Deterministic as it is.  Why do you disagree?

src/base.jl
daniel-thom: Same as above

src/base.jl
jd-lara: I am just inquiring to see if we need to test it.

daniel-thom: We should.  Can you add some test data to PowerSystems or PowerSystemsTestData?

jd-lara: Ok, I will make sure to make/add data for this test. I am almost sure that the 5-bus case already has data on this effect.

PR #614, opened 2020-09-30
Add `must_run` to ThermalStandard

raphaelsaavedra:


raphaelsaavedra:
I know that after editing the json I should run `InfrastructureSystems.generate_structs("./src/descriptors/power_system_structs.json", "./src/models/generated")` and then call JuliaFormatter.

However, as mentioned in slack in a conversation with @jd-lara, calling `generate_structs` is for some reason modifying **all** generated files, in parts that are seemingly unrelated to this PR. See example below.

image

I wanted to confirm if this is expected or if there's something wrong.

claytonpbarrows:
which version of IS are you using? I suspect that the issue is IS has been updated to enable all the planned changes for PSY1.0, but we haven't merged those changes to PSY#master yet. If you start from the `pre-10` branch it will probably make more sense.

raphaelsaavedra:
Thanks @claytonpbarrows, that was exactly the issue. I've rebased this PR to `pre-10`. This should be fine now.

jd-lara:
I looked into the ISO's documentation and it seems the solution based on an enum for the status aligns better with the data. We would keep available to essentially remove the Generator from the data set but use status to still create constraints and variables but provide a richer set of options that eventually can be extended. If everyone agrees, we can close this PR an open a new one with the enum.

raphaelsaavedra:
That does make sense to me, what is confusing is that `status` is supposed to denote the initial commitment status of the unit. By changing it to allow for `must_run`, it seems we have two conflicting features in the same field?

1. An initial condition of `must_run` doesn't make a lot of sense, how is it different from `on` in that case?
2. `must_run` is a parameter that in theory could be forecasted, as units can be set to `must_run` in a certain hour but not in other hours. On the other hand, `status` is a decision variable in a unit commitment problem. It's not clear to me how these two would not conflict.

andrewrosemberg:
I don't see the conflicting part.

1. A status initial condition of must_run indicates the same thing as the original status.
2. The forecaster of status would now comprise the must run, so no loss here.

raphaelsaavedra:
It’s not clear to me how a forecast of `status` behaves in a unit commitment. You’re adding a forecast to the very thing you’re trying to decide. For example, suppose a unit should be set to `must_run` on hours 1 through 12, but not in the rest. So you add a forecast for status that has value `must_run` for 1:12… and what for the rest? Would we have a possible value in the `Enum` that is intended to not interfere in the decisions?

Conversely, it's clearer if we have a field `must_run` just like `ThermalMultiStart` does. `status` keeps its role as initial condition and something you don't add forecasts to, and `must_run` is a parameter that can be forecasted to set the appropriate `must_run` flag in each hour.

jd-lara:
@raphaelsaavedra the idea is to force a status and assign the constraint accordingly. For instance one could implement the constraint `must_run_param <= on_status <= offline_param` and make the bounds ParameterJuMP parameters.

raphaelsaavedra:
> For instance one could implement the constraint must_run_param[t] <= on_status[t] <=(1 - offline_param[t]) and make the bounds ParameterJuMP parameters.

That makes complete sense to me, but my question was more on how to input the data rather than how to formulate the constraints.

It's clear to me that the idea is to be able to add stuff such as `must_run` and `out` in `status`. However, I personally don't see the advantages of doing this over having a field for `must_run` such as in `ThermalMultiStart` (note that there already is a field for `availability`, so `out` can already be forecasted there). But by all means, this is just my view as an outsider and I'm sure you guys have a better understanding of the framework, it's just that this solution seems more confusing to me (as per my example in the previous comment)

jd-lara:
@claytonpbarrows what's your take on this one?

sourabhdalvi:
Sorry, that I'm bit late to comment on this, but you can still use ThermalMultiStart with just one type of start and have the `must-run` constraint. So, while parsing you just populate the hot start values and leave others at zero and set `start_types`  to 1.

jd-lara:
@raphaelsaavedra I hope this clarifies the way to use MultiStart in a similar way as ThermalStandard.

PR #620, opened 2020-10-02
Fix validation of time series parameters

daniel-thom:


src/base.jl
```diff
@@ -963,48 +948,44 @@ function get_time_series_values(component::Component, time_series::TimeSeriesDat
...
+end
+
 """
-Return sorted time series initial times.
+Return the initial times for all forecasts.
 """
-#function get_time_series_initial_times(sys::System)
-#    return IS.get_time_series_initial_times(sys.data)
-#end
+generate_forecast_initial_times(sys::System) = IS.generate_forecast_initial_times(sys.data)
```
jd-lara, line 958: should we call this generate still?

src/base.jl
```diff
@@ -963,48 +948,44 @@ function get_time_series_values(component::Component, time_series::TimeSeriesDat
...
+"""
+get_forecast_interval(sys::System) = IS.get_forecast_interval(sys.data)
 
 """
 Return the resolution for all time series.
 """
-function get_time_series_resolution(sys::System)
-    return IS.get_time_series_resolution(sys.data)
-end
+get_time_series_resolution(sys::System) = IS.get_time_series_resolution(sys.data)
```
jd-lara, line 988: This will close #619

src/base.jl
```diff
@@ -963,48 +948,44 @@ function get_time_series_values(component::Component, time_series::TimeSeriesDat
...
+get_forecast_window_count(sys::System) = IS.get_forecast_window_count(sys.data)
 
 """
-Return the earliest initial_time for a time series.
+Return the horizon for all forecasts.
 """
-function get_time_series_initial_time(sys::System)
-    return IS.get_time_series_initial_time(sys.data)
-end
+get_forecast_horizon(sys::System) = IS.get_forecast_horizon(sys.data)
```
jd-lara, line 973: What's the return in the cases where the is not stored forecasts?

daniel-thom: 0

jd-lara (approved):

PR #625, opened 2020-10-15
Adding methods to support VariableCost TimeSeries

sourabhdalvi:


src/descriptors/power_system_structs.json
```diff
@@ -101,9 +101,9 @@
         {
           "name": "ancillary_services",
           "comment": "Bids for the ancillary services",
-          "null_value": "IdDict{Any,Float64}()",
-          "data_type": "IdDict{Any,Float64}",
-          "default": "IdDict{Any,Float64}()"
+          "null_value": "IdDict{String,IS.TimeSeriesMetadata}()",
```
jd-lara, line 104: If you use a String here, no need to use an `IdDict` these special dicts are useful if you use a service as a key

src/models/cost_function_timeseries.jl
```diff
@@ -0,0 +1,83 @@
...
+end
+
+function get_variable_cost(
+    device::Generator,
+    cost::MarketBidCost;
+    start_time::Union{Nothing, Dates.DateTime} = nothing,
+    len::Union{Nothing, Int} = nothing,
+)
+    time_series_key = get_variable(cost)
+    ts_contanier = IS.get_time_series_container(device)
```
jd-lara, line 13: I don't think you are using this container.

src/models/cost_function_timeseries.jl
```diff
@@ -0,0 +1,83 @@
...
+    return
+end
+
+function set_service_bid!(
+    sys::System,
+    component::Generator,
+    service::Service,
+    time_series_data::IS.TimeSeriesData,
+)
+    @assert get_name(time_series_data) == get_name(service)
```
jd-lara, line 76: It is better if you throw an error here.

src/models/cost_function_timeseries.jl
```diff
@@ -0,0 +1,83 @@
...
+)
+    add_time_series!(sys, component, time_series_data)
+    metadata_type = IS.time_series_data_to_metadata(typeof(time_series_data))
+    key = IS.TimeSeriesKey(metadata_type, get_name(time_series_data))
+    market_bid_cost = get_operation_cost(component)
+    set_variable!(market_bid_cost, key)
+    return
+end
+
+function set_service_bid!(
```
jd-lara, line 70: since you have the component and the service can also check if the device is contributing to the service.

src/models/cost_function_timeseries.jl
```diff
@@ -0,0 +1,83 @@
...
+        get_name(service),
+        start_time = start_time,
+        len = len,
+    )
+    return TimeSeries.TimeArray(time_stamps, map(VariableCost, raw_data))
+end
+
+function set_variable_cost!(
+    sys::System,
+    component::Generator,
```
jd-lara, line 59: These functions should be on StaticInjectionDevice since Storage can participate and also loads.

src/models/generated/ReserveDemandCurve.jl
```diff
@@ -18,7 +17,6 @@ Data Structure for a operating reserve with demand curve product for system simu
 - `name::String`
 - `available::Bool`
 - `time_frame::Float64`: the relative saturation time_frame, validation range: `(0, nothing)`, action if invalid: `error`
-- `operation_cost::TwoPartCost`: Cost for providing reserves  [`TwoPartCost`](@ref)
```
jd-lara, line 21: should this have a variable cost like the market bid and have the same functions add/set/ etc ?

jd-lara (changes requested):
Overall looks good but there are a couple of missing things. 

1. Dispatch on InjectionDevice the market bid cost. 
2. Add the same functions/structure to the MarketReserveCurve as with VariableCost in MarketBidCost.

src/models/generated/ReserveDemandCurve.jl
sourabhdalvi: Yup it should

src/descriptors/power_system_structs.json
```diff
@@ -70,10 +70,10 @@
       "docstring": "Data Structure Operational Cost to reflect market bids of energy and ancilliary services.\nCompatible with most US Market bidding mechanisms",
       "fields": [
         {
-          "name": "variable",
-          "comment": "variable cost representing the energy bid",
-          "null_value": "VariableCost((0.0, 0.0))",
-          "data_type": "VariableCost"
+            "name": "variable",
+            "comment": "no load cost",
```
jd-lara, line 74: This is an incorrect comment for this item

src/descriptors/power_system_structs.json
```diff
@@ -2885,6 +2885,12 @@
       "docstring": "Data Structure for a operating reserve with demand curve product for system simulations.",
       "parametric": "ReserveDirection",
       "fields": [
+        {
+            "name": "variable",
```
jd-lara, line 2889: same here

src/models/cost_function_timeseries.jl
```diff
@@ -0,0 +1,112 @@
+function get_variable_cost(
+    ts::IS.TimeSeriesData,
+    ::Union{MarketBidCost, ReserveDemandCurve},
```
jd-lara, line 3: @daniel-thom should we use union here or separate the code?

src/models/cost_function_timeseries.jl
```diff
@@ -0,0 +1,112 @@
...
+)
+    data = TimeSeries.TimeArray(TimeSeries.timestamp(ts), map(VariableCost, ts.data))
+    return data
+end
+
+function get_variable_cost(
```
jd-lara, line 9: We need to add something here that catches the case the key in the `MarketBidCost` is nothing

jd-lara (changes requested):
The changes look good. Tests are still needed.

src/models/cost_function_timeseries.jl
daniel-thom: This feels like unnecessary specificity to me.  Are we using this to dispatch differently in other cases?

src/models/cost_function_timeseries.jl
```diff
@@ -0,0 +1,112 @@
...
+    ts::IS.TimeSeriesData,
+    ::Union{MarketBidCost, ReserveDemandCurve},
+)
+    data = TimeSeries.TimeArray(TimeSeries.timestamp(ts), map(VariableCost, ts.data))
```
daniel-thom, line 5: This looks non-functional to me.
1. TimeSeries.timestamp(ts) is not valid.
2. Not all subtypes of TimeSeriesData have the field `data`.
3. The scaling_factor_multiplier is not accounted for.  Even if these time series are never scaling factors I think we should strive to use the same interface.
4. Is this intended to return all forecast windows or just one?

  I _think_ you want this instead:
```
data = map(VariableCost, get_time_series_array(component, ts, start_time))
```
component and start_time are not passed in.  You could pick the first initial time.

src/models/cost_function_timeseries.jl
```diff
@@ -0,0 +1,112 @@
...
+
+function get_services_bid(
+    device::StaticInjection,
+    cost::MarketBidCost,
+    service::Service;
+    start_time::Union{Nothing, Dates.DateTime} = nothing,
+    len::Union{Nothing, Int} = nothing,
+)
+    services_dict = get_ancillary_services(cost)
+    time_series_type = services_dict[get_name(service)]
```
daniel-thom, line 41: `ancillary_services` stores instances of `IS.TimeSeriesMetadata`, so this doesn't look correct.  Is that intentional?

src/models/cost_function_timeseries.jl
```diff
@@ -0,0 +1,112 @@
...
+end
+
+function verify_device_eligibility(
+    sys::System,
+    component::StaticInjection,
+    service::Service,
+)
+    contributing_devices = get_contributing_devices(sys, service)
+    if !in(get_name(component), contributing_devices)
+        error("Device $(get_name(component)) isn't eligibility to contribute to service $(get_name(service)).")
```
daniel-thom, line 109:
```suggestion
        error("Device $(get_name(component)) isn't eligible to contribute to service $(get_name(service)).")
```

src/descriptors/power_system_structs.json
```diff
@@ -70,10 +70,10 @@
       "docstring": "Data Structure Operational Cost to reflect market bids of energy and ancilliary services.\nCompatible with most US Market bidding mechanisms",
       "fields": [
         {
-          "name": "variable",
-          "comment": "variable cost representing the energy bid",
-          "null_value": "VariableCost((0.0, 0.0))",
-          "data_type": "VariableCost"
+            "name": "variable",
```
jd-lara, line 73: Why the whitespace noise here?

test/test_cost_functions.jl
```diff
@@ -11,3 +11,56 @@ res = [1.0, 0.1, 0.1]
 for (ix, v) in enumerate(bps)
     @test isapprox(v, res[ix])
 end
+
+@testset "Test MarketBidCost with Polynomial Cost Timeseries with Service Forecast " begin
```
jd-lara, line 15: We need a test that serializes and de-serializes a system with MarketBidCost. It should just work but it is better to test it.

jd-lara (changes requested):
The PR looks pretty good, we need more tests for serialization and also tests for the ORDC reserve.

test/test_cost_functions.jl
sourabhdalvi: I'll add that test

src/models/cost_function_timeseries.jl
```diff
@@ -0,0 +1,108 @@
...
+    cost::OperationalCost;
+    start_time::Union{Nothing, Dates.DateTime} = nothing,
+    len::Union{Nothing, Int} = nothing,
+)
+    time_series_key = get_variable(cost)
+    if isnothing(time_series_key)
+        error("Cost component has a `nothing` stored in field `variable`, Please use `set_variable_cost!` to add variable cost forecast.")
+    end
+    raw_data = IS.get_time_series(
+        IS.time_series_metadata_to_data(time_series_key.time_series_type),
```
daniel-thom, line 26: The innards of TimeSeriesKey are getting spread out quite a bit.  Here are some possible improvements:
- Create a TimeSeriesKey constructor that accepts a TimeSeriesData.
- Create a get_time_series method that takes a TimeSeriesKey instead of type/name.

What do you think?  @sourabhdalvi @jd-lara

jd-lara: I think it's better to have a get time series method that takes a time series key. The only con to that is that it might be exposed in the repl

daniel-thom: I wouldn't be opposed to `IS.get_time_series_internal`.

jd-lara: I am fine with that and opens the question for later if should add `_internal` for private API.

daniel-thom: It could also be `get_time_series_by_key`.  In Julia that feels redundant because TimeSeriesKey is a function argument.  But that is a common pattern I've seen elsewhere.

jd-lara: I think `IS.get_time_series_by_key` is the best approach. Even with the repetition, it is better.

src/models/cost_function_timeseries.jl
```diff
@@ -0,0 +1,125 @@
...
+    cost::MarketBidCost,
+    service::Service;
+    start_time::Union{Nothing, Dates.DateTime} = nothing,
+    len::Union{Nothing, Int} = nothing,
+)
+    time_series_key = get_variable(cost)
+    raw_data = IS.get_time_series_by_key(
+        time_series_key,
+        device,
+        get_name(service),
```
daniel-thom, line 67: Is this name different from the name inside `time_series_key`?

sourabhdalvi: yes, we decide to use the same time_series_key from variable cost field assuming that the type of forecast for service cost would be the same as  variable cost.

test/test_cost_functions.jl
```diff
@@ -11,3 +11,80 @@ res = [1.0, 0.1, 0.1]
...
+    name = "test"
+    horizon = 24
+    service_data = Dict(initial_time => ones(horizon), other_time => ones(horizon))
+    polynomial_cost = repeat([(999.0, 1.0)], 24)
+    data_polynomial =
+        SortedDict(initial_time => polynomial_cost, other_time => polynomial_cost)
+    sys = create_rts_system()
+    generators = collect(get_components(ThermalStandard, sys))
+    generator = get_component(ThermalStandard, sys, get_name(generators[1]))
+    market_bid = MarketBidCost(nothing)
```
daniel-thom, line 28: Could you allow
```
market_bid = MarketBidCost()
```
?  It feels odd to me to make the user pass `nothing`.

test/test_cost_functions.jl
```diff
@@ -11,3 +11,80 @@ res = [1.0, 0.1, 0.1]
...
+    horizon = 24
+    polynomial_cost = repeat([(999.0, 1.0)], 24)
+    data_polynomial =
+        SortedDict(initial_time => polynomial_cost, other_time => polynomial_cost)
+    pwl_cost = repeat([repeat([(999.0, 1.0)], 5)], 24)
+    data_pwl = SortedDict(initial_time => pwl_cost, other_time => pwl_cost)
+    for d in [data_polynomial, data_pwl]
+        @testset "Add deterministic from $(typeof(d)) to ReserveDemandCurve variable cost" begin
+            sys = System(100)
+            reserve = ReserveDemandCurve{ReserveUp}(nothing)
```
daniel-thom, line 82: Same point here

test/test_cost_functions.jl
sourabhdalvi: This is only for the test, I don't expect for the user to call this constructor. I saw the null constructor being used in other tests as well, so followed that pattern. This doesn't make much sense as an example as the user's would create the generator with the MarketBidCost but here I use the generator from RTS system and replace the ThreePartCost with MarketBidCost but I'm open to populate it with some values .

daniel-thom (approved):

jd-lara (approved):

PR #627, opened 2020-10-21
Time series docs

daniel-thom:
This includes Jose's additions.


jd-lara (approved):
Given that I already looked into it. Will approve.

docs/src/modeler/time_series.md
```diff
@@ -0,0 +1,283 @@
...
+not actually free up file space (HDF5 behavior). If you want to remove all or
+most time series instances then consider using `clear_time_series!`. It
+will delete the HDF5 file and create a new one. PowerSystems has plans to
+automate this type of workflow.
+
+```julia
+    clear_time_series!(sys)
+```
+
+## Data exploration
```
daniel-thom, line 132: Jose moved this section above the other retrieval sections.  I think I disagree with that move.  I think we should reinforce `get_time_series_array` and `ForecastCache` as the methods to get data.  

1. They account for scaling factors.
2. They will not load all data into memory.

jd-lara: I agree with Dan's assessment that we should reinforce  `get_time_series_array` and `ForecastCache`. The reason I moved it is that the first thing people usually want to do after loading data is looking at it.

docs/src/modeler/time_series.md
```diff
@@ -0,0 +1,293 @@
...
+values (scaling factors). For instance it the maximum active power returns the
+value 50.0 and the scaling factor at some time point is 0.65, the forecast
+value will correspond to 32.5.
+
+## Adding time series to the System
+Adding time series data to a system requires a component that is already
+attached to the system. Extending the example above:
+
+```julia
+    add_time_series!(sys, component, forecast)
```
claytonpbarrows, line 115: It seems like most of the example code blocks will run if you copy and paste them in the REPL, IF you've created `component`. Does it make sense to do that so that someone could follow along?

docs/src/modeler/time_series.md
claytonpbarrows: I'm moving this back with a few minor rewording. I think that the original order doesn't prohibit inspecting it... if they really want to see the  data as it's stored, they'll find the next section.

claytonpbarrows (commented):
Here are a couple comments and a few minor changes. Please review before merging.

claytonpbarrows (approved):

docs/src/modeler/time_series.md
```diff
@@ -0,0 +1,319 @@
...
+- Create from TimeSeries.TimeArray or DataFrames.DataFrame
+
+```julia
+    resolution = Dates.Hour(1)
+    dates = range(DateTime("2020-01-01T00:00:00"), step = resolution, length = 24)
+    data = TimeArray(dates, ones(24))
+    ts = SingleTimeSeries("max_active_power", data)
+```
+
+```julia
```
raphaelsaavedra, line 59: I am familiar with the `TimeArray` approach, but I was not with this one. It's nice to know that there is an alternative that doesn't necessarily depend on `TimeSeries.jl`.

However, I'm not sure the example is clear to me. The example above defines a `TimeArray` from `2020-01-01T00:00:00` to `2020-01-01T23:00:00` and assigns `ones(24)` to it, which means for each hour the forecast is `1.0`.

In the example below, we have `ones(24)` associated to each of the two hours defined, which seems very different. I think maybe for clarity it would be better if both examples are perfectly equivalent? I apologize if they are and I'm missing something.

claytonpbarrows: @raphaelsaavedra you are correct, they aren't equivalent. The first one creates a `SingleTimeSeries` object, and the second one creates a `Deterministic` object with two 24hr forecast windows. I agree that it's confusing to create two different types of objects here. I think the point was to demonstrate that both types of objects could be created.
We should either add a description of what's going on, align the two examples, or provide four examples to enumerate all the options.

nickrobinson251: i think this is something we've found a little confusing, so think just giving all 4 examples might be worth it.

jd-lara: We are also introducing a completely new time series formatting so there will be additional interfaces.

raphaelsaavedra: Thanks for the answer!

> the second one creates a Deterministic object with two 24hr forecast windows.

That's the root of my confusion (I'm just trying to understand this second approach). In the `TimeArray` approach, we use a `TimeArray` that explicitly specifies the hours to be considered in the forecast (via `range`) so that each hour is associated to a float (the range `2020-01-01T00:00:00:1 hour:2020-01-01T23:00:00` is associated with `ones(24)` in this example).

Conversely, in the example below, the `DateTime` `range` is never specified:

```
resolution = Dates.Hour(1)
data = Dict(
    DateTime("2020-01-01T00:00:00") => ones(24),
    DateTime("2020-01-01T01:00:00") => ones(24),
)
forecast = Deterministic("max_active_power", data, resolution)
```

What we have two `DateTime`s, each pointing to a forecast of `ones(24)`. Is this supposed to indicate "a forecast (with hourly `resolution`) starting from `DateTime("2020-01-01T00:00:00")` and with values `ones(24)`? That doesn't seem the case because that would mean the second forecast starts one hour after the first one.

Well, I guess in summary I think the example could use some further clarification, because it's not obvious to me what's happening.

jd-lara: @raphaelsaavedra the content in 42eda80 should help clarify this questions.

jd-lara (approved):

docs/src/modeler/time_series.md
```diff
@@ -0,0 +1,332 @@
...
+```
+
+## Creating Time Series Data
+
+PowerSystems supports several methods to load time series data (Forecasts or
+StaticTimeSeries) into a System.
+
+- Automated parsing during system construction, this method loads the data from CSV files. Refer to the [parsing documentation](parsing.md).
+
+- Create directly from data arranges in `TimeSeries.TimeArray` or `DataFrames.DataFrame`
```
daniel-thom, line 53:
```suggestion
- Create directly from data in `TimeSeries.TimeArray` or `DataFrames.DataFrame`
```

docs/src/modeler/time_series.md
daniel-thom: Yes.  We could have a section at the top that creates an example system with multiple components.

docs/src/user_guide/quick_start_guide.md
```diff
@@ -1,3 +1,59 @@
 # Quick Start Guide
 
 This section is still under development. Please visit [PowerSystems.jl Intro](https://nbviewer.jupyter.org/github/NREL-SIIP/SIIPExamples.jl/blob/master/notebook/PowerSystems_examples/PowerSystems_intro.ipynb) in the mean time.
+
+## Definitions to work with Time Series Data
```
daniel-thom, line 5: This is good content. Would it make more sense to move it to the other page?  To me a quick start guide is something that shows you how to create a system and get running.

jd-lara: I am torn about that, I feel this is a very basic part. I say we leave it here for now and in the PR for docs, we iron out the details

daniel-thom: I'll state my main comment in a different way.  I think a quick start guide should have a list of steps to get running, not paragraphs of text and definitions.

Also, the table doesn't render in GitHub Markdown.
https://github.com/NREL-SIIP/PowerSystems.jl/blob/1912306df97edea6cfd902d4dd79ee9bd36d7240/docs/src/user_guide/quick_start_guide.md

docs/src/modeler/time_series.md
```diff
@@ -0,0 +1,332 @@
+# Managing Time Series Data
+
+**Note**: This section assumes the reader is familiar with the definitions in the
+quick start guide
```
daniel-thom, line 4: If the definitions remain in the quick start guide then this should be defined as a link (relative paths work) so that the user can click on it.

jd-lara: Yes, I will add those to the right section when I get the docs PR. I have it noted

docs/src/modeler/time_series.md
raphaelsaavedra: > the key is the initial time of the forecast and the value
field is the forecast value

Thanks! That answers my question. My last question is: doesn't the fact that they start in different instants go against the time series validation?

> All forecasts must have identical parameters:  initial timestamp, horizon, interval, look-ahead window count.

jd-lara: What do you mean? "doesn't the fact that they start in different instants go against the time series validation?"

raphaelsaavedra: Maybe I am still misunderstanding the code. There are two time series in that forecast, one is `10.0*ones(24)` starting at `2020-01-01T00:00:00` and the other is `5.0*ones(24)` starting at `2020-01-01T01:00:00`. They start at different instants, so wouldn't that cause an error when validating the time series?

daniel-thom: I would state this a bit differently.  In the code a `forecast` is a collection of look-ahead windows.  Each forecast has one initial timestamp.  Each window in the forecast will have an initial time that increments by `interval`.  Each window is an entry in the dict above.

All `forecasts` in the system must have the same initial timestamp and other parameters.

raphaelsaavedra: Thanks! I think I understand it now :)

claytonpbarrows (approved):

PR #628, opened 2020-10-21
hybridv2

jd-lara:
Updated for Hybrid System into PSY 1.0 and minimal tests

### DIFF
--- a/test/test_hybrid_system.jl
+++ b/test/test_hybrid_system.jl
@@ -9,5 +9,22 @@ c_sys14() = System(
     branches14(nodes_14),
 )
 
-    gen = ThermalStandard()
+test_sys = c_sys14()
+
+    h_sys = HybridSystem(
+        name = "Test H",
+        available = true,
+        status = true,
+        bus = get_component(Bus, test_sys, "Bus 1"),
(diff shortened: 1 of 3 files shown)


test/test_hybrid_system.jl
```diff
@@ -1,13 +1,30 @@
...
+        active_power = 1.0,
+        reactive_power = 1.0,
+        thermal_unit = ThermalStandard(nothing),
+        electric_load = PowerLoad(nothing),
+        storage = GenericBattery(nothing),
+        renewable_unit = RenewableDispatch(nothing),
+        base_power = 100.0,
+        operation_cost = TwoPartCost(nothing),
+    )
+    add_component!(test_sys, h_sys)
```
daniel-thom, line 28: Do you want a custom `check_component_addition` method for this type?  You could ensure that the composed devices are not attached to the system.

jd-lara: We will have to add that too as we develop more. I prefer to have a simple addition following the default behavior and add restrictions as we progress.

daniel-thom (approved):
Looks good

PR #630, opened 2020-10-21
service addition to device

jd-lara:
Closes PSIMA-275

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -336,19 +336,7 @@ function add_component!(
     add_component!(sys, dyn_injector; static_injector = static_injector, kwargs...)
 end
 
-"""
-Similar to [`add_component!`](@ref) but for services.
-
-# Arguments
-- `sys::System`: system
-- `service::Service`: service to add
-- `contributing_devices`: Must be an iterable of type Device
-"""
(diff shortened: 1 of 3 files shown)


src/base.jl
```diff
@@ -366,6 +366,23 @@ function add_service!(sys::System, service::Service, contributing_devices; kwarg
...
+
+# Arguments
+- `device::Device`: Device
+- `service::Service`: Service
+- `sys::System`: system
+"""
+function add_service!(device::Device, service::Service, sys::System)
+    if sys.runchecks && !validate_struct(sys, service)
+        throw(InvalidValue("Invalid value for $service"))
+    end
```
daniel-thom, line 381: This check occurred when the service was added to the system.

src/base.jl
```diff
@@ -366,6 +366,23 @@ function add_service!(sys::System, service::Service, contributing_devices; kwarg
...
+    throw_if_not_attached(device, sys)
```
daniel-thom, line 382:
```suggestion
    throw_if_not_attached(device, sys)
    throw_if_not_attached(service, sys)
```

daniel-thom (approved):

PR #631, opened 2020-10-22
Fix tests

daniel-thom:


jd-lara (approved):
The docstrings look much better.

PR #632, opened 2020-10-22
Fix linting errors

daniel-thom:


src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -180,10 +180,10 @@ Builds a Ybus from the system. The return is a Ybus Array indexed with the bus n
 
 function Ybus(sys::System; check_connectivity::Bool = true)
     branches = get_components(ACBranch, sys)
-    nodes = sort(collect(get_components(Bus, sys)), by = x -> x.number)
+    nodes = sort!(collect(get_components(Bus, sys)), by = x -> x.number)
```
jd-lara, line 183: Why a bang here? This seems odd since we are returning the object back.

src/utils/power_flow.jl
```diff
@@ -126,7 +126,7 @@ Updates system voltages and powers with power flow results
 """
 function _write_pf_sol!(sys::System, nl_result)
     result = round.(nl_result.zero; digits = 7)
-    buses = enumerate(sort(collect(get_components(Bus, sys)), by = x -> get_number(x)))
+    buses = enumerate(sort!(collect(get_components(Bus, sys)), by = x -> get_number(x)))
```
daniel-thom, line 129: The linter did not actually suggest this change.  In general, I think that we should sort in place after collect rather than allocate a new array.   I understand that some of these changes are in performance-sensitive areas and if you prefer that I not make them, I can remove.

jd-lara: I think it makes sense to do this but then it will require collecting first and then running `sort!()`.

jd-lara (approved):

PR #634, opened 2020-10-23
Remove wrong service costs

andrewrosemberg:
Removes the wrong service costs added mainly [here](https://github.com/NREL-SIIP/PowerSystems.jl/pull/587). Correct approach was implemented in the [cost function update PR](https://github.com/NREL-SIIP/PowerSystems.jl/pull/596).


jd-lara (approved):
Thanks for the fix.

PR #635, opened 2020-10-23
Add a Symmetric ReserveDirection

andrewrosemberg:
Adds a symmetric `ReserveDirection`.


jd-lara (changes requested):
We need a test for this before being able to merge.

andrewrosemberg:
Tests added @jd-lara, only nightly is failing.

jd-lara (approved):

PR #643, opened 2020-10-30
Adding missing args to transform_single_time_series

sourabhdalvi:


src/base.jl
```diff
@@ -1017,8 +1017,13 @@ end
 """
 Transform all instances of SingleTimeSeries to DeterministicSingleTimeSeries.
 """
-function transform_single_time_series!(sys::System)
-    IS.transform_single_time_series!(sys.data)
+function transform_single_time_series!(sys::System, horizon::Int, interval::Dates.Period)
```
daniel-thom, line 1020: Thanks for fixing this oversight.  However, I think that we should follow the IS function signature and make the user pass in DeterministicTimeSeries.  This would be useful if we ever support a different kind of transformation.  However, if this is the only possible transformation we'll ever make, this is good.  Opinion @jd-lara ?

src/base.jl
```diff
@@ -1017,8 +1017,13 @@ end
...
+    IS.transform_single_time_series!(
+        sys.data,
+        IS.DeterministicSingleTimeSeries,
```
daniel-thom, line 1023: I forgot to import/re-export DeterministicSingleTimeSeries.  Would you mind making that change in this PR?  Just follow the example of Deterministic in src/PowerSystems.jl.

src/base.jl
jd-lara: This is probably the only transformation we will have to do. This is ok by me.

daniel-thom: @sourabhdalvi I think that Jose was saying that what you had originally was better...no type.

docs/src/user_guide/time_series.md
```diff
@@ -350,7 +350,7 @@ Here's an example:
     add_time_series!(sys, component, ts)
 
     # Transform it to Deterministic
-    transform_single_time_series!(sys)
+    transform_single_time_series!(sys, DeterministicSingleTimeSeries, 24, Hour(24))
```
daniel-thom, line 353: @jd-lara note this change.  It will trigger a merge conflict in your docs-galore PR that you could easily miss.

daniel-thom (commented):
Looks good, thanks again for fixing this.

daniel-thom (approved):

docs/src/user_guide/time_series.md
daniel-thom: Amidst our churn you left the type in this function call.
```suggestion
    transform_single_time_series!(sys, 24, Hour(24))
```

daniel-thom (approved):

PR #644, opened 2020-11-02
adding pu cconversion for inflow

claytonpbarrows:
Changes that need to propagate to 1.0:
 - p.u. conversion of inflow


claytonpbarrows:
This looks like it's already included

PR #651, opened 2020-11-06
Use views when concatenating PTDF matrix

raphaelsaavedra:
This PR makes two small changes.

First, it replaces `S_` with `S` as it's not necessary to have a separate variable for that.

Second, it adds `@views` which makes a big difference in performance / memory allocation when calculating the PTDF for large systems.


jd-lara (approved):
Thanks for the improvement

Issue #649, opened 2020-11-04
Inconsistency in phase shift angle units


raphaelsaavedra:
In [this line](https://github.com/NREL-SIIP/PowerSystems.jl/blob/master/src/models/supplemental_accessors.jl#L21), `α` is multiplied by `(π / 180)`, suggesting it should be in degrees. However, `α` should already be in radians according to the validation range indicated in `PhaseShiftingTransformer`:
```
α::Float64`, validation range: `(-1.571, 1.571)`, action if invalid: `warn`
```
in https://github.com/NREL-SIIP/PowerSystems.jl/blob/master/src/models/generated/PhaseShiftingTransformer.jl#L35

PR #652, opened 2020-11-06
Fix susceptance calculation

raphaelsaavedra:
From what we've seen, there are some bugs in these functions:

1. For `PhaseShiftingTransformer`, the function should return `imag(y_a)` instead of `1 /  imag(y_a)`, since `y_a` already is `1 / ...`.
2. There is a conversion to radians when `α` should already be in radians (#649) -- please confirm if this is really the case as I'm not 100% sure.
3. The first two functions are inconsistent with the third. Note that only the third one includes `im`, which means when we calculate `imag(y_a)` we actually get a minus sign in front of the expression due to the `im` in the denominator. A simple fix is to add the minus sign in the other functions for consistency.

A simple example of some of these problems:

```
julia> tr1
FBiqwdLBQB (PhaseShiftingTransformer):
   name: FBiqwdLBQB
   available: false
   active_power_flow: 0.0
   reactive_power_flow: 0.0
   arc: j28BGl4Bn7 -> 63tUEHHV7m: (Arc)
   r: 0.895
   x: 2.972
   primary_shunt: 0.0
   tap: 1.0
   α: 0.0
   rate: 6.69
   ext: Dict{String,Any}()
   forecasts: InfrastructureSystems.Forecasts: 0

julia> tr2 = TapTransformer(
       name=tr1.name, 
       available=tr1.available, 
       active_power_flow=tr1.active_power_flow, 
       reactive_power_flow=tr1.reactive_power_flow,
       arc=tr1.arc, 
       r=tr1.r, 
       x=tr1.x, 
       primary_shunt=tr1.primary_shunt, 
       tap=tr1.tap, 
       rate=tr1.rate
       ) # create identical tap transformer
FBiqwdLBQB (TapTransformer):
   name: FBiqwdLBQB
   available: false
   active_power_flow: 0.0
   reactive_power_flow: 0.0
   arc: j28BGl4Bn7 -> 63tUEHHV7m: (Arc)
   r: 0.895
   x: 2.972
   primary_shunt: 0.0
   tap: 1.0
   rate: 6.69
   services: 0-element Array{Service,1}
   ext: Dict{String,Any}()
   forecasts: InfrastructureSystems.Forecasts: 0

julia> PowerSystems.get_series_susceptance(tr1) # current version - wrong value
-3.2415238896366083

julia> get_series_susceptance(tr1) # fixed version (changed 1/imag to just imag)
-0.3084968780261265

julia> PowerSystems.get_series_susceptance(tr2) # different sign for tap transformer even though it is identical
0.3364737550471063
```

The small difference in absolute value is due to the resistance being considered only for `PhaseShiftingTransformer`.
This is also something to keep in mind - is it really the best approach to consider resistance for `PhaseShiftingTransformer` and disregard it for all other types of branch?

### DIFF
--- a/src/models/supplemental_accessors.jl
+++ b/src/models/supplemental_accessors.jl
@@ -15,7 +15,11 @@ _remove_aggregration_topology!(bus::Bus, ::Area) = bus.area = nothing
 Calculate the admittance of AC branches
 """
 get_series_susceptance(b::ACBranch) = -1 / get_x(b)
-get_series_susceptance(b::TapTransformer) = -1 / (get_x(b) * get_tap(b))
+function get_series_susceptance(b::TapTransformer)
+    y = 1 / (get_r(b) + get_x(b) * 1im)
+    y_a = y / (get_tap(b))
+    return imag(y_a)
+end
 function get_series_susceptance(b::PhaseShiftingTransformer)
     y = 1 / (get_r(b) + get_x(b) * 1im)
     y_a = y / (get_tap(b) * exp(get_α(b) * 1im))
--- a/src/models/supplemental_accessors.jl
+++ b/src/models/supplemental_accessors.jl
@@ -17,7 +17,7 @@ Calculate the admittance of AC branches
 get_series_susceptance(b::ACBranch) = imag(1 / (get_r(b) + get_x(b) * 1im))
 function get_series_susceptance(b::TapTransformer)
     y = 1 / (get_r(b) + get_x(b) * 1im)
-    y_a = y / (get_tap(b))
+    y_a = y / get_tap(b)
     return imag(y_a)
 end
 function get_series_susceptance(b::PhaseShiftingTransformer)
--- a/src/models/supplemental_accessors.jl
+++ b/src/models/supplemental_accessors.jl
@@ -14,12 +14,12 @@ _remove_aggregration_topology!(bus::Bus, ::Area) = bus.area = nothing
 """
 Calculate the admittance of AC branches
 """
-get_series_susceptance(b::ACBranch) = 1 / get_x(b)
-get_series_susceptance(b::TapTransformer) = 1 / (get_x(b) * get_tap(b))
+get_series_susceptance(b::ACBranch) = - 1 / get_x(b)
+get_series_susceptance(b::TapTransformer) = - 1 / (get_x(b) * get_tap(b))
 function get_series_susceptance(b::PhaseShiftingTransformer)
     y = 1 / (get_r(b) + get_x(b) * 1im)
-    y_a = y / (get_tap(b) * exp(get_α(b) * 1im * (π / 180)))
-    return 1 / imag(y_a)
+    y_a = y / (get_tap(b) * exp(get_α(b) * 1im))
+    return imag(y_a)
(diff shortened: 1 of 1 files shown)


jd-lara:
This pr looks good, I need to check that the conversion isn't there because we have bad parsing of the phase shifting data. We should merge this, but I need to confirm there isn't another issue with the conversion from matpower.

raphaelsaavedra:
A nice little test to ensure these stay consistent would be to have several identical branches of different types, compute their susceptance and see if the results are identical (similar to what I did in the small example I wrote). This wouldn't work at the moment due to the difference in the resistance consideration, but if this becomes consistent across at least transformers it would be a simple test to add.

src/models/supplemental_accessors.jl
```diff
@@ -14,12 +14,12 @@ _remove_aggregration_topology!(bus::Bus, ::Area) = bus.area = nothing
 """
 Calculate the admittance of AC branches
 """
-get_series_susceptance(b::ACBranch) = 1 / get_x(b)
-get_series_susceptance(b::TapTransformer) = 1 / (get_x(b) * get_tap(b))
+get_series_susceptance(b::ACBranch) = -1 / get_x(b)
+get_series_susceptance(b::TapTransformer) = -1 / (get_x(b) * get_tap(b))
```
jd-lara, line 18: This function should be 
```suggestion
function get_series_susceptance(b::TapTransformer)
    y = 1 / (get_r(b) + get_x(b) * 1im)
    y_a = y / (get_tap(b))
    return imag(y_a)
end
```

src/models/supplemental_accessors.jl
```diff
@@ -14,12 +14,12 @@ _remove_aggregration_topology!(bus::Bus, ::Area) = bus.area = nothing
 """
 Calculate the admittance of AC branches
 """
-get_series_susceptance(b::ACBranch) = 1 / get_x(b)
-get_series_susceptance(b::TapTransformer) = 1 / (get_x(b) * get_tap(b))
+get_series_susceptance(b::ACBranch) = -1 / get_x(b)
```
jd-lara, line 17: same here 
```suggestion
get_series_susceptance(b::ACBranch) = imag(1 / (get_r(b) + get_x(b) * 1im))
```

jd-lara (commented):
There seems to be a mismatch between the definition of susceptance in theory with the implementation in Matpower. See Line 62 of `makeBdc` function. 

There is also a mismatch in the calculation of the PTDF in the presence of phase shifters, refer also to `makeBdc` in matpower. I need to make more research into where the difference comes from.

src/models/supplemental_accessors.jl
jd-lara: same here

raphaelsaavedra:
Thanks, I guess this fixes the inconsistency in the consideration of resistance across those functions. I'll create a small test on this next.

raphaelsaavedra:
Seems like some PTDF tests are failing after the latest changes.

jd-lara:
> Seems like some PTDF tests are failing after the latest changes.

Yeah, that's because we are no longer following the same convention as Matpower.

PR #653, opened 2020-11-09
add fix to PTDF pst calculation

### LINKED ISSUE #649: Inconsistency in phase shift angle units

jd-lara:
closes #652 closes #649 @raphaelsaavedra after reviewing several references it seems that this is the best way to go ignoring the angle of the PST since the intended purpose of the PST is to change power flows and if the angle is included the PTDF becomes dependent on the system state. This will be consistent with Matpower and PowerWorld.

### DIFF
--- a/src/models/supplemental_accessors.jl
+++ b/src/models/supplemental_accessors.jl
@@ -15,11 +15,17 @@ _remove_aggregration_topology!(bus::Bus, ::Area) = bus.area = nothing
 Calculate the admittance of AC branches
 """
 get_series_susceptance(b::ACBranch) = 1 / get_x(b)
-get_series_susceptance(b::TapTransformer) = 1 / (get_x(b) * get_tap(b))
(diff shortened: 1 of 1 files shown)


raphaelsaavedra:
Sure, that looks fine to me

claytonpbarrows (approved):
LGTM

PR #654, opened 2020-11-10
Add configure_logging function

daniel-thom:


docs/src/api/public.md
```diff
@@ -131,12 +131,8 @@ Filter = t -> t ∉ [System]
...
-Order = [:type, :function]
-Filter = t -> t ∈ [InfrastructureSystems.configure_logging,
-                   InfrastructureSystems.open_file_logger,
-                   InfrastructureSystems.MultiLogger,
-                   InfrastructureSystems.LogEventTracker,
-                   ]
+Modules = [PowerSystems]
+Pages = ["utils/logging.jl"]
+Public = true
+Private = false
```
daniel-thom, line 137: Is this correct? It doesn't seem to build correctly on my local system.

jd-lara: Yes, it should be ok. I will check in the CI if it is failing

jd-lara (approved):
LGTM, thanks for the fix

PR #655, opened 2020-11-10
Add test coverage

daniel-thom:


src/base.jl
```diff
@@ -944,10 +944,6 @@ function make_time_series(
 end
 =#
 
-function get_time_series_names(::Type{T}, component::Component) where {T <: TimeSeriesData}
```
jd-lara, line 947: This is a useful one to restore in the future.

daniel-thom: It still exists in IS. If we want it then we need to re-export it in PSY.

jd-lara: Yes, please.

jd-lara:
@daniel-thom great, this PR restores the Coverage levels.

jd-lara (approved):

jd-lara:
@daniel-thom do you mind removing nightly from https://github.com/NREL-SIIP/PowerSystems.jl/blob/master/.github/workflows/pr_testing.yml also?

daniel-thom:
> 
> 
> @daniel-thom great, this PR restores the Coverage levels.

I'm not sure about restoration. This code was not tested previously. I don't know what caused the coverage drop.

PR #663, opened 2020-11-13
update branch naming

jd-lara:
Before fixing the broken tests, I would like to know if this PR is acceptable. The Branch names are currently based on the order that we parse the power models dict. As a result, the names change between runs sometimes and it makes hard to make code based on branch naming.

### DIFF
--- a/src/parsers/power_models_data.jl
+++ b/src/parsers/power_models_data.jl
@@ -79,7 +79,6 @@ function _get_pm_branch_name(device_dict, bus_f, bus_t)
     return string("$(get_name(bus_f)) - $(get_name(bus_t)) - $index")
 end
 
-
 """
 Creates a PowerSystems.Bus from a PowerSystems bus dictionary
 """
--- a/src/parsers/power_models_data.jl
+++ b/src/parsers/power_models_data.jl
@@ -72,6 +72,15 @@ function _get_pm_dict_name(device_dict)
 end
 
 """
+Internal branch name retreval from pm2ps_dict
+"""
+function _get_pm_branch_name(device_dict, bus_f, bus_t)
+    index = get(device_dict, "name", device_dict["index"])
+    return string("$(get_name(bus_f)) - $(get_name(bus_t)) - $index")
+end
+
+
+"""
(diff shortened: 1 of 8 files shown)


src/parsers/power_models_data.jl
```diff
@@ -583,13 +592,13 @@ function read_branch!(sys::System, data, bus_number_to_bus::Dict{Int, Bus}; kwar
         return
     end
 
-    _get_name = get(kwargs, :branch_name_formatter, _get_pm_dict_name)
+    _get_name = get(kwargs, :branch_name_formatter, _get_pm_branch_name)
```
daniel-thom, line 595: 1. It seems problematic if we are inferring information from the name. We should just need them to be unique. There's probably a detail I'm missing.
2. This same function is used in many other places. Notably, `make_dcline`. Is this a problem anywhere else?

jd-lara: Re 1. It is common to use names to refer to lines. In a large system, operators use names and numbers, for instance, see the print screen with real line names. 

image

Re 2. true, DC Line needs the same change.

claytonpbarrows (approved):
Generally, I think this is an improvement. Previously, we were inferring an identifier on branch circuits based on the order that we were encountering them, right? I vaguely recall adding the iteration index to them to ensure uniqueness, but this is better.

src/parsers/power_models_data.jl
```diff
@@ -71,6 +71,14 @@ function _get_pm_dict_name(device_dict)
     return get(device_dict, "name", string(device_dict["index"]))
 end
 
+"""
+Internal branch name retreval from pm2ps_dict
+"""
+function _get_pm_branch_name(device_dict, bus_f, bus_t)
+    index = get(device_dict, "name", device_dict["index"])
+    return string("$(get_name(bus_f))-$(get_name(bus_t))-i_$index")
```
daniel-thom, line 79: You can remove the call to `string`

jd-lara: thanks

daniel-thom (approved):

PR #680, opened 2020-12-17
Cb/parser updates

claytonpbarrows:
Several parser updates to address PSIMA-252 and PSIMA-237
Requires https://github.com/NREL-SIIP/PowerSystemsTestData/pull/44 which will break tests in PSY#master.
Also requires https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/185


src/parsers/power_system_table_data.jl
```diff
@@ -790,6 +814,40 @@ function calculate_variable_cost(data::PowerSystemTableData, gen, cost_colnames,
...
+    data::PowerSystemTableData,
+    gen,
+    cost_colnames::_CostPointColumns,
+    base_power,
+)
+    vom = isnothing(gen.variable_cost) ? 0.0 : gen.variable_cost
+
+    var_cost = [(getfield(gen, c), getfield(gen, mw)) for (c, mw) in cost_colnames.columns]
+    var_cost = unique([
+        (tryparse(Float64, string(c[1])), tryparse(Float64, string(c[2])))
```
daniel-thom, line 827: What will happen if `tryparse` returns `nothing`? Should this be `parse` instead? Might want to reconsider other uses of `tryparse` in this file.

claytonpbarrows: `tryparse` returns a `nothing` if it fails while `parse` returns an error. In this case, there are sometimes columns with `"NA"` or some other non numeric value that we want to just ignore. So, for this case, I think it achieves the desired outcome. But I'll double check the other usages.

daniel-thom (approved):

PR #682, opened 2020-12-24
Updating Printing 

sourabhdalvi:
- Adding show method for PowerSystems Components that uses the getter function instead of get field
- Other change here is in `_get_multiplier` when unit settings is nothing  was to return 1.0 i.e DEVICE_BASE, now this is changed to NATURAL_UNITS.
- Changes to Test to reflect the changes in `_get_multiplier`

Requires IS#187 to be merged and tagged for Documentation to pass

### DIFF
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -371,6 +371,8 @@ import InfrastructureSystems:
     get_time_series_names,
     get_next_time_series_array!,
     get_next_time,
+    get_units_info,
+    set_units_info!,
     to_json,
     from_json,
     serialize,
--- a/src/base.jl
+++ b/src/base.jl
@@ -271,8 +271,11 @@ Clear any value stored in ext.
 """
 clear_ext!(sys::System) = IS.clear_ext!(sys.internal)
 
-function set_unit_system!(component::Component, settings::SystemUnitsSettings)
-    component.internal.units_info = settings
+function set_units_setting!(
+    component::Component,
+    settings::Union{SystemUnitsSettings, Nothing},
+)
+    set_units_info!(get_internal(component), settings)
     return
 end
 
@@ -296,7 +299,7 @@ function get_units_base(system::System)
 end
 
 function get_units_setting(component::T) where {T <: Component}
-    return component.internal.units_info
+    return get_units_info(get_internal(component))
 end
 
 function has_units_setting(component::T) where {T <: Component}
@@ -330,7 +333,7 @@ function add_component!(
     skip_validation = false,
     kwargs...,
 ) where {T <: Component}
-    set_unit_system!(component, sys.units_settings)
+    set_units_setting!(component, sys.units_settings)
     @assert has_units_setting(component)
 
     check_attached_buses(sys, component)
@@ -395,7 +398,7 @@ function _add_service!(
         throw_if_not_attached(device, sys)
     end
 
-    set_unit_system!(service, sys.units_settings)
+    set_units_setting!(service, sys.units_settings)
     # Since this isn't atomic, order is important. Add to system before adding to devices.
     IS.add_component!(sys.data, service; skip_validation = skip_validation, kwargs...)
 
@@ -462,7 +465,7 @@ function add_service!(
         throw_if_not_attached(_service, sys)
     end
 
-    set_unit_system!(service, sys.units_settings)
+    set_units_setting!(service, sys.units_settings)
     IS.add_component!(sys.data, service; skip_validation = skip_validation, kwargs...)
 end
 
@@ -496,7 +499,7 @@ function add_service!(
     skip_validation = _validate_or_skip(sys, service, skip_validation)
     set_contributing_services!(sys, service, contributing_services)
 
-    set_unit_system!(service, sys.units_settings)
+    set_units_setting!(service, sys.units_settings)
     IS.add_component!(sys.data, service; skip_validation = skip_validation, kwargs...)
 end
 
@@ -1367,7 +1370,7 @@ function handle_component_addition!(sys::System, component::RegulationDevice; kw
         # This will not be true during deserialization, and so won't run then.
         remove_component!(sys, component.device)
         # The line above removed the component setting so needs to be added back
-        set_unit_system!(component.device, component.internal.units_info)
+        set_units_setting!(component.device, component.internal.units_info)
     end
     return
 end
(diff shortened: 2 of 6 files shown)


src/utils/print.jl
```diff
@@ -40,3 +40,81 @@ function Base.show(io::IO, ::MIME"text/plain", data::PowerSystemTableData)
...
+                print(io, "\n   SystemUnitSetting not defined, default to NATURAL_UNITS")
+            else
+                print(io, "\n   ")
+                show(io, MIME"text/html"(), obj.units_info)
+            end
+            continue
+        elseif obj isa IS.TimeSeriesContainer || obj isa InfrastructureSystemsType
+            val = summary(getfield(ist, name))
+        elseif obj isa Vector{<:InfrastructureSystemsComponent}
+            val = summary(getfield(ist, name))
```
claytonpbarrows, line 109: can you consolidate these two conditions?

claytonpbarrows (commented):
This looks good.
Can you also add one (somewhat related) fix? Can you add an `uppercase()` to [this line](https://github.com/NREL-SIIP/PowerSystems.jl/blob/master/src/base.jl#L281) to make the set units call a bit more flexible?

src/utils/print.jl
```diff
@@ -40,3 +40,100 @@ function Base.show(io::IO, ::MIME"text/plain", data::PowerSystemTableData)
...
+    end
+    print(io, ")")
+end
+
+function Base.show(io::IO, ::MIME"text/plain", ist::Component)
+    default_units = false
+    if isnothing(ist.internal.units_info)
+        print(io, "\n")
+        @warn("SystemUnitSetting not defined, using NATURAL_UNITS for dispalying device specification.")
+        ist.internal.units_info =
```
daniel-thom, line 70: Can you use `set_unit_system!` and `get_units_setting` instead of dot accessors? I'm not sure why those names aren't symmetric. @jd-lara ? Those functions also use dot accessors, which implicitly require a field named `internal`. They could call `get_internal` instead.

sourabhdalvi: I think that is better way of doing this, I'll also look into updating those function

daniel-thom (commented):
I think this is a simple fix to the problem and am OK with it. However, there is a bit of code repetition that could be problematic in the future. It doesn't have to be now but perhaps @jd-lara could fix the cross-package design issue with the system settings and serialization. Then the original solution could be used here.

jd-lara:
@daniel-thom We can open an issue to address this. I can take a look into the units settings problem next week.

src/utils/print.jl
```diff
@@ -40,3 +40,108 @@ function Base.show(io::IO, ::MIME"text/plain", data::PowerSystemTableData)
...
+            print(io, "\n   ", name, ": ", val)
+        end
+    finally
+        if default_units
+            set_units_setting!(ist, nothing)
+        end
+    end
+end
+
+function Base.show(io::IO, ::MIME"text/html", ist::Component)
```
daniel-thom, line 107: Do we need this function? It doesn't appear to have any customization for html, so shouldn't the one above be sufficient?

src/utils/print.jl
```diff
@@ -40,3 +40,108 @@ function Base.show(io::IO, ::MIME"text/plain", data::PowerSystemTableData)
...
+            print(io, ", ")
+        end
+        print(io, val)
+    end
+    print(io, ")")
+end
+
+function Base.show(io::IO, ::MIME"text/plain", ist::Component)
+    default_units = false
+    if has_units_setting(ist)
```
daniel-thom, line 67: Shouldn't this be `!has_units_setting`?

src/utils/print.jl
sourabhdalvi: I wasn't sure about this to implemented both function, but I did do some test and I think I can do without this.

src/base.jl
```diff
@@ -271,16 +271,19 @@ Clear any value stored in ext.
 """
 clear_ext!(sys::System) = IS.clear_ext!(sys.internal)
 
-function set_unit_system!(component::Component, settings::SystemUnitsSettings)
-    component.internal.units_info = settings
+function set_units_setting!(
```
daniel-thom, line 274: @jd-lara Are you OK with the name change? If not, Sourabh could add a decrecated `set_unit_system! = set_units_setting!` function.

jd-lara: This could be problematic since it will define the system unit system. I don't object but this is awkward

daniel-thom: I added a comment that these function names were unnecessarily not symmetric.
```
function set_unit_system!(component::Component, settings::SystemUnitsSettings)
function get_units_setting(component::T) where {T <: Component}
function has_units_setting(component::T) where {T <: Component}
```
Sourabh is making the names consistent and not changing any functionality. To your point about changing the system setting: that may or may not be true. In the case where it is true, we shouldn't allow it. That could be dangerous. That should not be exported to users. I'm really not liking this whole solution...

src/utils/print.jl
```diff
@@ -40,3 +40,66 @@ function Base.show(io::IO, ::MIME"text/plain", data::PowerSystemTableData)
...
+    end
+    print(io, ")")
+end
+
+function Base.show(io::IO, ::MIME"text/plain", ist::Component)
+    default_units = false
+    if !has_units_setting(ist)
+        print(io, "\n")
+        @warn(
+            "SystemUnitSetting not defined, using NATURAL_UNITS for dispalying device specification."
```
daniel-thom, line 70: typo: "displaying"

src/PowerSystems.jl
```diff
@@ -371,6 +371,8 @@ import InfrastructureSystems:
     get_time_series_names,
     get_next_time_series_array!,
     get_next_time,
+    get_units_info,
+    set_units_info!,
```
daniel-thom, line 375: I'm not sure that we should export these. What are you thinking here?

sourabhdalvi: well my intent wasn't to export them but just use them in PSY, maybe I should just use `IS. get_units_info` in the code if this approach exposes these function to the user.

daniel-thom: Oh sorry, I misread. For some reason I thought these were exported in the other PR. Please ignore.

daniel-thom (approved):

jd-lara (approved):

PR #683, opened 2021-01-04
Adding changes from JuliaFormatter update

sourabhdalvi:


jd-lara (approved):
I will approve because the tests are passing but the changes are massive.

PR #684, opened 2021-01-06
Updating unit settings methods 

sourabhdalvi:
- The previous method used dot operator to set the units settings, now it uses `set_units_info!` and `get_units_info` from IS. 
- Bug fix in hybrid system unit settings methods, pre vous method tried to set units for each one of the device fields even the ones that are nothing. New method uses `get_component` method that return valid components of a HybridSystem.

### DIFF
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -371,6 +371,8 @@ import InfrastructureSystems:
     get_time_series_names,
     get_next_time_series_array!,
     get_next_time,
+    get_units_info,
+    set_units_info!,
     to_json,
     from_json,
     serialize,
--- a/src/models/HybridSystem.jl
+++ b/src/models/HybridSystem.jl
@@ -120,7 +120,8 @@ end
 IS.get_name(value::HybridSystem) = value.name
 
 function get_component(value::HybridSystem)
-    components = [value.thermal_unit, value.electric_load, value.storage, value.renewable_unit]
+    components =
+        [value.thermal_unit, value.electric_load, value.storage, value.renewable_unit]
     filter!(x -> !isnothing(x), components)
     return components
 end
--- a/src/models/HybridSystem.jl
+++ b/src/models/HybridSystem.jl
@@ -119,7 +119,7 @@ end
 
 IS.get_name(value::HybridSystem) = value.name
 
-function get_component(value::HybridSystem)
+function _get_component(value::HybridSystem)
     components =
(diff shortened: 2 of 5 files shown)


src/PowerSystems.jl
```diff
@@ -371,6 +371,8 @@ import InfrastructureSystems:
     get_time_series_names,
     get_next_time_series_array!,
     get_next_time,
+    get_units_info,
+    set_units_info!,
```
daniel-thom, line 375: Calling `set_units_info!` is potentially dangerous because of our implementation. What's your thinking behind exporting these to users? Do they need to be able to call them?

src/models/HybridSystem.jl
```diff
@@ -118,12 +118,19 @@ function HybridSystem(::Nothing)
...
-    value.electric_load.internal.units_info = settings
-    value.storage.internal.units_info = settings
-    value.renewable_unit.internal.units_info = settings
+
+function get_component(value::HybridSystem)
+    components =
+        [value.thermal_unit, value.electric_load, value.storage, value.renewable_unit]
+    filter!(x -> !isnothing(x), components)
+    return components
+end
```
daniel-thom, line 127: Did you intend to include this function?

jd-lara: should this be `get_components` ?

jd-lara:
We should not export a method to change the units settings at the device level since it will change the system's settings. Only the system based level unit setting change should be exported.

src/models/HybridSystem.jl
jd-lara: @daniel-thom we shouldn't export this one, we have specific ones for each component.

src/models/HybridSystem.jl
```diff
@@ -118,12 +118,19 @@ function HybridSystem(::Nothing)
...
 
 IS.get_name(value::HybridSystem) = value.name
-function set_unit_system!(value::HybridSystem, settings::SystemUnitsSettings)
-    value.internal.units_info = settings
-    value.thermal_unit.internal.units_info = settings
-    value.electric_load.internal.units_info = settings
-    value.storage.internal.units_info = settings
-    value.renewable_unit.internal.units_info = settings
+
+function get_component(value::HybridSystem)
```
jd-lara, line 122:
```suggestion
function _get_components(value::HybridSystem)
```

jd-lara (approved):

Issue #687, opened 2021-01-14
Supporting multiple assets with the same name


egbutter:
I attempted to parse a PSSE raw file, and the parser raised an error that a bus with that name already exists. The RAW file does have the same name for multiple buses, but this is expected. System files sometimes do repeat the same name for multiple buses tied to the same asset, and PSSE allows this. Rather than raising a hard error, it would be more useful to include a flag to append the bus number / uuid. 

```
sys =  System(PowerModelsData(joinpath(home_dir, "XXXX.raw")));
ArgumentError: XXXX     is already stored for type Bus
```

claytonpbarrows:
You should be able to do:
```
sys = System(joinpath(home_dir, "XXXX.raw"), bus_name_formatter = x -> string(x["name"]) * "-" * string(x["index"]))
```

Sorry, this feature is not well documented.

jd-lara:
@claytonpbarrows and @daniel-thom can we add a better error message about this?

egbutter:
@claytonpbarrows thanks for the tip. This solved the immediate problem. I'm still not able to parse this file due to other issues I will need to investigate myself, I think.

jd-lara:
closed via #690

PR #690, opened 2021-01-15
adding better error messages for pti parsing

claytonpbarrows:
- addresses #687 by indicating resolutions in error messages.
 - requires https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/191

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -664,9 +664,6 @@ end
 """
 Check to see if the component of type T with name exists. If T is an abstract
 type then the names of components across all subtypes of T must be unique.
-
-Throws ArgumentError if T is not a concrete type and there is more than one component with
-    requested name
 """
 function has_component(::Type{T}, sys::System, name::AbstractString) where {T <: Component}
     return IS.has_component(T, sys.data.components, name)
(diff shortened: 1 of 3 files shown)


src/base.jl
```diff
@@ -661,6 +661,17 @@ function remove_component!(
     handle_component_removal!(sys, component)
 end
 
+"""
+Check to see if the component of type T with name exists. If T is an abstract
+type then the names of components across all subtypes of T must be unique.
+
+Throws ArgumentError if T is not a concrete type and there is more than one component with
+    requested name
```
daniel-thom, line 669: This is not really an error, so I think it would be better to return true in this case.

src/base.jl
```diff
@@ -661,6 +661,14 @@ function remove_component!(
     handle_component_removal!(sys, component)
 end
 
+"""
+Check to see if the component of type T with name exists. If T is an abstract
+type then the names of components across all subtypes of T must be unique.
```
daniel-thom, line 666: You can delete `If T is an abstract type then the names of components across all subtypes of T must be unique.`, right? We do allow duplicate names across subtypes of an abstract type. There is a limitation is in `get_component`. That function will throw an exception if you specify an abstract type with a name that is duplicate. It will work if you specify the concrete type.

This is something that will likely confuse users.

daniel-thom (approved):

PR #692, opened 2021-01-16
Replace all enums with new scoped_enum macro

daniel-thom:
This does two things:
1. Adopt the new `scoped_enum` from InfrastructureSystems (https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/192/files)
2. Replace any regular enums with scoped_enum

I tested deserialization of a System created with the old version and it worked.

I do have a concern that this will cause an API breakage for users. It depends on whether they actually use something like `BusTypes.BusType`, which is hopefully unlikely.

### DIFF
--- a/src/definitions.jl
+++ b/src/definitions.jl
@@ -4,80 +4,70 @@ const From_To_Float = NamedTuple{(:from, :to), Tuple{Float64, Float64}}
 const FromTo_ToFrom_Float = NamedTuple{(:from_to, :to_from), Tuple{Float64, Float64}}
 
 "From http://www.pserc.cornell.edu/matpower/MATPOWER-manual.pdf Table B-4"
-IS.@scoped_enum GeneratorCostModel begin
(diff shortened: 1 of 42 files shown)


jd-lara (approved):
I will label the change as breaking since it does imply an API change. This isn't too bad, but we need to make clear the possible breaking changes.

jd-lara:
> This does two things:
> 
> 1. Adopt the new `scoped_enum` from InfrastructureSystems (https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/192/files)
> 2. Replace any regular enums with scoped_enum
> 
> I tested deserialization of a System created with the old version and it worked.
> 
> I do have a concern that this will cause an API breakage for users. It depends on whether they actually use something like `BusTypes.BusType`, which is hopefully unlikely.

So far, this PR does break some of my code that creates systems but it isn't too bad. However, it does imply an API change.

PR #696, opened 2021-01-21
minor pm parser_updates

claytonpbarrows:
this PR passes kwargs through to parse_file to allow for control on pm_validation routines

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -18,7 +18,8 @@ const SYSTEM_KWARGS = Set((
     :time_series_read_only,
     :timeseries_metadata_file,
     :unit_system,
-    :validate,
+    :pm_data_corrections,
+    :import_all,
 ))
 
 # This will be used in the future to handle serialization changes.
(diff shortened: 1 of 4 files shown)


src/base.jl
```diff
@@ -18,6 +18,7 @@ const SYSTEM_KWARGS = Set((
     :time_series_read_only,
     :timeseries_metadata_file,
     :unit_system,
+    :validate,
```
jd-lara, line 21: How does a user distinguish between validate and run checks? If I set run checks to false and validate to true was to that mean?

jd-lara (changes requested):
the validate keyword needs changing

src/base.jl
jd-lara: We also need to update the docstring

claytonpbarrows: fair... I was just directly passing the pm kwarg... I'll make it more clear

jd-lara: FWIW I don't think that validate is correct in PM either since it doesn't validate really but mutates the underlying data.

jd-lara (approved):

jd-lara (approved):

PR #701, opened 2021-01-30
adding option to avoid data validation checks when deserializing system

claytonpbarrows:

PR #703, opened 2021-02-02
Simplify system checks

daniel-thom:
1. Makes `System.runchecks` the only way to control validation of components. Deprecates the use of `skip_validation` in `add_component!`.
2. Adds functions to check components. Exports those functions along with `check!`.
3. Fixes a bug where the kwarg runchecks was not honored during deserialization.

Dependent on this InfrastructureSystems PR: https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/194.

### DIFF
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -308,11 +308,17 @@ export get_reserve_limit_dn
 export get_participation_factor
 export get_cost
 export get_units_base
+export get_runchecks
+export set_runchecks!
+export check!
+export check_component
+export check_components
 
 export configure_logging
 export open_file_logger
 export MultiLogger
 export LogEventTracker
+export UnitSystem
 
 #################################################################################
 # Imports
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -261,7 +261,7 @@ export set_service_bid!
 export iterate_windows
 export get_window
 export transform_single_time_series!
-export correct_component!
+export sanitize_component!
 export validate_component
 export validate_component_with_system
 
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -261,6 +261,9 @@ export set_service_bid!
 export iterate_windows
 export get_window
 export transform_single_time_series!
+export correct_component!
+export validate_component
+export validate_component_with_system
 
 #export make_time_series
 export get_bus_numbers
@@ -310,7 +313,7 @@ export get_cost
 export get_units_base
 export get_runchecks
 export set_runchecks!
-export check!
+export check
 export check_component
 export check_components
 
@@ -390,6 +393,7 @@ import InfrastructureSystems:
     UnitSystem,
     SystemUnitsSettings,
     open_file_logger,
+    validate_struct,
     MultiLogger,
     LogEventTracker
 
--- a/src/base.jl
+++ b/src/base.jl
@@ -120,7 +120,7 @@ function System(base_power::Float64, buses::Vector{Bus}, components...; kwargs..
     end
 
     if get(kwargs, :runchecks, true)
-        check!(sys)
+        check(sys)
     end
 
     return sys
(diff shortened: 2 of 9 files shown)


src/PowerSystems.jl
```diff
@@ -308,11 +308,17 @@ export get_reserve_limit_dn
...
+export set_runchecks!
+export check!
+export check_component
+export check_components
 
 export configure_logging
 export open_file_logger
 export MultiLogger
 export LogEventTracker
+export UnitSystem
```
daniel-thom, line 321: @jd-lara I think this may fix https://github.com/NREL-SIIP/InfrastructureSystems.jl/issues/193

src/base.jl
```diff
@@ -1089,12 +1118,38 @@ function validate_struct(sys::System, value::IS.InfrastructureSystemsType)
     return true
 end
 
+"""
+Check system consistency and validity.
```
daniel-thom, line 1122: @claytonpbarrows @jd-lara This function no longer mutates the system. I think that we should remove the `!` in this PR. Do you have any intention of this mutating any attribute in the future? Since this PR exports the functions it will make users wonder what it may change.

src/base.jl
```diff
@@ -1114,7 +1169,7 @@ function IS.deserialize(
     ::Type{System},
     filename::AbstractString;
     time_series_read_only = false,
-    runchecks = true,
+    kwargs...,
```
daniel-thom, line 1172: I changed this to `kwargs...` so that I could tell if the user passed it in or not.

src/base.jl
claytonpbarrows: I think it's ok to remove the `!` here. I think it's better to have separation between the checks and the corrections. Which I suppose answers your second question... I agree that we should change this to `correct`. This would start a newly distinct set of data corrections functionality.

claytonpbarrows (approved):
Thanks!

src/base.jl
daniel-thom: @claytonpbarrows `correct_component!` or `sanitize_component`?

jd-lara:
@daniel-thom does this PR closes #701 ?

jd-lara (approved):

src/PowerSystems.jl
jd-lara: I will test it to see.

src/base.jl
claytonpbarrows: lets go with `sanitize_component!`

claytonpbarrows: do we also need to change `check_angle_limits!` to `sanitize_angle_limits!`?

daniel-thom:
@claytonpbarrows @jd-lara You can see the minor additions I made to the docs here: https://github.com/daniel-thom/PowerSystems.jl/blob/check-components/docs/src/model_developer_guide/adding_custom_types.md#custom-validation.

You can judge whether we need a comprehensive one-pager on component checks (probably).

claytonpbarrows:
looks like there is a type in that doc `compoment` appears a couple times.

claytonpbarrows:
Custom Validation

You can implement three methods to perform custom validation or correction for your type. PowerSystems calls all of these functions in add_component!.

  `sanitize_component!(component::Component, sys::System)`: intended to make standard data corrections (e.g. voltage angle in degrees -> radians)
  `validate_component(component::Component)`: intended to check component field values for internal consistency
   `validate_component_with_system(component::Component, sys::System)`: intended to check component field values for consistency with system

daniel-thom:
> Maybe some small descriptions for the usage in this first round before we writeup the full doc page:
> 
> Custom Validation
> 
> You can implement three methods to perform custom validation or correction for your type. PowerSystems calls all of these functions in add_component!.
> 
> `sanitize_component!(component::Component, sys::System)`: intended to make standard data corrections (e.g. voltage angle in degrees -> radians)
> `validate_component(component::Component)`: intended to check component field values for internal consistency
> `validate_component_with_system(component::Component, sys::System)`: intended to check component field values for consistency with system

Fixed

claytonpbarrows (approved):

jd-lara (approved):

PR #704, opened 2021-02-04
Outage parsing Updates for Tabular Data Parser

scdhulipala:


src/descriptors/power_system_inputs.json
```diff
@@ -259,6 +259,18 @@
         }
     ],
     "generator": [
+        {
+            "unit": "%",
+            "name": "fotr",
+            "description": "Forced Outage Rate",
+            "default_value": 0.0
```
claytonpbarrows, line 266: I think this should be null

src/descriptors/power_system_inputs.json
```diff
@@ -259,6 +259,18 @@
...
+        },
+        {
+            "unit": "hours",
+            "name": "mttr",
+            "description": "Mean Time to Recover (in Hours)",
+            "default_value": 100
```
claytonpbarrows, line 272: I think this should be null

src/parsers/power_system_table_data.jl
```diff
@@ -969,6 +998,10 @@ function make_thermal_generator(data::PowerSystemTableData, gen, cost_colnames,
         time_limits = timelimits,
         operation_cost = op_cost,
         base_power = base_power,
+        ext = Dict(
```
claytonpbarrows, line 1001: This will always add outage info to `ext`. I think we want to only add it if it exists and the columns are specified in the configuration.yaml file. So, I'd suggest making another function like `add_outage_info!(component, gen)` that adds the outage info to the component after it's created if the `(gen.fotr, gen.mttr) != (nothing,nothing)`

claytonpbarrows (changes requested):
this looks pretty good. one suggested change to make it so that we aren't adding meaningless outage info when we don't intend to.

src/descriptors/power_system_inputs.json
scdhulipala: Yeah, I made this change in the new commit.

src/parsers/power_system_table_data.jl
scdhulipala: Yeahm this makes sense. The new commit includes these changes.

claytonpbarrows (commented):
I think this generally looks fine. I'm still a bit hesitant to merge it. My hesitation is because this will be the first time that we enable parsing data into `ext` fields. I think that this is OK in this case because (1) we don't rely upon it in any models (yet), (2) this implementation shouldn't hurt anything, and (3) it's difficult (I think?) to extend the parser to add this info without making modifications like these. However, this does raise the question of weather or not we should attempt to make this type of activity more easily extended. Thoughts @daniel-thom ?

daniel-thom:
> I think this generally looks fine. I'm still a bit hesitant to merge it. My hesitation is because this will be the first time that we enable parsing data into `ext` fields. I think that this is OK in this case because (1) we don't rely upon it in any models (yet), (2) this implementation shouldn't hurt anything, and (3) it's difficult (I think?) to extend the parser to add this info without making modifications like these. However, this does raise the question of weather or not we should attempt to make this type of activity more easily extended. Thoughts @daniel-thom ?

First, I would question whether `ext` is the best solution here. Should these fields be defined in the `ThermalStandard` struct? There could be a `Union{Nothing, OutageInfo}`. You could also create a new struct. That struct could contain a `ThermalStandard` and employ method forwarding.

src/parsers/power_system_table_data.jl
```diff
@@ -936,6 +958,17 @@ function make_reactive_params(
     return reactive_power, reactive_power_limits
 end
 
+function add_outage_info!(component::StaticInjection, gen)
+    outage_rates = outage_to_rate((gen.fotr, gen.mttr))
+    outage_probability = outage_info(outage_rates.λ, outage_rates.μ)
+
+    component.ext = Dict(
```
daniel-thom, line 965: Using `ext` in this way could be very problematic because it will discard any existing data. Instead, you should do this:
```
ext = get_ext(component)
for fn in fieldnames(outage_info)
    ext[string(fn)] = getfield(outage_probability, fn)
end
```

src/parsers/power_system_table_data.jl
```diff
@@ -936,6 +958,17 @@ function make_reactive_params(
...
+        fn => getfield(outage_probability, Symbol.(fn)) for
+        fn in string.(fieldnames(outage_info))
+    )
+    return component
```
daniel-thom, line 969: You shouldn't need to return component.

src/parsers/power_system_table_data.jl
scdhulipala: Yeah, makes sense. The new commit includes these changes.

claytonpbarrows:
Closing in favor of solution in extension package

PR #706, opened 2021-02-09
DNMY: Refactoring Storage 

sourabhdalvi:


src/descriptors/power_system_structs.json
```diff
@@ -65,6 +65,69 @@
       ],
       "supertype": "OperationalCost"
     },
+    {
+        "struct_name": "StorageManagementCost",
```
jd-lara, line 69: Not sure if StorageManagementCost is the best descriptor here.

src/models/generated/BatteryEMS.jl
```diff
@@ -17,9 +17,8 @@ This file is auto-generated. Do not edit.
         reactive_power::Float64
         reactive_power_limits::Union{Nothing, Min_Max}
         base_power::Float64
+        operation_cost::Union{Nothing, OperationalCost}
```
jd-lara, line 20: this will be a breaking API change. we need to add a constructor that makes the conversion correctly

jd-lara (commented):
This PR makes a lot of API changes in the Storage objects, it will force a 2.0 release. We need to evaluate how to better address these changes.

src/deprecated.jl
```diff
@@ -0,0 +1,24 @@
+
+# BEGIN 1.2.1  deprecations 
+@deprecate get_energy_value(val::BatteryEMS) get_energy_surplus_cost(
```
jd-lara, line 3: We need a test for this. Please add a new test file for deprecations. `test_deprecations.jl`

sourabhdalvi: added the test file

jd-lara (approved):
The PR looks good to me but we can't merge until we fix the breaking change for PSI. We need to wait to merge.

src/descriptors/power_system_structs.json
jd-lara: @claytonpbarrows what do you think about this name?

claytonpbarrows: I don't hate `StorageManagementCost`. But what about `StorageOperationalCost`?

PR #707, opened 2021-02-18
Cb/bigdata bugfixes

claytonpbarrows:
requires https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/195


src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -186,3 +162,106 @@ function Ybus(sys::System; check_connectivity::Bool = true)
...
+    lookup::L
+end
+
+"""
+Builds a Adj from a collection of buses and branches. The return is a Adj Array indexed with the bus numbers and the branch names.
+
+# Keyword arguments
+- `check_connectivity::Bool`: Checks connectivity of the network using Goderya's algorithm
+"""
+function Adj(branches, nodes; check_connectivity::Bool = true)
```
jd-lara, line 183: we should abstract the code in the PTDF matrix to do the same and have it as an internal method and avoid code duplication.

jd-lara (changes requested):

src/utils/network_calculations/ybus_calculations.jl
claytonpbarrows: Do you mean replace [this code](https://github.com/NREL-SIIP/PowerSystems.jl/blob/master/src/utils/network_calculations/ptdf_calculations.jl#L34-L51) with `Adj(branches, nodes)`? If so, they are not quite the same thing. The PTDF needs an incidence matrix (N x B), and `Adj` creates an adjacency matrix (N x N).

claytonpbarrows: @jd-lara, I made an attempt at minimizing code duplication here. My best effort netted a 4 line reduction in code. So, I'm inclined to leave it as is for readability/simplicity.

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -186,3 +162,106 @@ function Ybus(sys::System; check_connectivity::Bool = true)
...
+- `check_connectivity::Bool`: Checks connectivity of the network using Goderya's algorithm
+"""
+function Adjacency(branches, nodes; check_connectivity::Bool = true)
+    buscount = length(nodes)
+    bus_ax = get_number.(nodes)
+    axes = (bus_ax, bus_ax)
+    bus_lookup = _make_ax_ref(bus_ax)
+    look_up = (bus_lookup, bus_lookup)
+    a = SparseArrays.spzeros(Int, buscount, buscount)
+    for b in branches
```
jd-lara, line 190: This code is pretty much the same as 

https://github.com/NREL-SIIP/PowerSystems.jl/blob/d83466c5b73058d9d1f3673b7a511248edc82d59/src/utils/network_calculations/ptdf_calculations.jl#L34-L51

src/PowerSystems.jl
```diff
@@ -206,6 +207,7 @@ export get_dynamic_components
 
 export solve_powerflow!
 export solve_powerflow
+export is_connected
```
jd-lara, line 210: should we rename this has_islands?

claytonpbarrows: I think of the "island" terminology as an operational situation rather than a network connectivity issue. I prefer "connected" or "is_connected". But, I could be persuaded otherwise.

jd-lara: Is connected seems to imply singular, doesn't represent that there is a system that is continuously connected.

claytonpbarrows: maybe something more like `PM.check_connectivity()`

jd-lara: we can use validate_connectivity() since in PSY check is meant to throw an error and validate return true/false

src/PowerSystems.jl
```diff
@@ -180,6 +180,7 @@ export VariableReserveNonSpinning
 export PTDF
 export Ybus
 export LODF
+export Adjacency
```
jd-lara, line 183:
```suggestion
export AdjacencyMatrix
```

jd-lara (approved):
One minor suggestion about the name. Otherwise looks good.

src/PowerSystems.jl
claytonpbarrows: It's a `<:PowerNetworkMatrix`, so I'm inclined to keep it as `Adjacency`

PR #727, opened 2021-04-27
Add full support for HybridSystem

daniel-thom:
- Add supertype StaticInjectionSubystem.
- Add support for time series at the subcomponent level.
- Add serialization support.

This will allow construction of a HybridSystem with subcomponents that are either attached or not attached to the system. If they are attached, they are removed. If they have time series, those are moved to the HybridSystem.

All time series methods are now implemented or intentionally left out except
- ForecastCache
- StaticTimeSeriesCache

Solving that will be ugly. Let's talk @claytonpbarrows @jd-lara 

Also, I had to change the time series storage code in IS. Now this PR is dependent on the following IS PR: https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/218

### DIFF
--- a/src/models/static_injection_subsystem.jl
+++ b/src/models/static_injection_subsystem.jl
@@ -8,12 +8,13 @@ Subtypes must implement:
       subcomponent::Component,
       name::AbstractString,
   )
+- does_time_series_name_match_subcomponent(hybrid::HybridSystem, subcomponent, name)
 
 The subcomponents in subtypes are not attached to the System.
 """
 abstract type StaticInjectionSubystem <: StaticInjection end
 
-function get_time_series(
+function IS.get_time_series(
     ::Type{T},
     subsystem::StaticInjectionSubystem,
     subcomponent::InfrastructureSystemsComponent,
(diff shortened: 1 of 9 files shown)


src/models/static_injection_subsystem.jl
```diff
@@ -0,0 +1,33 @@
+"""
+Abstract type for a subsystem that contains multiple instances of StaticInjection
```
jd-lara, line 2: I think we need to clarify that this is applicable to Subsystems that compose StaticInjection. Otherwise, components that have arrays of devices could also apply to the description

test/common.jl
```diff
@@ -25,6 +25,37 @@ function create_rts_multistart_system(time_series_resolution = Dates.Hour(1))
     return System(data; time_series_resolution = time_series_resolution)
 end
 
+function create_rts_system_with_hybrid_system(; add_forecasts = true)
```
jd-lara, line 28: @sourabhdalvi should we move this to PSIB?

jd-lara (approved):
LGTM, I left a minor comment about the docstring.

jd-lara:
> - Add supertype StaticInjectionSubystem.
> - Add support for time series at the subcomponent level.
> - Add serialization support.
> 
> This will allow construction of a HybridSystem with subcomponents that are either attached or not attached to the system. If they are attached, they are removed. If they have time series, those are moved to the HybridSystem.
> 
> There are a number of time series methods implemented on component that I did not yet implement HybridSystem + subcomponent. Basically, all of them would have to have forwarding based on the time series name. I would like feedback on what is required. All?
> 
> - get_time_series_array
> - get_time_series_values
> - copy_time_series!
> - has_time_series 
> - ForecastCache
> - StaticTimeSeriesCache

All of them are needed (with probably the exception of copy), specially the cache one for simulation.

Issue #728, opened 2021-04-28
Resolve scaling of connectivity checks in the Ybus. 


claytonpbarrows:
```julia

julia> results = solve_powerflow(sys)
[ Info: Unit System changed to UnitSystem.SYSTEM_BASE = 0
ERROR: OutOfMemoryError()
Stacktrace:
  [1] Array
    @ ./boot.jl:448 [inlined]
  [2] spmatmul(A::SparseArrays.SparseMatrixCSC{Int64, Int64}, B::SparseArrays.SparseMatrixCSC{Int64, Int64})
    @ SparseArrays /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/SparseArrays/src/linalg.jl:195
  [3] *
    @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/SparseArrays/src/linalg.jl:166 [inlined]
  [4] _goderya(ybus::SparseArrays.SparseMatrixCSC{ComplexF64, Int64})
    @ PowerSystems /projects/siipspia/cbarrows/.julia/packages/PowerSystems/8Tn5A/src/utils/network_calculations/ybus_calculations.jl:232
  [5] validate_connectivity(M::SparseArrays.SparseMatrixCSC{ComplexF64, Int64}, nodes::Vector{Bus}, bus_lookup::Dict{Int64, Int64})
    @ PowerSystems /projects/siipspia/cbarrows/.julia/packages/PowerSystems/8Tn5A/src/utils/network_calculations/ybus_calculations.jl:252
  [6] Ybus(branches::InfrastructureSystems.FlattenIteratorWrapper{ACBranch}, nodes::Vector{Bus}, fixed_admittances::Vector{FixedAdmittance}; check_connectivity::Bool)
    @ PowerSystems /projects/siipspia/cbarrows/.julia/packages/PowerSystems/8Tn5A/src/utils/network_calculations/ybus_calculations.jl:147
  [7] Ybus (repeats 2 times)
    @ /projects/siipspia/cbarrows/.julia/packages/PowerSystems/8Tn5A/src/utils/network_calculations/ybus_calculations.jl:140 [inlined]
  [8] _solve_powerflow(system::System, finite_diff::Bool; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
[...]

PR #730, opened 2021-05-04
Cb/new utils

claytonpbarrows:
this PR adds a scalable network connectivity calculation to address #728  and some methods for adding and removing multiple components.


src/base.jl
```diff
@@ -402,6 +402,27 @@ function add_component!(
...
+# Examples
+```julia
+sys = System(100.0)
+
+buses = [bus1, bus2, bus3]
+generators = [gen1, gen2, gen3]
+foreach(x -> add_component!(sys, x), Iterators.flatten((buses, generators)))
+```
+"""
+function add_components!(sys::System, components)
```
claytonpbarrows, line 422: @daniel-thom should I specify the type of `components`? If so, what do you think I should use?

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -243,9 +246,37 @@ function _goderya(ybus::SparseArrays.SparseMatrixCSC)
...
+
+    preferred_method = length(buses) > 2000 ? find_connected_components : _goderya
+
+    if method === nothing
+        method = preferred_method
+    end
+
+    if method == _goderya
+        @info "Validating connectivity with Goderya algorithm"
+        length(buses) > 2000 &&
```
claytonpbarrows, line 264: `2000` is pretty arbitrary at this point. It seems reasonable, but if there is a better way to guess at what size system is going to cause issues, I'm all ears.

src/base.jl
```diff
@@ -647,6 +668,16 @@ function remove_components!(::Type{T}, sys::System) where {T <: Component}
     end
 end
 
+function remove_components!(
+    ::Type{T},
+    sys::System,
+    filter_func::Function,
+) where {T <: Component}
+    for component in get_components(T, sys, filter_func)
```
daniel-thom, line 676:
```
for component in collect(get_components(T, sys, filter_func))
```

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -212,7 +212,10 @@ Builds a Adjacency from the system. The return is an N x N Adjacency Array index
 - `check_connectivity::Bool`: Checks connectivity of the network using Goderya's algorithm
 """
 function Adjacency(sys::System; check_connectivity::Bool = true)
-    nodes = sort!(collect(get_components(Bus, sys)), by = x -> get_number(x))
+    nodes = sort(
```
daniel-thom, line 215: sort!

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -268,3 +299,37 @@ function validate_connectivity(M, nodes, bus_lookup)
...
+    neighbors = Dict(i => [] for i in buses)
+    for line in branches
+        arc = get_arc(line)
+        if get_from(arc) ∈ buses && get_to(arc) ∈ buses
+            push!(neighbors[get_from(arc)], get_to(arc))
+            push!(neighbors[get_to(arc)], get_from(arc))
+        end
+    end
+
+    component_lookup = Dict(i => Set{Bus}([i]) for i in buses)
```
daniel-thom, line 325: I'm not sure that this is valid for mutable objects. Could you use bus number as the key? `IdDict{Bus}` is valid.

src/base.jl
daniel-thom: No. This is best. There is no way to specify any type of iterable.

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -268,3 +299,37 @@ function validate_connectivity(M, nodes, bus_lookup)
...
+# this function borrows heavily from "calc_connected_compoents" in PowerModels.jl
+function find_connected_components(sys::System)
+    buses = get_components(Bus, sys, x -> get_bustype(x) != BusTypes.ISOLATED)
+    branches = get_components(Branch, sys, get_available)
+
+    return find_connected_components(branches, buses)
+end
+
+function find_connected_components(branches, buses)
+    neighbors = Dict(i => [] for i in buses)
```
daniel-thom, line 316: This will iterate over buses many times. You could do it only once.

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -268,3 +299,37 @@ function validate_connectivity(M, nodes, bus_lookup)
...
+    component_lookup = Dict(i => Set{Bus}([i]) for i in buses)
+    touched = Set{Bus}()
+
+    for i in buses
+        if !(i in touched)
+            PowerSystems._dfs(i, neighbors, component_lookup, touched)
+        end
+    end
+
+    return (Set(values(component_lookup)))
```
daniel-thom, line 334:
```
Set(values(component_lookup))
```

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -243,9 +246,37 @@ function _goderya(ybus::SparseArrays.SparseMatrixCSC)
...
+    )
+    branches = get_components(Branch, sys, get_available)
+
+    preferred_method = length(buses) > 2000 ? find_connected_components : _goderya
+
+    if method === nothing
+        method = preferred_method
+    end
+
+    if method == _goderya
```
daniel-thom, line 262: It seems a little off that you don't actually call this function. Maybe this is internal and doesn't matter, but maybe the user input should be something different, like a string or enum.

src/utils/network_calculations/ybus_calculations.jl
jd-lara: we need to do something based on the YBus matrix size not buses.

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -243,9 +246,37 @@ function _goderya(ybus::SparseArrays.SparseMatrixCSC)
     return I
 end
 
-function validate_connectivity(sys::System)
-    a = Adjacency(sys; check_connectivity = false)
-    return validate_connectivity(a.data, collect(get_components(Bus, sys)), a.lookup[1])
+function validate_connectivity(sys::System; method::Union{Nothing, Function} = nothing)
+    buses = sort(
```
jd-lara, line 250: You are calling the same code for the adjacency and the validation. Can we move it to a separate function.

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -243,9 +246,37 @@ function _goderya(ybus::SparseArrays.SparseMatrixCSC)
...
+    preferred_method = length(buses) > 2000 ? find_connected_components : _goderya
+
+    if method === nothing
+        method = preferred_method
+    end
+
+    if method == _goderya
+        @info "Validating connectivity with Goderya algorithm"
+        length(buses) > 2000 &&
+            @warn "The Goderya algorithm is memory intensive on large networks and may not scale well"
```
jd-lara, line 265: Move this warning to the `_goderya` function.

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -243,9 +246,37 @@ function _goderya(ybus::SparseArrays.SparseMatrixCSC)
...
+        a = Adjacency(branches, buses; check_connectivity = false)
```
jd-lara, line 266: adjacency recollects the buses again here.

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -268,3 +299,37 @@ function validate_connectivity(M, nodes, bus_lookup)
...
+"""
+# this function borrows heavily from "calc_connected_compoents" in PowerModels.jl
+function find_connected_components(sys::System)
+    buses = get_components(Bus, sys, x -> get_bustype(x) != BusTypes.ISOLATED)
+    branches = get_components(Branch, sys, get_available)
+
+    return find_connected_components(branches, buses)
+end
+
+function find_connected_components(branches, buses)
```
jd-lara, line 315: This code is already implemented in the PowerFlow calculations. We need to find similarities and avoid duplicates.

src/PowerSystems.jl
```diff
@@ -210,13 +210,15 @@ export get_dynamic_components
 export solve_powerflow!
 export solve_powerflow
 export validate_connectivity
+export find_connected_components
```
jd-lara, line 213: Is there any reason to export this?

claytonpbarrows: I think it's useful because it returns the buses that makeup the different components. I created it because I had developed a workflow where I parsed a dataset that had multiple connected components and I needed to figure out what they were to address the issue. If I'm the only one, I can not export, but I think others may find it useful.

src/PowerSystems.jl
```diff
@@ -212,13 +212,15 @@ export get_dynamic_components
 export solve_powerflow!
 export solve_powerflow
 export validate_connectivity
+export calc_connected_components
```
jd-lara, line 215: if we are keeping it, can we remane to `calculate` or `get`. It is preferable to have full word in exported functions.

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -243,12 +246,52 @@ function _goderya(ybus::SparseArrays.SparseMatrixCSC)
...
-    a = Adjacency(sys; check_connectivity = false)
-    return validate_connectivity(a.data, collect(get_components(Bus, sys)), a.lookup[1])
+function validate_connectivity(sys::System; method::Union{Nothing, Function} = nothing)
+    buses = sort!(
+        collect(get_components(Bus, sys, x -> get_bustype(x) != BusTypes.ISOLATED)),
+        by = b -> get_number(b),
+    )
+    branches = get_components(Branch, sys, get_available)
+
+    a = Adjacency(branches, buses; check_connectivity = false)
```
jd-lara, line 256:
```suggestion
 a = Adjancency(sys, check_connectivity = false)
```

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -212,7 +212,10 @@ Builds a Adjacency from the system. The return is an N x N Adjacency Array index
 - `check_connectivity::Bool`: Checks connectivity of the network using Goderya's algorithm
 """
 function Adjacency(sys::System; check_connectivity::Bool = true)
```
jd-lara, line 214: You might need to pass the method for connectivity too here

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -243,12 +246,52 @@ function _goderya(ybus::SparseArrays.SparseMatrixCSC)
...
+    a = Adjacency(branches, buses; check_connectivity = false)
+
+    return validate_connectivity(a.data, buses, a.lookup[1]; method = method)
+end
+
+function validate_connectivity(
+    M,
+    nodes,
+    bus_lookup;
+    method::Union{Nothing, Function} = nothing,
```
jd-lara, line 265:
```suggestion
    method::Function = preferred_method
```

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -268,3 +311,27 @@ function validate_connectivity(M, nodes, bus_lookup)
     end
     return connected
 end
+
+# this function extends the PowerModels.jl implementation to accept a System
+function calc_connected_components(sys::System)
```
jd-lara, line 316: all of these functions will have to accept kwargs... to define the connectivity check method

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -268,3 +311,27 @@ function validate_connectivity(M, nodes, bus_lookup)
...
+    a = Adjacency(sys; check_connectivity = false)
+    return calc_connected_components(a.data, a.lookup[1])
+end
+
+# this function extends the PowerModels.jl implementation to accept an adjacency matrix and bus lookup
+function calc_connected_components(
```
jd-lara, line 322: Let's make this function internal because otherwise it will be exported at the top. 

```suggestion
function _calc_connected_components(
```

jd-lara (changes requested):
We need to address the kwargs passing between the functions that conduct connectivity checks

src/PowerSystems.jl
claytonpbarrows: I left it as `calc` because it's now redefinition of the function borrowed from PowerModels

src/utils/network_calculations/ybus_calculations.jl
```diff
@@ -243,12 +246,52 @@ function _goderya(ybus::SparseArrays.SparseMatrixCSC)
...
+        end
+    end
+
+    return connected
 end
 
-function validate_connectivity(M, nodes, bus_lookup)
+function goderya_connectivity(M, nodes, bus_lookup)
+    @info "Validating connectivity with Goderya algorithm"
+    Base.summarysize(M) > 200_000 &&
```
jd-lara, line 292: We should limit to 15k buses the use of this method. If we have a full matrix with Float64 it will take 4gb of memory. Since most users would start with a commercial laptop we should limit the memory footprint to 4 gb. 

Goderya will tend to generate full matrices from the M * M operation.

src/base.jl
```diff
@@ -642,9 +663,23 @@ Remove all components of type T from the system.
...
+    for component in components
         handle_component_removal!(sys, component)
     end
+    return components
+end
+
+function remove_components!(
+    ::Type{T},
+    sys::System,
+    filter_func::Function,
```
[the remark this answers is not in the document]
jd-lara: Is this better than keeping consistency across the API?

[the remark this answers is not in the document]
claytonpbarrows: Should I just make `filter_func` an optional argument like it is in `get_components`? I was thinking that the interface would look like the `get_components` interface.

daniel-thom (approved):

PR #731, opened 2021-05-04
Allow passing single `start_up` value

nickrobinson251:
For easier use with generators that are not multi-stage

Based on a [slack conversation](https://app.slack.com/client/T018DC9RCDV/C019J02EHNU/thread/C018SP6N7SR-1620141073.002600). This allows passing a single `start_up` value to `MarketBidCost`, for easier use with generators that are not multi-stage. We still repesent start-up costs internally as a NamedTuple (with hot/warn/cold values, as in the Market Bid Forms), but allow passing a single value that sets only the `hot` value, because this is what operators use when they don’t have multiple stages.

- i ran the formatter, but let me know if there are any style comments. 
- i use `cost.start_up` in the tests because equality is not defined on `MarketBidCost` (so couldn't do `MarketBidCost(; startup=X, ...) == MarketBidCost(; startup=(hot=X, ...), ...)`)
- What is the convention for bumping versions in the PR ? Shall i add a commit to bump the version from v1.4.2 -> v1.5.0?

### DIFF
--- a/test/test_cost_functions.jl
+++ b/test/test_cost_functions.jl
@@ -66,8 +66,8 @@ end
 end
 
 @testset "Test MarketBidCost with single `start_up::Number` value" begin
-    expected = (hot=1.0, warm=0.0, cold=0.0)  # should only be used for the `hot` value.
-    cost = MarketBidCost(; start_up=1, no_load=rand(), shut_down=rand())
+    expected = (hot = 1.0, warm = 0.0, cold = 0.0)  # should only be used for the `hot` value.
+    cost = MarketBidCost(; start_up = 1, no_load = rand(), shut_down = rand())
     @test cost.start_up == expected
 end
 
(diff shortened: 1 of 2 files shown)


raphaelsaavedra:
Thanks for working on this!

test/test_cost_functions.jl
```diff
@@ -65,6 +65,12 @@ end
     @test first(TimeSeries.values(cost_forecast)).cost == first(data_pwl[initial_time])
 end
 
+@testset "Test MarketBidCost with single `start_up::Number` value" begin
+    expected = (hot = 1.0, warm = 0.0, cold = 0.0)  # should only be used for the `hot` value.
+    cost = MarketBidCost(; start_up = 1, no_load = rand(), shut_down = rand())
+    @test cost.start_up == expected
```
jd-lara, line 71: Can you use a getter function here please? We always encourage getter functions and people usually check the tests for example code.

jd-lara (changes requested):

test/test_cost_functions.jl
raphaelsaavedra: I assume you mean this:

```suggestion
    @test get_start_up(cost) == expected
```

jd-lara: yes

PR #737, opened 2021-05-08
Improve `set_service_bid!` performance by avoiding `get_contributing_devices`

raphaelsaavedra:
Closes https://github.com/NREL-SIIP/PowerSystems.jl/issues/736.

Fixes the performance issues with `set_service_bid!` due to the checks involving `get_contributing_devices`. Thanks @daniel-thom for devising the solution.


jd-lara (changes requested):
Can you please remove the semver change, we will release with other PRs as well that require a changed to 1.5

jd-lara (approved):

PR #738, opened 2021-05-11
Add support for compression at the System level

daniel-thom:


docs/src/modeler_guide/time_series.md
```diff
@@ -96,6 +96,31 @@ it in memory. Here is an example of how to do this:
 sys = System(100.0; time_series_in_memory = true)
 ```
 
+### Compression
```
jd-lara, line 99: It would be useful to cross link to the documentation in IS.

jd-lara (approved):

docs/src/modeler_guide/time_series.md
daniel-thom: Good point. I'll do it after there is a real link in the new docs.

jd-lara: https://nrel-siip.github.io/InfrastructureSystems.jl/stable/dev_guide/time_series/

PR #746, opened 2021-05-19
add conversion to power_trajectory

jd-lara:
The breaking of PowerSimulations is expected. cc. @sourabhdalvi


sourabhdalvi (approved):
LGTM

oxinabox:
This PR broke code that we are using for simulations at Invenia.
I am presently looking into why.
Could be we were accessing some internals we shouldn't have been.

PR #751, opened 2021-05-21
Bug fix in get_value

sourabhdalvi:


jd-lara (approved):
sorry to leave this bug...

PR #756, opened 2021-05-29
Add error message when calculating PTDF for a system without reference bus

raphaelsaavedra:
Currently it just throws an enigmatic `BoundsError: attempt to access 0-element Vector{Int64} at index [1]`.


claytonpbarrows (approved):
LGTM. Thanks!

PR #761, opened 2021-06-11
deprecate current control in favor of voltagemodecontrol

jd-lara:
PowerSimulationsDynamics is expected to fail the tests


rodrigomha (approved):
Looks good. We take care later on PSID.

PR #764, opened 2021-06-11
Optimize deepcopy of System

### LINKED ISSUE #762: deepcopy on a System with time_series_in_memory should use Base.deepcopy
We reimplement `Base.deepcopy(sys::System)` in order to handle the time series HDF5 file. This is not necessary if the time series data is stored in memory. We should use Base.deepcopy.

daniel-thom:
This PR fixes #762. It will also make deepcopy of a system with HDF5 time series data much faster because there will not be a serialize and deserialize.

This requires https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/242. CI will fail until a new IS tag is available. The Project.toml will have to require that new tag.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -259,14 +259,13 @@ function IS.to_json(sys::System, filename::AbstractString; force = false, runche
 end
 
 function Base.deepcopy(sys::System)
-    # We store time series in an HDF5 file that would not be copied as part of deepcopy.
-    # The HDF5 file could have data buffered in memory, so we would have to close it, make
-    # a copy, and attach it to a new system.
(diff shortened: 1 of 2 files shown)


jd-lara (approved):

jd-lara (approved):

raphaelsaavedra:
Thanks! Now that https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/242 is in, is this good to go?

jd-lara:
> Thanks! Now that [NREL-SIIP/InfrastructureSystems.jl#242](https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/242) is in, is this good to go?

I will make the release today

PR #769, opened 2021-06-16
Adding a set_value function to convert values to the correct unit system

### LINKED ISSUE #763: Implement set_value! considering the BASE_SYSTEM
getter functions return values according to the base setting but the setter functions operate based on the devices’ base
so if a unit has a base of 300 MW and the system’s base is 100 MW and you do `set_max_active_power!(unit, 0.9)` you will effectively set the power of the unit with base 300 to 0.9*300 MW this causes confusion since it is inconsistent the way that the unit settings are being handled

x-ref: https://nrel-siip.slack.com/archives/C018SP6N7SR/p1623370476055400

sourabhdalvi:
Fixes #763 

Needs IS PR#244 to be merged and tagged to pass test.


jd-lara (approved):
We can merge as soon as the PR in IS merged and tagged. You might need to rebase the branch before merging also.

PR #771, opened 2021-06-18
add convert to area power fields

jd-lara:


claytonpbarrows (approved):
@jd-lara I added the LoadZone conversion too.

PR #774, opened 2021-06-25
Compare values

daniel-thom:
This PR does two things:
- Fixes NREL-SIIP/PowerSystems.jl#604
- Adds an option to generate new UUIDs during deserialization.

The default behavior for `compare_values` is now to ignore UUIDs by default. We still use this to compare all values during serialization/deserialization.

The use case this addresses is the following:

```
sys1 = PSB.build_system(PSB.PSITestSystems, "test_RTS_GMLC_sys")
sys2 = PSB.build_system(PSB.PSITestSystems, "test_RTS_GMLC_sys")

gen1 = first(get_components(ThermalStandard, sys1))
gen2 = first(get_components(ThermalStandard, sys2))

set_active_power!(gen1, get_active_power(gen1) + 0.1)
PSY.compare_values(gen1, gen2)
┌ Error: values do not match
│   T = ThermalStandard
│   name = :active_power
│   val1 = 0.65
│   val2 = 0.55
└ @ PowerSystems ~/sandboxes/PowerSystems.jl/src/base.jl:1694
```

I decided against implementing `==` for these reasons:

- This is a diagnostic function. As such, it loops over all fields recursively to find all mismatches and then print them out. `==` will return on the first mismatch. `==` will also not print out mismatches.
- In some cases InfrastructureSystems types are immutable structs and we don't want to change the default Julia behavior. There is not a clean division in the type hierarchy between mutable and immutable structs.

### DIFF
--- a/docs/src/modeler_guide/system.md
+++ b/docs/src/modeler_guide/system.md
@@ -117,6 +117,21 @@ The serialization process stores 3 files
 2. Validation data file (`*.json` file)
 3. Time Series data file (`*.h5` file)
 
+To deserialize:
+
+```julia
+system2 = System("system.json")
+```
+
+PowerSystems generates UUIDs for the System and all components in order to have
+a way to uniquely identify objects. During deserialization it restores the same
+UUIDs.  If you will modify the System or components after deserialization then
+it is recommended that you set this flag to generate new UUIDs.
+
+```julia
+system2 = System("system.json", assign_new_uuids = true)
+```
+
 ## Reducing REPL printing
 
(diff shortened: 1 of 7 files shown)


jd-lara (approved):
LGTM, this is a great addition. if you have a chance please add a Documentation entry in https://nrel-siip.github.io/PowerSystems.jl/stable/modeler_guide/system/ mentioning this.

daniel-thom:
> LGTM, this is a great addition. if you have a chance please add a Documentation entry in https://nrel-siip.github.io/PowerSystems.jl/stable/modeler_guide/system/ mentioning this.

Good point. Fixed.

jd-lara (approved):

PR #776, opened 2021-06-26
Add renewable generic models structs

rodrigomha:
To do:
- [x] REGCA
- [x] Tests

@m-bossart please do your Pull request of REGCA to this branch


src/PowerSystems.jl
```diff
@@ -163,12 +165,15 @@ export ReactivePowerDroop
...
 export ReactivePowerPI
+export ActiveRenewableTypeAB
+export ReactiveRenewableTypeAB
 
 # InnerControl Export
 export InnerControl
 export CurrentControl
 export VoltageModeControl
 export CurrentModeControl
+export InnerREECB
```
jd-lara, line 176: We need a better name here. `REECCurrentControl`

src/PowerSystems.jl
```diff
@@ -163,12 +165,15 @@ export ReactivePowerDroop
 export ActivePowerDroop
 export ActivePowerPI
 export ReactivePowerPI
+export ActiveRenewableTypeAB
```
jd-lara, line 168:
```suggestion
export ActiveRenewablePlantControllerAB
```

src/PowerSystems.jl
```diff
@@ -139,6 +139,7 @@ export SteamTurbineGov1
 # Converter Exports
 export Converter
 export AverageConverter
+export REGCA1
```
jd-lara, line 142:
```suggestion
export NortonEquivalentConverter
```

src/PowerSystems.jl
rodrigomha: What about `ActiveRenewableControllerAB`?

src/PowerSystems.jl
rodrigomha: recall that REEC means Renewable Energy Electric Controller, so we need at least the B. We could do `RenewableEnergyCurrentControlB` or `RECurrentControlB`.

src/PowerSystems.jl
rodrigomha: Not possible since REGCA is not only the NortonEquivalentConverter. We could do maybe `RenewableEnergyConverterTypeA` or `REConverterA`, something like that?

src/PowerSystems.jl
jd-lara: Ok sounds good

src/PowerSystems.jl
jd-lara: RECurrentControlB is good

src/PowerSystems.jl
jd-lara: RenewableEnergyConverterTypeA

PR #777, opened 2021-06-28
Add regca1

m-bossart:
Adds the data structure for REGCA1 model. 
Includes a simple test for creating a power systems dynamic component.


jd-lara (changes requested):
You need to run the code autogeneration to check in the new struct.

PR #783, opened 2021-07-19
Fix Powerflow issues

### LINKED ISSUE #782: Power flow case doesn't handle storage correctly
When using PSID with battery, the battery isn't being correctly handled in the power flow

rodrigomha:
The following code closes #782 by allowing all StaticInjection to be used in PV Buses. In addition, it allows to split the power based on the `base_power` if multiple devices are connected in the same PV or REF bus.

### DIFF
--- a/src/utils/power_flow.jl
+++ b/src/utils/power_flow.jl
@@ -132,16 +132,19 @@ end
 Updates system voltages and powers with power flow results
 """
 function _write_pf_sol!(sys::System, nl_result)
+    function _is_available_source(x, bus::Bus)
+        return get_available(x) && x.bus == bus && !isa(x, ElectricLoad)
+    end
+
     result = round.(nl_result.zero; digits = 7)
     buses = enumerate(sort!(collect(get_components(Bus, sys)), by = x -> get_number(x)))
-    sys_basepower = get_base_power(sys)
 
     for (ix, bus) in buses
         if bus.bustype == BusTypes.REF
             P_gen = result[2 * ix - 1]
             Q_gen = result[2 * ix]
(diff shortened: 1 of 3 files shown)


src/utils/power_flow.jl
```diff
@@ -142,19 +142,20 @@ function _write_pf_sol!(sys::System, nl_result)
...
         elseif bus.bustype == BusTypes.PV
             Q_gen = result[2 * ix - 1]
             θ = result[2 * ix]
-            injection_components = get_components(Generator, sys, x -> get_available(x))
-            devices = [d for d in injection_components if d.bus == bus]
-            if length(devices) == 1
-                generator = devices[1]
-                gen_basepower = get_base_power(generator)
-                set_reactive_power!(generator, Q_gen)
+            injection = get_components(StaticInjection, sys, x -> get_available(x))
```
jd-lara, line 154: we can make a single call here that filters in the getter function 

```suggestion
            injection = get_components(StaticInjection, sys, x -> get_available(x) // x.bus == bus && !isa(x, ElectricLoad) )
```

jd-lara (changes requested):
We need to add an entry to the power flow docs to document this behavior

src/utils/power_flow.jl
```diff
@@ -142,19 +142,20 @@ function _write_pf_sol!(sys::System, nl_result)
             Q_gen = result[2 * ix]
             injection = get_components(StaticInjection, sys, x -> get_available(x))
```
jd-lara, line 143: Apply the same change here please.

jd-lara (approved):

PR #797, opened 2021-08-31
Add examples for show_components

daniel-thom:
There are no functional changes here; just additions to docstrings and the docs.


jd-lara (approved):
Thanks, this is great

PR #798, opened 2021-09-01
get time series counts bugfix

jd-lara:


daniel-thom (approved):
Thanks

test/test_system.jl
```diff
@@ -299,6 +299,18 @@ end
...
+    c_sys5 = PSB.build_system(
+        PSITestSystems,
+        "c_sys5_uc";
+        add_forecasts = true,
+        force_build = true,
+    )
+    _, ts_count, forecast_count = get_time_series_counts(c_sys5)
+    @test ts_count == 0
+    @test forecast_count == 3
+end
```
daniel-thom, line 312: I took the liberty of fixing a few bugs since I should have made this test in the first place.

Question to you @jd-lara: You were specifying `add_single_time_series = true` but this build function only supports forecasts. PSB doesn't throw errors if it receives unexpected kwargs. Are you OK with the change.

daniel-thom (commented):
LGTM

PR #810, opened 2021-09-12
Cb/update pm parsers

claytonpbarrows:
another minor update to pm parser


jd-lara (approved):

jd-lara:
@claytonpbarrows we need to wait to tag this since it breaks PSID's tests on the renewable models

rodrigomha:
The error in PSID is due to a change in a line name String. We can correct that one in the next release of PSID.

PR #811, opened 2021-09-14
Add check_components method

daniel-thom:
Adds new methods to `check_components` in order to allow PowerSimulations to perform checks on specific component types.

Also updates docstrings for a changed exception in IS.

### DIFF
--- a/test/test_validation.jl
+++ b/test/test_validation.jl
@@ -119,7 +119,7 @@ end
     @test_logs(
         (:error, r"Invalid range"),
         @test_throws(
-            PSY.InvalidRange,
+            PSY.InvalidValue,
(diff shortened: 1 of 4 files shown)


test/test_validation.jl
```diff
@@ -144,7 +144,7 @@ end
         (:error, r"Invalid range"),
         match_mode = :any,
         @test_throws(
```
jd-lara, line 146: I think this test needs update from invalid range to value too. Now it passes because of the match  condition

jd-lara (commented):
I think the tests need update

test/test_validation.jl
daniel-thom: It's actually correct. We may need to discuss. I didn't realize this part when we merged the IS PR.

PR #819, opened 2021-10-04
Update to latest pretty_table interface

daniel-thom:
The previous code was using a deprecated call.


daniel-thom:
@jd-lara Is there a reason that we don't add the JuliaFormatter package to either the main project or test project? It's annoying that an update to my global package environment can cause an inadvertent CI failure.

jd-lara (approved):

PR #836, opened 2021-11-18
correct matpower parsing

jd-lara:


jd-lara:
@rodrigomha this PR is addressing an issue in the power flow from Matpower and for some reason is breaking some tests from PSID. It seems we parse some of these cases for matpower files.

src/parsers/pm_io/matpower.jl
```diff
@@ -228,6 +234,14 @@ function _parse_matpower_string(data_string::String)
...
         for (i, gen_row) in enumerate(matlab_data["mpc.gen"])
             gen_data = row_to_typed_dict(gen_row, _mp_gen_columns)
+            bus_data = get(pv_bus_lookup, gen_data["gen_bus"], nothing)
+            if bus_data !== nothing
+                if bus_data["bus_type"] ∈ MP_FIX_VOLTAGE_BUSES &&
+                   bus_data["vm"] != gen_data["vg"]
+                    @info "Correcting vm in bus $(gen_data["gen_bus"]) to $(gen_data["vg"]) to match generator set-point"
+                    bus_data["vm"] = gen_data["vg"]
+                end
+            end
```
kdheepak, line 244: This will work fine when all the generators that are connected to a bus have the same `vg`. If generators with different `vg` are connected to the same bus, is that a data error? Can we throw a warning for that case?

jd-lara: We probably should do something to handle that case, but it could be possible to have that and only one generator available which would be valid.

jd-lara: @kdheepak 109a97e

kdheepak: Great. That works, although using a `Set{Int}` instead of `Dict{Int, Float64}` for tracking the corrected buses might be more appropriate.

jd-lara: With `Set` I can't store the previous value.

jd-lara:
@daniel-thom we are logging the errors from the test datasets now. Is it reasonable to disable the test about error logs?

daniel-thom:
> @daniel-thom we are logging the errors from the test datasets now. Is it reasonable to disable the test about error logs?

I'll make a case for keeping this "feature." It makes developers keep their tests tidy. If you take it away then developers can let error messages spew across the console. When a new developer runs the tests and sees messages in red scrolling across the screen they might think something is wrong. They have to really pay attention to the final result.

This is currently the case in the PowerSimulations dev branch. This "feature" got disabled and then someone let an error message go unsuppressed. A 76-line error message spans the screen every time you run the tests. I find this messy.

It's usually not a big deal to suppress expected error messages with the test_logs macro. If that is too onerous, you can also set the console log level to Logging.Error.level + 1.

That's my case (recommendation). As in the past for this question, I'm OK being out-voted.

jd-lara:
@daniel-thom I managed to fix most of the tests throwing errors due to bad input data in https://github.com/NREL-SIIP/PowerSystems.jl/pull/836/commits/37cb8a39a4506aed87b0384e3dc8419a7ebbe6d3 let me know if this is a good solution. 

@claytonpbarrows merging this will deviate us from PowerModels while @ccoffrin fixes https://github.com/lanl-ansi/PowerModels.jl/issues/798

daniel-thom (approved):

PR #837, opened 2021-12-01
Improve handling of set_name! on components

daniel-thom:
CI will fail until https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/268 is merged and tagged.

The purpose of this PR is block users from changing the name of a component that is attached to a system with `set_name!(component, name)`.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -694,11 +694,7 @@ Throws ArgumentError if the type is not stored.
 """
 # the argument order in this function is un-julian and should be deprecated in 2.0
 function remove_components!(::Type{T}, sys::System) where {T <: Component}
-    components = IS.remove_components!(T, sys.data)
-    for component in components
-        handle_component_removal!(sys, component)
-    end
-    return components
+    return remove_components!(sys, T)
 end
 
 function remove_components!(sys::System, ::Type{T}) where {T <: Component}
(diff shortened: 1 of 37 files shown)


src/base.jl
```diff
@@ -694,11 +694,7 @@ Throws ArgumentError if the type is not stored.
 """
 # the argument order in this function is un-julian and should be deprecated in 2.0
 function remove_components!(::Type{T}, sys::System) where {T <: Component}
-    components = IS.remove_components!(T, sys.data)
-    for component in components
-        handle_component_removal!(sys, component)
-    end
-    return components
+    return remove_components!(sys, T)
```
daniel-thom, line 697: This removes redundant code.

src/base.jl
```diff
@@ -1649,12 +1663,12 @@ function handle_component_addition!(sys::System, component::RegulationDevice; kw
 end
 
 function handle_component_addition!(sys::System, component::Branch; kwargs...)
-    handle_component_addition_common!(sys, component)
+    _handle_branch_addition_common!(sys, component)
```
daniel-thom, line 1666: This was a bad function name.

src/base.jl
```diff
@@ -1740,6 +1759,10 @@ function handle_component_removal!(sys::System, dyn_injector::DynamicInjection)
     @assert found
 end
 
+function _handle_component_removal_common!(component)
+    clear_units!(component)
```
daniel-thom, line 1763: We were not clearing the units in all removal functions. That should be fixed now.

src/descriptors/power_system_structs.json
```diff
@@ -221,6 +221,7 @@
         {
           "null_value": "init",
           "name": "name",
+          "exclude_setter": true,
```
daniel-thom, line 224: People usually copy/paste these objects when creating new structs. You all will need to remember to ensure people add this field.

claytonpbarrows (approved):
This looks good to me. As I understand it, the only downside is that the `exclude_setter: true` entry is critical for every `name` field in the struct definitions.

daniel-thom:
> This looks good to me. As I understand it, the only downside is that the `exclude_setter: true` entry is critical for every `name` field in the struct definitions.

Correct. They will get the current behavior.

daniel-thom:
@claytonpbarrows Note that we'll need to apply these changes to FuelSystems after this gets merged.

jd-lara (approved):

PR #838, opened 2021-12-11
Cb/cdm parser bugs

claytonpbarrows:
addresses issue where DataFrame columns aren't `::String` anymore. 

*Requires https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/270*

### DIFF
--- a/src/parsers/common.jl
+++ b/src/parsers/common.jl
@@ -69,17 +69,16 @@ function get_generator_type(fuel, unit_type, mappings::Dict{NamedTuple, DataType
     generator = nothing
 
     # Try to match the unit_type if it's defined. If it's nothing then just match on fuel.
-    for ut in (unit_type, nothing)
-        for fu in (fuel, nothing)
-            key = (fuel = fu, unit_type = ut)
-            if haskey(mappings, key)
-                generator = mappings[key]
-                break
-            end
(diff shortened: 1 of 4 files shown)


src/parsers/common.jl
```diff
@@ -69,15 +69,16 @@ function get_generator_type(fuel, unit_type, mappings::Dict{NamedTuple, DataType
...
+    for ut in (unit_type, nothing), fu in (fuel, nothing)
+        key = (fuel = fu, unit_type = ut)
         if haskey(mappings, key)
             generator = mappings[key]
             break
         end
     end
 
     if isnothing(generator)
+        @show (fuel = fuel, unit_type = unit_type)
```
daniel-thom, line 81: Stale?

src/parsers/power_system_table_data.jl
```diff
@@ -816,6 +816,8 @@ function calculate_variable_cost(
         # if there is only one point, use it to determine the constant $/MW cost
         var_cost = var_cost[1][1] * fuel_cost + vom
         fixed = 0.0
+    else
+        @warn "Unable to calculate variable cost for $(gen.name)" var_cost maxlog = 5
```
daniel-thom, line 820: cost_colnames.columns would have to be empty to get here. Is that possible?

daniel-thom (approved):

PR #839, opened 2021-12-16
Add NODE

m-bossart:
Work in progress, starting point to determine structure of the NODE component. 

It seems like we would need a constant current load as the static injector to which the NODE can be added? And if the NODE is intended to reproduce the steady state as well then the static injector could be set to zero?


jd-lara:
@m-bossart this type should be defined separately in the NODE repo since this won't be a too general use struct.

jd-lara:
@m-bossart I will close this since we will be implementing this in a separate package.

PR #840, opened 2021-12-17
Fix serialization of Union types with Nothing

daniel-thom:
This fixes a bug when deserializing a HybridSystem that has a interconnection_impedance with a complex number instead of nothing.


jd-lara (commented):
We can merge this because this is a good fix to have. But this field should have been just `::Complex` and have a default of 0.0.

jd-lara (approved):

PR #843, opened 2022-01-06
Add subcomponent time series to subsystem

daniel-thom:
The purpose of this PR is to make HybridSystem subcomponent time series available at the HybridSystem for use in PowerSimulations. If you add a HybridSystem with subcomponents already attached to the System then the HybridSystem will have references to the their time series in the format `<typeof(subcomponent)>__<time_series_name>`.

There is one possible problem with this solution. If the user adds time series data to a subcomponent _after_ adding the HybridSystem to the System then it is up to that user to call `copy_subcomponent_time_series!`. Is this a deal breaker?

### DIFF
--- a/test/test_hybrid_system.jl
+++ b/test/test_hybrid_system.jl
@@ -39,6 +39,8 @@ end
     thermal_unit = nothing
     subcomponents = collect(get_subcomponents(h_sys))
     @test length(subcomponents) == 4
+    expected_time_series_names = Set{String}()
+    num_time_series = 0
     for subcomponent in subcomponents
         @test !PSY.is_attached(subcomponent, sys)
         @test IS.is_attached(subcomponent, sys.data.masked_components)
@@ -47,12 +49,20 @@ end
         elseif subcomponent isa ThermalStandard
             thermal_unit = subcomponent
         end
+        for ts in get_time_series_multiple(subcomponent)
+            push!(
+                expected_time_series_names,
+                PSY.make_subsystem_time_series_name(subcomponent, ts),
+            )
+            num_time_series += 1
+        end
     end
     @test electric_load !== nothing
     @test thermal_unit !== nothing
 
     ts = collect(get_time_series_multiple(h_sys, type = TimeSeriesData))
-    @test length(ts) == 0
+    @test length(ts) == num_time_series
+    @test Set((get_name(x) for x in ts)) == expected_time_series_names
 
     @test get_time_series(SingleTimeSeries, electric_load, "max_active_power") isa
           SingleTimeSeries
@@ -117,4 +127,10 @@ end
     forecast = Deterministic(name, data, resolution)
     add_time_series!(sys, thermal_unit, forecast)
     @test get_time_series(Deterministic, thermal_unit, name) isa Deterministic
+    copy_subcomponent_time_series!(h_sys, thermal_unit)
+    @test get_time_series(
+        Deterministic,
+        h_sys,
+        PSY.make_subsystem_time_series_name(thermal_unit, forecast),
+    ) isa Deterministic
 end
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -233,6 +233,7 @@ export remove_time_series!
 export check_time_series_consistency
 export clear_time_series!
 export copy_time_series!
+export copy_subcomponent_time_series!
 export add_component!
 export add_components!
 export remove_component!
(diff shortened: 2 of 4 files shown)


test/test_hybrid_system.jl
```diff
@@ -117,4 +127,10 @@ end
     forecast = Deterministic(name, data, resolution)
     add_time_series!(sys, thermal_unit, forecast)
     @test get_time_series(Deterministic, thermal_unit, name) isa Deterministic
+    copy_subcomponent_time_series!(h_sys, thermal_unit)
```
daniel-thom, line 130: This is an example of the potential problem.

jd-lara (approved):
I think this is an ok work flow, since a user needs to call `add_subcomponent` we just need to document this behavior for the time series in the case of hybrids.

test/test_hybrid_system.jl
claytonpbarrows: Is this something we can detect and throw a log message? I'm thinking we'd have something like:

```Julia
julia> add_time_series!(sys, thermal_unit, forecast) 
┌ Warning: thermal_name is a Subcomponent of a HybridSys Component, call `copy_subcomponent_time_series!(h_sys, subcomponent)` to enable time series access from the HybridSys
```

claytonpbarrows (approved):
Just the one request about a log message if it's possible/easy. Otherwise, I think this is OK.

test/test_hybrid_system.jl
daniel-thom: The subcomponent doesn't know that it's attached to a Hybrid. The system could detect that the subcomponent is "masked" and log something about that. The system doesn't know anything about Hybrid/subcomponent, so if we wanted to log something like what you wrote, it could cause a problem if we reused the same pattern in the future.

daniel-thom:
> add_subcomponent

What do you mean by `add_subcomponent`?

jd-lara:
> > add_subcomponent
> 
> What do you mean by `add_subcomponent`?

This could be an issue when a user adds a new component to the hybrid that comes with its own time series. But they already have to call `add_subcomponent`

daniel-thom:
> > > add_subcomponent
> > 
> > 
> > What do you mean by `add_subcomponent`?
> 
> This could be an issue when a user adds a new component to the hybrid that comes with its own time series. But they already have to call `add_subcomponent`

I guess you mean the functions like `set_thermal_unit!` and `set_electric_load!`. I had forgotten about those. In those cases the subcomponents cannot be attached to the system, and so cannot have time series. The issue here is that the user is allowed to add time series to subcomponents regardless of when the subcomponents got attached to the system or Hybrid.

jd-lara:
> > > > add_subcomponent
> > > 
> > > 
> > > What do you mean by `add_subcomponent`?
> > 
> > 
> > This could be an issue when a user adds a new component to the hybrid that comes with its own time series. But they already have to call `add_subcomponent`
> 
> I guess you mean the functions like `set_thermal_unit!` and `set_electric_load!`. I had forgotten about those. In those cases the subcomponents cannot be attached to the system, and so cannot have time series. The issue here is that the user is allowed to add time series to subcomponents regardless of when the subcomponents got attached to the system or Hybrid.

Confirm that a user that wants to add an existing component with time series to a hybrid needs to:  

1. Collect the time series object and remove it from the system and the component. 
2. remove the component from the system 
3. add the component to the hybrid 
4. attach the time series back to the component 
5. call `copy_subcomponent_time_series!`

daniel-thom:
> > > > > add_subcomponent
> > > > 
> > > > 
> > > > What do you mean by `add_subcomponent`?
> > > 
> > > 
> > > This could be an issue when a user adds a new component to the hybrid that comes with its own time series. But they already have to call `add_subcomponent`
> > 
> > 
> > I guess you mean the functions like `set_thermal_unit!` and `set_electric_load!`. I had forgotten about those. In those cases the subcomponents cannot be attached to the system, and so cannot have time series. The issue here is that the user is allowed to add time series to subcomponents regardless of when the subcomponents got attached to the system or Hybrid.
> 
> Confirm that a user that wants to add an existing component with time series to a hybrid needs to:
> 
> 1. Collect the time series object and remove it from the system and the component.
> 2. remove the component from the system
> 3. add the component to the hybrid
> 4. attach the time series back to the component
> 5. call `copy_subcomponent_time_series!`

Steps 1-4 are managed by PSY. I think the only potential problem is adding new time series to subcomponents.

test/test_hybrid_system.jl
claytonpbarrows: No worries then. If it causes strife in the future we can try to log messages based on masked status. For now, we can just make it clear in the docs.

PR #844, opened 2022-01-07
improving flexibility of table data parsing

claytonpbarrows:
adding some flexibility to the tabular parser:
 - removed requirement for bus_id column and defaulting id's based on row order
 - handles NA fuel entries
 - improved some flexibility on Int->Float conversion

### DIFF
--- a/src/parsers/power_system_table_data.jl
+++ b/src/parsers/power_system_table_data.jl
@@ -298,7 +298,7 @@ Add buses and areas to the System from the raw data.
 """
 function bus_csv_parser!(sys::System, data::PowerSystemTableData)
     ix = 1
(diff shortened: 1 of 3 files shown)


src/parsers/power_system_table_data.jl
```diff
@@ -297,6 +297,7 @@ Add buses and areas to the System from the raw data.
 
 """
 function bus_csv_parser!(sys::System, data::PowerSystemTableData)
+    ix = 1
```
daniel-thom, line 300: You should be able to let julia handle the index with
```
for (i, bus) in iterate_rows(data, InputCategory.BUS)
```

daniel-thom (approved):

PR #856, opened 2022-02-15
Updating copy subcomponent time series method

sourabhdalvi:
The previous implementation of `copy_subcomponent_time_series` would create a Set of existing time series names and types but this method would use the original name of the time series  object rather than the name stored in the time series metadata. This fix would allow the user to add two time series of the same type and label from different subcomponents of a hybrid system.


src/models/static_injection_subsystem.jl
```diff
@@ -16,25 +16,40 @@ function copy_subcomponent_time_series!(
     subsystem::StaticInjectionSubsystem,
     subcomponent::Component,
 )
-    existing_ts = Set((typeof(x), get_name(x)) for x in get_time_series_multiple(subsystem))
-    name_mapping = Dict{String, String}()
+    # the existing_ts can remove entreis from the set if the Subsystem has two device of
```
jd-lara, line 19: Check spelling in the comment.

jd-lara (approved):

PR #857, opened 2022-02-16
Cb/unit conversion

claytonpbarrows:
adds a few extra unit conversion options and flexible handling of storage position parameters for the tabular data parsing process.


jd-lara:
Does this branch close any of the open issues about parsing table data?

src/parsers/common.jl
```diff
@@ -123,18 +123,48 @@ function convert_units!(
     value::Float64,
     unit_conversion::NamedTuple{(:From, :To), Tuple{String, String}},
 )
-    if unit_conversion.From == "degree" && unit_conversion.To == "radian"
+    if uppercase(unit_conversion.From) == uppercase("degree") &&
```
jd-lara, line 126: Would it be better in this case to use `normalize` https://docs.julialang.org/en/v1/stdlib/Unicode/#Unicode.normalize

jd-lara (approved):

PR #858, opened 2022-03-02
Add get_components method with filter_func

daniel-thom:
This follows the Julia parameter-order convention where a function is listed first so that do blocks can be used for the function.

The existing implementation is unchanged, so users can choose their preferred way of using the feature.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -822,14 +822,26 @@ Call collect on the result if an array is desired.
 ```julia
 iter = PowerSystems.get_components(ThermalStandard, sys)
 iter = PowerSystems.get_components(Generator, sys)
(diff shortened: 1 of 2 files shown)


src/parsers/power_system_table_data.jl
```diff
@@ -1195,6 +1195,7 @@ function get_storage_by_generator(data::PowerSystemTableData, gen_name::Abstract
     end
 
     if length(head) != 1
+        @show head
```
kdheepak, line 1198: Why is this in here?

daniel-thom: accidental addition. I'll delete it. Thanks.

kdheepak (approved):

jd-lara (approved):

PR #859, opened 2022-03-03
Fix Saturation calculations

rodrigomha:
Fix a bug when computing saturation coefficients for Synchronous Machines.


jd-lara (approved):

jd-lara:
The fails in PSID are expected since we need to update the tests to account for the change in definitions.

PR #865, opened 2022-03-04
Fix saturation bug

rodrigomha:
The following fix a bug on the checking when to neglect saturation


jd-lara (approved):
Thanks for the fix. PSID test remains broken

PR #868, opened 2022-03-10
Disable standalone mode for HTML display

daniel-thom:


src/utils/print.jl
```diff
@@ -29,6 +29,7 @@ function Base.show(io::IO, ::MIME"text/html", sys::System)
         sys.data,
         backend = :html,
         tf = PrettyTables.tf_html_simple,
+        standalone = false,
```
kdheepak, line 32: So we want the same thing on line 53 too?

daniel-thom: Those are not html-specific. There are two other places, however.

kdheepak: That should resolve it! I was commenting from my phone and wasn’t able to read the code properly I think.

kdheepak (approved):

PR #869, opened 2022-03-18
Add 2 new AVRs

rodrigomha:
To do: Testing for the new 2 AVRs


jd-lara:
@rodrigomha you can rebase this PR now and make it not a draft to integrate the AVR's into PSID.

PR #870, opened 2022-03-21
Update generated structs with disabled formatter

daniel-thom:


jd-lara (approved):
Tests will fail until the new release of IS. WIll wait until that's done to merge

PR #880, opened 2022-04-14
Add Exponential Load model

rodrigomha:
The following PR adds the struct of the Exponential Load model used in dynamic studies.


src/descriptors/power_system_structs.json
```diff
@@ -1418,6 +1418,121 @@
...
+        },
+        {
+          "name": "services",
+          "data_type": "Vector{Service}",
+          "comment": "Services that this device contributes to",
+          "null_value": "Device[]",
+          "default": "Device[]"
+        },
+        {
+          "name": "dynamic_injector",
```
jd-lara, line 1506: Do we need a dynamic injector here?

rodrigomha: Probably for consistency. In the future when implementing the DynamicExponentialLoad we probably should attach it to an ExponentialLoad, but it is likely that any `StaticLoad` can hold any dynamic load.

So in the future, all dynamic loads should be attached to `PowerLoad` if we remove dynamic_injector of this struct.

jd-lara (approved):

PR #885, opened 2022-05-03
Add induction motors structs

rodrigomha:
@jd-lara I think we may want to go at some point with adding DynamicLoad abstract at some point.

From this point onwards we are expecting to add more and more Dynamic Load models, so it may be a good idea to think a strategy for that.

### DIFF
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -80,6 +80,8 @@ export RegulationDevice
 
 export GenericDER
 export AggregateDistributedGenerationA
+export SingleCageInductionMachine
+export SimplifiedSingleCageInductionMachine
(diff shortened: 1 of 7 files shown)


jd-lara:
I think the big question is how distinct are DynamicLoads from DynamicInjectors that merit and extra level of abstraction

rodrigomha:
> I think the big question is how distinct are DynamicLoads from DynamicInjectors that merit and extra level of abstraction

Yeah I agree, is just that at some point if we have like 10 different dynamic loads, it may be useful to have some specific wrappers in PSID for just DynamicLoads.

jd-lara (approved):

PR #893, opened 2022-05-25
Add Induction Motor Additional Params

rodrigomha:
Add additional parameters to avoid calculation at each integration step in PSID


rodrigomha:
By the way, the issue is related to Sundials KLU... Not sure what happen.

jd-lara (approved):

PR #895, opened 2022-05-31
fix shunt location in ybus

### LINKED ISSUE #879: Problem with parsing of PSSE transformer data
I have been using PowerSystems.jl to parse data from PSSE .raw files to Julia and I have noticed that the transformer data parsed in Julia does not correspond to the data created by PSSE for power flow. I have found out that the main issue is that the transformer impedance as given in PSSE data might have to be scaled up by factor: (WINDV1/NOMV1)^2 i.e. desired voltage (nominal voltage of the bus)/nominal voltage (voltage on the name plate) of the primary winding. Similar issue was existing in PowerModels.jl:
https://github.com/lanl-ansi/PowerModels.jl/issues/629.

Having in mind this, I would propose that parsing of the transformer data should be double-checked and that the package should also parse WINDV1, NOMV1, WINDV2, NOMV2 as transformer parameters. What would be your thoughts on this?

jd-lara:
closes #879. @SSte990 it seems to me that we had an issue with the location of the shunt when making the Ybus.


jd-lara:
@claytonpbarrows we had an issue in Ybus that was never caught.

rodrigomha (approved):
LGTM: Wait for test to pass

PR #897, opened 2022-06-08
Add REGCA new struct for future PSCAD validation

rodrigomha:


src/PowerSystems.jl
```diff
@@ -147,6 +147,7 @@ export SteamTurbineGov1
 export Converter
 export AverageConverter
 export RenewableEnergyConverterTypeA
```
jd-lara, line 149: We need to add a note here for 2.0 that this should become `RenewableEnergyCurrentConverterTypeA`

jd-lara (approved):

PR #898, opened 2022-06-15
Jd/use pfdata

jd-lara:
@daniel-thom and @nickrobinson251 this is a prototype to create a system from a PowerFlowData.jl network object. I only implemented the bus creation. 

it seems like it should be possible to implement this data read, but we will lose some of the data validation from PowerModels.


jd-lara:
@daniel-thom and @nickrobinson251, I implemented a mechanism to read data from PSSe 30 in this PR. I only have a small test case to validate the fields but this is a good MVP for what that would look like.

jd-lara:
This PR requires and equivalent refactor of PSB to make it compatible with the StandardLoad parsing.

animatedgoshawk:
@jd-lara 

Sorry for the ping, but I figured this was a good place to put this note since the `psy2` branch isn't merged yet: I tried running the merged PowerFlowData-based parser on a PSS/E v30 file, and got the following error inside `read_bus!`:

```julia
ERROR: UndefVarError: MatpowerBusTypes not defined
Stacktrace:
 [1] read_bus!(sys::System, buses::PowerFlowData.Buses30, data::PowerFlowData.Network; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ PowerSystems ~/.julia/packages/PowerSystems/7WGGj/src/parsers/powerflowdata_data.jl:150
```

My Julia isn't good enough to propose a fix (working on that) but I assume this is intended to work now? Wanted to give you a heads up if so.

PR #903, opened 2022-07-05
Speed-up battery parsing

josephmckinsey:
The PowerSystems table parsing has O(n^2) complexity for parsing batteries, where every row of the storage data frame is iterated or each battery. This pull request creates a map to the storage head and tail instead as a preprocessing step.


daniel-thom (approved):
Looks good

PR #905, opened 2022-08-01
Add Sauer Pai struct

marenat:


rodrigomha (changes requested):
You need to run the generate structs. To do so in your PowerSystems folder and project run the following:

```
using InfrastructureSystems
```

rodrigomha (approved):
Looks good. The PSID error is something that we are fixing it there and is not related with this PR.

jd-lara:
@marenat This PR is missing the tests for the new struct that it can be added to a system

rodrigomha:
A better example would be this one:

https://github.com/NREL-SIIP/PowerSystems.jl/blob/1020c4ae5322fe088358d2e544152b375acc19e6/test/test_dynamic_generator.jl#L49

You can add the SauerPai model in that test set

rodrigomha (approved):
With the additional test LGTM

PR #923, opened 2022-11-07
Add new active constant power and generic load models

rodrigomha:
The following add the struct for the 12-states Active Load Dynamic Model taken from: https://ieeexplore.ieee.org/document/6415284
and a 2-states Generic Load Model from https://ieeexplore.ieee.org/document/317575


src/descriptors/power_system_structs.json
```diff
@@ -2000,6 +2000,322 @@
       ],
       "supertype": "DynamicInjection"
     },
+    {
+      "struct_name": "GenericDynamicLoad",
```
jd-lara, line 2004: Why Generic? is there something more descriptive?

rodrigomha: Is because of the paper. Maybe we can go with `ExponentialDynamicLoad`?

jd-lara: Yeah, that is better

jd-lara (approved):

PR #925, opened 2022-11-11
Update LCL Filter model

rodrigomha:
The following add the parameter to decide if a filter should be algebraic or differential for PSID.


jd-lara (commented):
Not sure about this change. If the model is different it defeats the purpose of using Julia dispatch

rodrigomha:
We will deal with this aspect in PSID.

PR #941, opened 2022-12-29
Jd/misc fixes

jd-lara:
Group of miscellaneous changes pending for PSY 2.0 

- Remove unnecessary enums and code by aligning bus types with Matpower's 
- Add Switchedsunts 
- rename OuterControl fields to avoid type instability 
- Use Min_Max in structs definitions consistently
- Add limits to the phaseshift transformer 
- change the use of get_components to be Julia consistent. 
- Added must run flag to ThermalStandard


src/models/generated/AVRTypeI.jl
```diff
@@ -34,7 +34,7 @@ Parameters of an Automatic Voltage Regulator Type I - Resembles IEEE Type DC1
 - `Te::Float64`: Field Circuit Time Constant in s, validation range: `(0, nothing)`
 - `Tf::Float64`: Stabilizer Time Constant in s, validation range: `(0, nothing)`
 - `Tr::Float64`: Voltage Measurement Time Constant in s, validation range: `(0, nothing)`
-- `Va_lim::NamedTuple{(:min, :max), Tuple{Float64, Float64}}`: Limits for pi controler `(Va_min, Va_max)`
+- `Va_lim::Min_Max`: Limits for pi controler `(Va_min, Va_max)`
```
daniel-thom, line 37: While we're breaking things, I find these aliases unpredictable. Could we standardize the naming conventions?
```
Min_Max -> MinMax
Up_Down -> UpDown
StartUp_ShutDown -> StartUpShutDown
From_To_Float -> FromToFloat
FromTo_ToFrom_Float -> I have no idea and don't see it being used
```

We could keep the originals to avoid breaking people - but change all uses.

daniel-thom: If you disagree, no worries, I'll still approve.

jd-lara: I am fine with the rename, I will make the change and push a commit

jd-lara: @daniel-thom I made the use of Alias consistent. I think that `FromTo_ToFrom` needs an exception because `FromToToFrom`is very confusing to read.

daniel-thom (approved):

PR #960, opened 2023-01-31
Add Standard DC and T-Model DC Lines

rodrigomha:
A standard DC line model only has a resistor, and is suitable for AC/DC Power Flow studies when you only need conductance. 

A T-Model Line is a common implementation of HVDC lines that consists on two DC Lines, using two equivalent series smoothing reactors with inductance `l_dc` and shunt capacitor `c_dc`. The series resistor `r_dc` represents the line resistance.

image

The terminal voltages of the HVDC lines are defined by the HVDC controller (could be LCC or VSC) and the differential equations of the T-Line model are simply:

```
l_{dc} \dot{I}_{dcr} =  V_{dcr} - r_{dc} I_{dcr} - V_{dcl}
c_{dc} \dot{V}_{dcl} = I_{dcr} - I_{dci}
l_{dc} \dot{I}_{dci} =  V_{dcl} - r_{dc} I_{dci} - V_{dci}
```

Note `DCLine` has a default value of inductor `l = 0.0`, but the T-Line Model could be built using three `DCBus`, two `DCLine` and a shunt capacitor in the mid `DCBus`.


jd-lara:
we need to add the additional node similarly to the ThreeWInding Transformer I think.

src/descriptors/power_system_structs.json
```diff
@@ -1245,6 +1245,159 @@
             ],
             "supertype": "DCBranch"
         },
+        {
```
jd-lara, line 1248: What's the difference between this and a T line without the shunt? For PSI the shunt doesn't really matter.

src/models/generated/Arc.jl
```diff
@@ -6,23 +6,23 @@ This file is auto-generated. Do not edit.
 
 """
     mutable struct Arc <: Topology
```
jd-lara, line 8: I am not sure if we should allow `Arc(DCBus, ACBus)`

src/models/generated/DCLine.jl
```diff
@@ -0,0 +1,119 @@
...
+- `arc::Arc`
+- `r::Float64`: Resistance system per-unit value
+- `active_power_limits_from::MinMax`
+- `active_power_limits_to::MinMax`
+- `l::Float64`: Inductance system per-unit value
+- `services::Vector{Service}`: Services that this device contributes to
+- `ext::Dict{String, Any}`
+- `internal::InfrastructureSystemsInternal`: power system internal reference, do not modify
+"""
+mutable struct DCLine <: DCBranch
```
jd-lara, line 37: I don't think we need them both, the T model generalizes fine.

src/models/generated/TModelHVDCLine.jl
```diff
@@ -0,0 +1,128 @@
...
+- `r::Float64`: Series Resistance system per-unit value
+- `l::Float64`: Series Inductance system per-unit value
+- `c::Float64`: Shunt capacitance system per-unit value
+- `active_power_limits_from::MinMax`
+- `active_power_limits_to::MinMax`
+- `services::Vector{Service}`: Services that this device contributes to
+- `ext::Dict{String, Any}`
+- `internal::InfrastructureSystemsInternal`: power system internal reference, do not modify
+"""
+mutable struct TModelHVDCLine <: DCBranch
```
jd-lara, line 39: If this is a DCBranch, do we require that a DC branch connect two DC buses. As a result, TwoTerminalHVDC would imply an ACBranch 🤔

src/models/generated/DCLine.jl
rodrigomha: The T-Model is way more advanced for many applications. Power Flows could suffice with a DC line with a resistor.

If we want to use the T-model for a Power Flow on which we only care about the series resistor, we would have to create a T-Line with a shunt = 0, and a resistor with value `0.5 r`, then we would need to aggregate both resistors.

If we include a shunt, then we will need to define this virtual bus similar to the three-winding transformer.

My suggestion is keeping the DCLine and use this one for initial implementations on PSI. In any, case we can use the T-model and ignore everything for PSI, if you think about this.

TLDR: Use DCLine for PSI studies, use T-model for PSID.

src/models/generated/TModelHVDCLine.jl
rodrigomha: This goes in the Arc change. We would need to have a DCArc and ACArc in this case. The generalization of using Union is just the straightforward approach.

jd-lara (approved):
This will work for now. The potential changes to the rules on the Arcs needs to be determined later

PR #962, opened 2023-02-01
Jd/two terminal rename

jd-lara:
This PR renames existing HVDC structures into TwoTerminal ones and sets the lines as ACBranches since they now connect to AC Buses. DCBranches will be reserved for Arcs that connect DC Buses


claytonpbarrows (approved):
LGTM

PR #967, opened 2023-02-06
Change limits of HVDC T-model line to rate

rodrigomha:


jd-lara:
PSB needs to be updated too I guess.

jd-lara (approved):

rodrigomha:
> PSB needs to be updated too I guess.

I updated TestData already. I don't think PSB needs to be updated, it is just calling the functions in PSTestData.

jd-lara:
Ok, yeah makes sense

PR #976, opened 2023-03-07
[DNMY] Update naming and issues detected when updating PSID

rodrigomha:
We are currently updating PSID for PSY 2.0 and some minor issues will be updated in this branch


jd-lara (approved):
LGTM!

Issue #973, opened 2023-02-28
Print Time Series Names (summary) when displaying component summary


rodrigomha:

rodrigomha:
Basically when you display a component you'll see something like this:
![image](https://user-images.githubusercontent.com/24548207/223292111-2228cd46-be80-4117-b416-db6abdf99bc9.png)
so you need to go directly in the container to find the name, in this case, `requirements`.
![image](https://user-images.githubusercontent.com/24548207/223292272-2c6058b8-2e17-4384-b4fd-ee4c1bdc05ea.png)

I think is a good idea to depict one single name if both are the same.

PR #978, opened 2023-03-10
Improve printing of time series instances

### LINKED ISSUE #971: Give better error message if a user stores a component inside the System ext field
If a user stores a component inside the System ext field and then tries to serialize the system to JSON, they will receive an error that is difficult to debug and correct. Refer to #970. We could perform a sanity check on the contents of the ext field before trying to serialize the system. Components and containers of components are definitely not allowed.

daniel-thom:
Fixes #971.

@rodrigomha Here is how this shows up. We could print other fields, reverse the order of these two fields. Let me know what you'd like.
```
215_HYDRO_3 (HydroEnergyReservoir):
   name: 215_HYDRO_3
   available: true
   bus: Barton (Bus)
   active_power: 0.5
   reactive_power: 0.16
   rating: 0.5249761899362675
   prime_mover: PrimeMovers.HY = 16
   active_power_limits: (min = 0.0, max = 0.5)
   reactive_power_limits: (min = -0.1, max = 0.16)
   ramp_limits: (up = 0.5, down = 0.5)
   time_limits: (up = 0.0, down = 0.0)
   base_power: 100.0
   storage_capacity: 20.0
   inflow: 4.0
   initial_storage: 10.0
   operation_cost: TwoPartCost
   storage_target: 1.0
   conversion_factor: 1.0
   time_at_status: 10000.0
   services: 0-element Vector{Service}
   dynamic_injector: nothing
   ext: Dict{String, Any}()
   time_series_container:
      max_active_power: SingleTimeSeries
      storage_capacity: SingleTimeSeries
      inflow: SingleTimeSeries
```

### DIFF
--- a/src/utils/print.jl
+++ b/src/utils/print.jl
@@ -199,8 +199,13 @@ function Base.show(io::IO, ::MIME"text/plain", ist::Component)
                 print(io, "\n   ")
                 show(io, MIME"text/plain"(), obj.units_info)
                 continue
-            elseif obj isa IS.TimeSeriesContainer ||
-                   obj isa InfrastructureSystemsType ||
+            elseif obj isa IS.TimeSeriesContainer
+                val = ""
+                for (key, metadata) in obj.data
+                    ts_type = IS.time_series_metadata_to_data(metadata)
+                    val *= "\n      $(key.name): $ts_type"
+                end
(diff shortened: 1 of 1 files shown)


rodrigomha (approved):
I like it 👍 . I think this one closes #973, not #971

PR #984, opened 2023-04-03
Fix ThermalStandard must_run in tabular data parser

### LINKED ISSUE #981: TableDataParaser not catching must run from csv file
Is not reading must run from the csv file when creating `SystemTableData`

```julia
 return ThermalMultiStart(;
        name = get_name(thermal_gen),
        available = get_available(thermal_gen),
        status = get_status(thermal_gen),
        bus = get_bus(thermal_gen),
        active_power = get_active_power(thermal_gen),
        reactive_power = get_reactive_power(thermal_gen),
        rating = get_rating(thermal_gen),
        prime_mover = get_prime_mover(thermal_gen),
        fuel = get_fuel(thermal_gen),
        active_power_limits = get_active_power_limits(thermal_gen),
        reactive_power_limits = get_reactive_power_limits(thermal_gen),
        ramp_limits = get_ramp_limits(thermal_gen),
        power_trajectory = power_trajectory,
        time_limits = get_time_limits(thermal_gen),
        start_time_limits = startup_timelimits,
        start_types = start_types,
        operation_cost = op_cost,
        base_power = get_base_power(thermal_gen),
        time_at_status = get_time_at_status(thermal_gen),
        must_run = gen.must_run,
    )
end
```

scdhulipala:

Check the [contributor guidelines](https://nrel-siip.github.io/PowerSystems.jl/stable/code_base_developer_guide/developer/)

1. Fix tabular data parser to parse _must_run_  for **ThermalStandard**
2. [Issue Fix](https://github.com/NREL-SIIP/PowerSystems.jl/issues/981)

closes #981

### DIFF
--- a/src/parsers/power_system_table_data.jl
+++ b/src/parsers/power_system_table_data.jl
@@ -1082,9 +1082,6 @@ function make_thermal_generator_multistart(
 
     op_cost = MultiStartCost(var_cost, no_load_cost, fixed, startup_cost, shutdown_cost)
 
-    gen_must_run = isnothing(gen.must_run) ? false : gen.must_run
-    gen_must_run = isa(gen_must_run,Bool) ? gen_must_run : parse(Bool,lowercase(String(gen_must_run)))
-
     return ThermalMultiStart(;
         name = get_name(thermal_gen),
         available = get_available(thermal_gen),
@@ -1105,7 +1102,7 @@ function make_thermal_generator_multistart(
         operation_cost = op_cost,
         base_power = get_base_power(thermal_gen),
         time_at_status = get_time_at_status(thermal_gen),
-        must_run = gen_must_run,
+        must_run = get_must_run(thermal_gen),
     )
 end
 
--- a/src/parsers/power_system_table_data.jl
+++ b/src/parsers/power_system_table_data.jl
@@ -1005,10 +1005,8 @@ function make_thermal_generator(data::PowerSystemTableData, gen, cost_colnames, 
     op_cost = ThreePartCost(var_cost, fixed, startup_cost, shutdown_cost)
 
     gen_must_run = isnothing(gen.must_run) ? false : gen.must_run
-    gen_must_run = if isa(gen_must_run, Bool)
-        gen_must_run
-    else
-        parse(Bool, lowercase(String(gen_must_run)))
+    if !isa(gen_must_run, Bool)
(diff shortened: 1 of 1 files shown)


src/parsers/power_system_table_data.jl
```diff
@@ -1020,6 +1020,7 @@ function make_thermal_generator(data::PowerSystemTableData, gen, cost_colnames,
         time_limits = timelimits,
         operation_cost = op_cost,
         base_power = base_power,
+        must_run = gen.must_run,
```
jd-lara, line 1023: where is the must run info coming from?

scdhulipala: It is defined in power_system_structs.json and the user_descriptors.yml of the data, so the data would come from gen.csv. I'm not sure if that is the answer you're looking for.

jd-lara: But that value isn't been read and passed to the `ThermalStandard` call.

scdhulipala: Yeah, gen has a field must_run because it is defined in both user_descriptors and power_system_inputs.

jd-lara: Yes, that value is read from the CSV into the gen dict but you aren't reading it from that dict. Did you test this?

scdhulipala: I'm running some more tests now with Pedro. I made this a draft for now. I will open a PR once I get the go ahead from Pedro.

scdhulipala: I tested this on Pedro's ERCOT test data. This does what it is supposed to in general but will fail due to CSV data parsing. For instance, I printed gen.must_run for his data for debugging purposes and sometimes Bool is read as an InlineString.
gen_must_run = false
gen_must_run = InlineStrings.String7("False")

jd-lara: Why not fix it in this PR?

jd-lara:
Using parse is dangerous during runtime applications. It's better to employ a string comparison

scdhulipala:
> Using parse is dangerous during runtime applications. It's better to employ a string comparison

The only reason I was usingparse was because must_run could be either a true/false (or) 1/0 which could be parsed as a String by the CSV parser (because of 'NA' values). Should I use stringcompariosn for all these cases then?

PR #988, opened 2023-04-18
Fixing unit conversion reserves 

### LINKED ISSUE #972: get_time_series methods don't convert time series values to natural units for Services
Seems like when retrieving time series values when the unit setting is set to "natural_units" are not converted, other devices like PowerLoads have time series  converted to natural units but not for services. attached an example for this on the RTS
For reserves :
image
For Loads :
image

sourabhdalvi:
This PR fixes #972, which will convert the requirement field for Reserves.

### DIFF
--- a/src/descriptors/power_system_structs.json
+++ b/src/descriptors/power_system_structs.json
@@ -4162,7 +4162,8 @@
                         "min": 0,
                         "max": null
                     },
-                    "validation_action": "error"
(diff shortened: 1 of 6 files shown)


jd-lara (approved):
I will approve but we need to make sure this PR doesn't screw up anything in PSI when creating parameters there.

PR #1011, opened 2023-08-28
Jd/updates to generic battery

jd-lara:
Updates the storage structs gearing towards the re-write in SSS.

### DIFF
--- a/src/models/generated/BatteryEMS.jl
+++ b/src/models/generated/BatteryEMS.jl
@@ -20,7 +20,7 @@ This file is auto-generated. Do not edit.
         reactive_power::Float64
         reactive_power_limits::Union{Nothing, MinMax}
         base_power::Float64
-        operation_cost::Union{Nothing, StorageManagementCost}
+        operation_cost::StorageManagementCost
         storage_target::Float64
         cycle_limits::Int
         services::Vector{Service}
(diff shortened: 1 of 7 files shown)


rodrigomha (approved):
LGTM

src/models/generated/BatteryEMS.jl
```diff
@@ -73,9 +75,11 @@ mutable struct BatteryEMS <: Storage
     reactive_power_limits::Union{Nothing, MinMax}
     "Base power of the unit in MVA"
     base_power::Float64
-    operation_cost::Union{Nothing, OperationalCost}
+    operation_cost::StorageManagementCost
     "Storage target at the end of simulation as ratio of storage capacity."
     storage_target::Float64
+    "Storage Maximum number of cycles per year"
+    cycle_limits::Int
```
sourabhdalvi, line 82: rather than having it as static value  to limit the cycles, I think it would be great if we can also define the cost for exceeding the cycling limit

sourabhdalvi (changes requested):

src/models/generated/BatteryEMS.jl
jd-lara: that is defined in the cost function. This is just the limit data

PR #1014, opened 2023-09-11
Add current lims to psy3

rodrigomha:
Currently is not working. I think is related to having an abstract type in the limiter field of the `DynamicInverter` struct.

### DIFF
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -195,6 +195,11 @@ export VoltageModeControl
 export CurrentModeControl
 export RECurrentControlB
 
+# InverterLimiters Export
+export MagnitudeCurrentLimiter
+export InstantaneousCurrentLimiter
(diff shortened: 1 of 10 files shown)


src/PowerSystems.jl
```diff
@@ -195,6 +195,11 @@ export VoltageModeControl
 export CurrentModeControl
 export RECurrentControlB
 
+# InverterLimiters Export
+export MagnitudeCurrentLimiter
```
jd-lara, line 199: We missing the circle one?

rodrigomha: Magnitude is circle, Instantaneous is Square, based on this paper: https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=9973369

We can change the name to circular, and square

jd-lara (approved):

PR #1027, opened 2023-11-17
Add DEGOV

m-bossart:
Adds DEGOV governor for purpose of including delays in PSID.


m-bossart:
@jd-lara I think the last test will pass when the new version of PowerNetworkMatrices.jl is registered?

jd-lara (approved):

PR #1039, opened 2024-01-09
Initial support for supplemental attributes

daniel-thom:


jd-lara:
we can create a PSY4 branch and merge the changes to that so that we can test adequately

src/base.jl
```diff
@@ -1258,6 +1271,104 @@ function transform_single_time_series!(sys::System, horizon::Int, interval::Date
...
+
+"""
+Returns an iterator of supplemental attributes. T can be concrete or abstract.
+Call collect on the result if an array is desired.
+
+# Examples
+```julia
+iter = get_supplemental_attributes(ForcedOutage, sys)
+iter = get_supplemental_attributes(Outage, sys)
+iter = get_supplemental_attributes(x -> x.forced_outage_rate >= 0.5, ForcedOutage, sys)
```
daniel-thom, line 1323: Do you have a better value to put in this example?

src/PowerSystems.jl
```diff
@@ -241,6 +248,7 @@ export ForecastCache
 export StaticTimeSeriesCache
 export NormalizationFactor
 export NormalizationTypes
+export TimeSeriesCounts
```
jd-lara, line 251: Do we need to export this?

jd-lara (approved):
The only details is wether we want to export time series counts

src/outages.jl
```diff
@@ -0,0 +1,63 @@
+abstract type Outage <: SupplementalAttribute end
```
jd-lara, line 1: @scdhulipala please confirm with @GordStephen that this structure is adequate for the purposes of integrating PSY into PRAS.

src/PowerSystems.jl
daniel-thom: We export `get_time_series_counts` which returns an instance of this struct. I don’t think a public function should return a private struct. We could argue whether that function should be exported.

src/outages.jl
GordStephen: As far as information needed, l think this looks good. Since there's multiple ways to store the same data, will there be a convention for which takes priority, or is that decision left to the consumer? E.g. FOR and MTTR won't necessarily be internally consistent with outage and recovery transition probabilities, or may be provided as time series vs static values. Also, seems like TimeSeriesContainer is pretty flexible as far as contents, will there be guidance around canonical names for the different possible time series stored within (presumably they'll be expected to match the struct property names)?

jd-lara: As usual in PSY the data is left to the user. We can decide to limit the fields to one or the other if the additional flexibility doesn't help. I'll proceed to merge this branch into the PSY4 release branch to make further changes in a forthcoming PR.

jd-lara:
partially addresses 

https://github.com/NREL-Sienna/PowerSystems.jl/issues/943

Issue #849, opened 2022-01-21
handle "PVe" categories in table data parser


claytonpbarrows:
Currently, we handle values of "PV", but not "PVe" which matches our enumerable prime_mover values. The table data parser should handle PVe as well. [see this issue](https://github.nrel.gov/SIIP/REEDS2SIIP/issues/11)

PR #1045, opened 2024-01-31
Fix case sensitivity in enum parsers

GabrielKS:
`parse_enum_mapping` was converting input to uppercase but the keys of the underlying dictionary were not all uppercase, so, e.g., `parse_enum_mapping(PrimeMovers, "PVe")` wouldn't work and https://github.com/NREL-Sienna/PowerSystems.jl/issues/849 wasn't truly solved. Now everything is `Unicode.normalize(casefold=true)`'d on both sides.

### DIFF
--- a/src/parsers/common.jl
+++ b/src/parsers/common.jl
@@ -19,28 +19,28 @@ merge!(
     ),
 )
 
-const STRING2PRIMEMOVER = Dict((string(x) => x) for x in instances(PrimeMovers))
+const STRING2PRIMEMOVER =
+    Dict((normalize(string(x); casefold = true) => x) for x in instances(PrimeMovers))
 merge!(
     STRING2PRIMEMOVER,
     Dict(
-        "W2" => PrimeMovers.WT,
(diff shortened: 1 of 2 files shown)


src/parsers/common.jl
```diff
@@ -19,28 +19,28 @@ merge!(
     ),
 )
 
-const STRING2PRIMEMOVER = Dict((string(x) => x) for x in instances(PrimeMovers))
+const STRING2PRIMEMOVER =
```
daniel-thom, line 22: The keys used to be all-caps...I guess we should have future-proofed ourselves in some way. Thanks. Do we need to do the same thing to `STRING2FUEL` now?

jd-lara: Probably yes

GabrielKS: Probably should, though those happen to all be uppercase so far

GabrielKS: Done

daniel-thom (approved):

PR #1047, opened 2024-02-02
Add support for subsystems

daniel-thom:
This requires InfrastructureSystems code from https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/326. So, CI will fail.

### DIFF
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -387,6 +387,20 @@ export get_renewable_unit
 export get_interconnection_rating
 export get_interconnection_impedance
 
+# Subsystems
+export add_subsystem!
+export get_subsystems
+export get_num_subsystems
+export remove_subsystem!
+export add_component_to_subsystem!
+export get_subsystem_components
+export remove_component_from_subsystem!
+export remove_component_from_subsystems!
+export has_component
+export get_assigned_subsystems
+export has_subsystems
+export is_assigned_to_subsystem
+
 export set_runchecks!
 export check
 export check_component
@@ -578,6 +592,7 @@ include("outages.jl")
 
 # Definitions of PowerSystem
 include("base.jl")
+include("subsystems.jl")
 include("data_format_conversions.jl")
 
 #Data Checks
--- a/src/subsystems.jl
+++ b/src/subsystems.jl
@@ -185,7 +185,7 @@ end
 
 function _check_num_subsystems(sys::System)
     num_buses = length(sys.bus_numbers)
-    if num_buses == get_num_subsystems(sys)
+    if get_num_subsystems(sys) >= num_buses
         throw(
             IS.InvalidValue(
                 "The number of subsystems cannot exceed the number of buses: $num_buses",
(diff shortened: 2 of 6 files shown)


src/PowerSystems.jl
```diff
@@ -387,6 +387,20 @@ export get_renewable_unit
 export get_interconnection_rating
 export get_interconnection_impedance
 
+# Subsystems
+export add_subsystem!
+export get_subsystems
+export get_num_subsystems
```
jd-lara, line 393: lets use a full word  here.

jd-lara (approved):
The changes look good to me and the functionality is what we can cover without the implementation in PowerSimulations. I think that we can merge this branch and work in subsequent PR's if new issues come to light. 

@tengis-nrl can also open PRs with the testing data to increase the confidence in this code.

tengis-nrl (approved):

src/PowerSystems.jl
daniel-thom: `get_number_of_subsystems` is quite verbose. We have several get_num_* functions already.
```
get_num_components
get_num_members
get_num_steps
get_num_supplemental_attributes
get_num_time_series
get_num_time_series_by_type
```
We do have this one also: `get_number_of_dimensions`.

src/subsystems.jl
```diff
@@ -0,0 +1,283 @@
...
+    _check_branch_consistency(sys, component)
+    _check_device_service_consistency(sys, component)
+    _check_subcomponent_consistency(sys, component)
+    _check_topological_consistency(sys, component)
+    return
+end
+
+function _check_num_subsystems(sys::System)
+    num_buses = length(sys.bus_numbers)
+    if num_buses == get_num_subsystems(sys)
```
tengis-nrl, line 188: Is it not necessary to use =< instead of ==?

src/PowerSystems.jl
jd-lara: ok fine

src/subsystems.jl
daniel-thom: Technically, the code is correct because this gets called at every addition of a subsystem. But your point is good and I will change it.

src/subsystems.jl
```diff
@@ -0,0 +1,283 @@
...
+    component::RegulationDevice,
+)
+    if is_assigned_to_subsystem(sys, component.device, subsystem_name)
+        remove_component_from_subsystem!(sys, subsystem_name, component.device)
+    end
+end
+
+"""
+Return a Generator of all components in the subsystem.
+
```
tengis-nrl, line 115: Are we missing some code here?

daniel-thom: Management of subsystems occurs in the InfrastructureSystems (IS) layer. We implement storage of components and related objects in this sub-package so that other packages can benefit from the logic. The real function is here: https://github.com/daniel-thom/InfrastructureSystems.jl/blob/subsystems/src/subsystems.jl#L74

tengis-nrl: Thank you!

PR #1052, opened 2024-02-14
Adding Retest

tengis-nrl:

Check the [contributor guidelines](https://nrel-sienna.github.io/PowerSystems.jl/stable/code_base_developer_guide/developer/)

3. If you are contributing a new struct: please refer to the [testing requirements for new structs](https://nrel-sienna.github.io/PowerSystems.jl/stable/code_base_developer_guide/adding_new_types/#Testing-the-addition-of-new-struct-to-the-code-base)


test/PowerSystemsTests.jl
tengis-nrl, line 1: Hi @GabrielKS, this file was added to use ReTest instead of the current testing package. There is a file called InfrastructureSystemsTests.jl in IS that has almost the same code as this one. @daniel-thom was wondering how we can avoid code duplication. Specifically, he asked "Is there a way to avoid that without making the ReTest.jl package a dependency of PSY? Keep in mind that there are dozens of Sienna packages that should use this test pattern." Any suggestion?

test/load_tests.jl
```diff
@@ -0,0 +1,13 @@
+using Revise
```
GabrielKS, line 1: Copied from the IS PR: Seems like even if the best practice is to install Revise globally, if our tests are going to depend it we ought to have it in `test/Project.toml`

test/PowerSystemsTests.jl
GabrielKS: Dan and I talked about this a while ago and the options I see are:

1. Put the stuff that would be duplicated into InfrastructureSystems.jl. Advantage: no massive code duplication across Sienna repos, just call the relevant thing from IS. Disadvantage: ReTest.jl would need to be a non-test dependency of IS (and therefore, transitively, of all the other Sienna packages)
2. Current approach: reimplement for every repository. Advantage: ReTest.jl can be just a test dependency. Disadvantage: massive code duplication.
3. Create a new package that is just for testing, put the stuff that would be duplicated there. Advantages: no massive duplication and ReTest is just a test dependency. Disadvantage: we have to maintain a whole separate package for this.
4. Same as 3 but with a submodule or something instead of a package. Same advantages as 3. Disadvantage: it's really annoying to work with submodules.

I don't think we're going to come up with anything that doesn't have some of those disadvantages. I might not fully appreciate the costs of maintaining a whole separate package, but option 3 looks the most attractive to me. We could also do something like this to remove the duplication of all our formatting scripts.

test/PowerSystemsTests.jl
```diff
@@ -0,0 +1,107 @@
...
+            @info IS.report_log_summary(multi_logger)
+        end
+    finally
+        # Guarantee that the global logger is reset.
+        global_logger(logger)
+        nothing
+    end
+end
+
+
```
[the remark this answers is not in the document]
GabrielKS: @tengis-nrl you should learn how to invoke the formatter set up in each Sienna repository so you can deal with these issues proactively. I can help you with that sometime if you need

GabrielKS (changes requested):

test/PowerSystemsTests.jl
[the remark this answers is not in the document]
tengis-nrl: How about during the next time we meet? Thanks!

test/PowerSystemsTests.jl
jd-lara: For the time being we will keep 2. as the option since it also creates a headache to maintain package dependencies to do any of the others.

test/load_tests.jl
jd-lara: remove, we need to remove from IS too.

daniel-thom:
I like option 3 the best. Good point about the formatter scripts. Hopefully, this won’t change often. Also, the formatter version would be fixed in this package.

test/load_tests.jl
daniel-thom: This file is optional. Users don’t have to call it. I say leave it the way it is. The instructions should advise you to install TestEnv.jl and Revise.jl globally.

jd-lara:
> I like option 3 the best. Good point about the formatter scripts. Hopefully, this won’t change often. Also, the formatter version would be fixed in this package.

@pesap suggested adding a modified GitHub action that can commit the required formatter changes.

test/load_tests.jl
GabrielKS: Okay, that works for me

jd-lara:
@daniel-thom do you want to merge this or go ahead with the creation of the testing package?

daniel-thom:
> @daniel-thom do you want to merge this or go ahead with the creation of the testing package?

I think we should create `SiennaDevUtils.jl`, add these test functions and the formatter, and then change IS and PSY to use it.

jd-lara:
closing since the preferred solution of creating a testing package will be implemented.

PR #1056, opened 2024-02-16
Variable Cost Refactor Part 1: Function Data

GabrielKS:
The first part of a significant refactor of Sienna cost data structures. Relates to https://github.com/NREL-Sienna/PowerSystems.jl/issues/934, https://github.com/NREL-Sienna/PowerSystems.jl/issues/985, https://github.com/NREL-Sienna/PowerSystems.jl/issues/1007.

At the time of draft PR creation, all tests pass except for these:
  * `Test field validation` in `test_validation.jl`: an issue unrelated to these changes, a fix will become available on `psy4` soon and I'll rebase onto that
  * Several tests in `test_cost_functions.jl`: waiting on a decision about whether to refactor aspects of InfrastructureSystems (e.g., the `Deterministic` struct)

…provided that [this commit](https://github.com/NREL-Sienna/PowerSystemsTestData/tree/7910da87f7749f1baa17286315542bc593e16605) of PowerSystemsTestData is used (by changing the PowerSystemsCaseBuilder `DATA_DIR` to a local clone of that).

This is not quite ready for full review but I'm putting it out here to track progress.

PR #1057, opened 2024-02-20
Add conversion logic for subsystems and attributes

GabrielKS:
A quick check during deserialization to add empty `subsystems` and `attributes` fields to pre-PSY4 systems


GabrielKS:
I thought it would make sense to make this change separately from my cost function changes, but maintaining the two separate feature branches is getting too complicated. Closing this; the changes are incorporated into https://github.com/NREL-Sienna/PowerSystems.jl/pull/1056/.

PR #1060, opened 2024-02-27
Update CI for psy4 feature branches

daniel-thom:


.github/workflows/manifest_check.yml
```diff
@@ -0,0 +1,20 @@
+name: Test-CI
```
jd-lara, line 1:
```suggestion
name: Check Manifest
```

.github/workflows/manifest_check.yml
jd-lara, line 1: We probably don't need this check on every single OS

.github/workflows/manifest_check.yml
```diff
@@ -0,0 +1,20 @@
...
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    branches:
+      - 'main'
```
jd-lara, line 7: to test the action may be add psy4 once just to make sure it works

.github/workflows/manifest_check.yml
```diff
@@ -0,0 +1,20 @@
...
+
+jobs:
+  test:
+    steps:
+      - uses: actions/checkout@v3
```
jd-lara, line 12: do we need the checkout action for this?

jd-lara (commented):
Can we close this PR?

Issue #1032, opened 2023-12-12
StandardLoad parser broken 


jd-lara:
The PSSe parser for standard loads is broken if only the constant impedance portions are populated

PR #1065, opened 2024-03-03
Repair StandardLoad parsing from .mat/.raw files

HaleyRoss:
Removed if statement from the read_loads!() function that caused StandardLoads being read from .mat and .raw files to be dropped if the constant_active_power and constant_reactive_power arguments were 0. 

This is correcting issue #1032


src/parsers/power_models_data.jl
```diff
@@ -485,7 +483,7 @@ function make_thermal_gen(gen_name::AbstractString, d::Dict, bus::ACBus, sys_mba
     thermal_gen = ThermalStandard(;
         name = gen_name,
         status = Bool(d["gen_status"]),
-        available = true,
+        available = Bool(d["gen_status"]),
```
jd-lara, line 486: good catch

jd-lara (approved):

PR #1080, opened 2024-03-19
Fix component-is-attached-to-system check

### LINKED ISSUE #1016: Give error or warning when assigning services to a device from another system
Currently, when assigning a services to a device from another system it does not give any warning or error.

```
sys1 = build_system("sys1")
sys2 = build_system("sys2")
device2 = first(get_component(ThermalStandard, sys2))
for service1 in get_components(Reserve, sys1)
    add_service!(device2, service1, sys2)
end
```

daniel-thom:
Fixes #1016. The check for whether a component is attached to a system was inadequate. It checked whether a component of the given type and name was attached to the system. If you built two systems with the same components and names, that check would incorrectly pass. This PR fixes that bug.

There were numerous tests that only passed because of this same-type-and-name comparison. This PR also fixes those tests.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -180,8 +180,24 @@ function System(
     services = nothing,
     kwargs...,
 )
+    for component in Iterators.flatten((generators, loads))
+        if get_name(component) == "init"
+            set_bus!(component, first(buses))
+        end
+    end
     _services = isnothing(services) ? [] : services
-    return System(base_power, buses, generators, loads, _services; kwargs...)
+    _branches = isnothing(branches) ? [] : branches
+    _storage = isnothing(storage) ? [] : storage
+    return System(
+        base_power,
+        buses,
+        generators,
+        loads,
+        _branches,
+        _storage,
+        _services;
+        kwargs...,
+    )
 end
 
 """Constructs a System from a file path ending with .m, .RAW, or .json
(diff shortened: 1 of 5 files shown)


src/base.jl
```diff
@@ -180,8 +180,24 @@ function System(
...
     kwargs...,
 )
+    for component in Iterators.flatten((generators, loads))
+        if get_name(component) == "init"
+            set_bus!(component, first(buses))
+        end
+    end
     _services = isnothing(services) ? [] : services
-    return System(base_power, buses, generators, loads, _services; kwargs...)
+    _branches = isnothing(branches) ? [] : branches
```
daniel-thom, line 189: Apparently, the demo constructor never added branches or services...

src/base.jl
```diff
@@ -1049,11 +1065,9 @@ end
...
 """
 function is_attached(component::T, sys::System) where {T <: Component}
-    return is_attached(T, get_name(component), sys)
-end
-
-function is_attached(::Type{T}, name, sys::System) where {T <: Component}
-    return !isnothing(get_component(T, sys, name))
+    existing_component = get_component(T, sys, get_name(component))
+    isnothing(existing_component) && return false
+    return component === existing_component
```
daniel-thom, line 1070: This is the fix.

jd-lara (approved):

PR #1088, opened 2024-04-03
Add new parsing capabilities for DERA1

rodrigomha:
This is a first idea on how to handle DynamicInjectors parsing on top of components.


jd-lara (commented):
we probably need a test for this

jd-lara (approved):

PR #1089, opened 2024-04-10
Construct a system from a subsystem

### LINKED ISSUE #1086: Support export of subsystem to a new system
Using subsystems capability to export a sub set of the data

daniel-thom:
Fixes #1086

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -274,11 +274,11 @@ function _post_deserialize_handling(sys::System; runchecks = true, assign_new_uu
     if assign_new_uuids
         IS.assign_new_uuid!(sys)
         for component in get_components(Component, sys)
-            assign_new_uuid!(sys, component)
+            IS.assign_new_uuid!(sys, component)
         end
         for component in
             IS.get_masked_components(InfrastructureSystemsComponent, sys.data)
-            assign_new_uuid!(sys, component)
+            IS.assign_new_uuid!(sys, component)
         end
         # Note: this does not change UUIDs for time series data because they are
         # shared with components.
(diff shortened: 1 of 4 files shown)


src/base.jl
```diff
@@ -2098,13 +2152,27 @@ function handle_component_removal!(sys::System, device::Device)
     return
 end
 
+function handle_component_removal!(sys::System, component::RegulationDevice)
```
daniel-thom, line 2155: We had two existing bugs where we were not removing masked components from RegulationDevice or StaticInjectionSubsystem.

src/subsystems.jl
```diff
@@ -240,8 +240,8 @@ function _check_topological_consistency(sys::System, component::Component)
                 sys,
                 component,
                 val,
-                "A component must be assigned to the same subsystems as its topological component(s).";
-                symmetric_diff = true,
+                "A component must be assigned to at least one subsystem as its topological component(s).";
```
daniel-thom, line 243: This check was not correct for a valid test system, but the code was not being exercised.

jd-lara: I wonder why this test system was working

src/base.jl
```diff
@@ -274,11 +274,11 @@ function _post_deserialize_handling(sys::System; runchecks = true, assign_new_uu
     if assign_new_uuids
         IS.assign_new_uuid!(sys)
         for component in get_components(Component, sys)
-            assign_new_uuid!(sys, component)
+            IS.assign_new_uuid!(sys, component)
```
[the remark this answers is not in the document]
jd-lara: This is ok, I don't think we should export this on PSY.

src/base.jl
```diff
@@ -319,6 +319,60 @@ function System(sys_file::AbstractString, dyr_file::AbstractString; kwargs...)
...
 
+"""
+Construct a System from a subsystem of an existing system.
+"""
+function from_subsystem(sys::System, subsystem::AbstractString; runchecks = true)
+    if !in(subsystem, get_subsystems(sys))
+        error("subsystem = $subsystem is not stored")
+    end
+
+    # It would be faster to create an empty system and then populate it with
```
jd-lara, line 330: The "real test" to this will be on the EI data set.

jd-lara (approved):

test/test_subsystems.jl
```diff
@@ -70,7 +70,8 @@ function create_system_with_2_test_subsystems()
         subsystems[2],
         get_component(ACBus, c_sys5, "nodeC"),
     )
-
+    PSY.check(c_sys5)
+    PSY.check_components(c_sys5)
```
daniel-thom, line 74: The tests were passing because the checks were not being run.

PR #1090, opened 2024-04-19
Support time series redesign in InfrastructureSystems

daniel-thom:
This PR is tied to the changes in https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/349.

### DIFF
--- a/docs/src/tutorials/basics.md
+++ b/docs/src/tutorials/basics.md
@@ -192,7 +192,7 @@ get_forecast_initial_times(sys)
 We can find the names of all time series attached to a component:
 
 ```@repl basics
-ts_names = get_time_series_names(Deterministic, loads[1])
+show_time_series(loads[1])
 ```
 
 We can access a specific time series for a specific component:
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -335,7 +335,6 @@ export get_horizon
 export get_forecast_initial_times
 export list_time_series_info
 export show_time_series
-#export get_forecast_total_period
 export get_resolution
 export get_data
 export iterate_components
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -264,6 +264,8 @@ export Deterministic
 export Probabilistic
 export SingleTimeSeries
 export DeterministicSingleTimeSeries
+export StaticTimeSeriesInfo
+export ForecastInfo
 export Scenarios
 export ForecastCache
 export StaticTimeSeriesCache
@@ -321,10 +323,9 @@ export has_supplemental_attributes
 export iterate_supplemental_attributes
 export get_time_series
 export get_time_series_array
-export get_time_series_resolution
+export list_time_series_resolutions
 export get_time_series_timestamps
 export get_time_series_values
-export get_time_series_names
 export get_time_series_counts
 export get_scenario_count
 export get_percentiles
@@ -332,7 +333,9 @@ export get_next_time_series_array!
 export get_next_time
 export get_horizon
 export get_forecast_initial_times
-export get_forecast_total_period
+export list_time_series_info
+export show_time_series
+#export get_forecast_total_period
 export get_resolution
 export get_data
 export iterate_components
@@ -471,7 +474,9 @@ import InfrastructureSystems:
     Deterministic,
     Probabilistic,
     SingleTimeSeries,
+    StaticTimeSeriesInfo,
     DeterministicSingleTimeSeries,
+    ForecastInfo,
     Scenarios,
     ForecastCache,
     StaticTimeSeriesCache,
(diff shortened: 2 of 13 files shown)


docs/src/tutorials/basics.md
```diff
@@ -192,7 +192,7 @@ get_forecast_initial_times(sys)
 We can find the names of all time series attached to a component:
 
 ```@repl basics
-ts_names = get_time_series_names(Deterministic, loads[1])
+show_time_series(loads[1])
```
daniel-thom, line 195: I want to emphasize the fact that I'm proposing this to be the way users view time series for a component. I have removed time series display from `Base.show(component)`. I can make a good case for this, but you may disagree and we can talk about it.

src/PowerSystems.jl
```diff
@@ -321,18 +323,19 @@ export has_supplemental_attributes
...
-export get_time_series_names
 export get_time_series_counts
 export get_scenario_count
 export get_percentiles
 export get_next_time_series_array!
 export get_next_time
 export get_horizon
 export get_forecast_initial_times
-export get_forecast_total_period
+export list_time_series_info
```
daniel-thom, line 336: This is a new way of finding time series information programmatically without actually reading the data from the .h5. (I don't love the name, but couldn't think of anything else.) There is also `list_time_series_metadata`, which I intentionally did not re-export here. Similar to the prior implementation where you could view the metadata inside `component.time_series_metadata`, it is not helpful to users because it has `*Metadata` struct names.

src/base.jl
```diff
@@ -1318,7 +1308,7 @@ get_forecast_initial_times(sys::System) = IS.get_forecast_initial_times(sys.data
 """
 Return the total period covered by all forecasts.
 """
-get_forecast_total_period(sys::System) = IS.get_forecast_total_period(sys.data)
+#get_forecast_total_period(sys::System) = IS.get_forecast_total_period(sys.data)
```
daniel-thom, line 1311: Let me know if I need to restore this. It is not used in PSI.

src/models/static_injection_subsystem.jl
```diff
@@ -20,8 +20,8 @@ function copy_subcomponent_time_series!(
     # the same type with the same time series type and label. Currently in the HybridSystem
     # use case there can only be one devoe of each type.
     existing_ts = Set(
-        (typeof(ts), get_name(ts_m)) for
-        (ts, ts_m) in IS.get_time_series_with_metadata_multiple(subsystem)
+        (typeof(ts), get_name(ts)) for
+        ts in IS.get_time_series_multiple(subsystem)
```
daniel-thom, line 24: The method on the left was never needed.

test/test_hybrid_system.jl
```diff
@@ -64,18 +64,22 @@ end
...
+        collect(get_time_series_multiple(h_sys; type = DeterministicSingleTimeSeries))
+    @test length(sts) == 2
+    @test length(forecasts) == 2
+    @test issubset((get_name(x) for x in sts), expected_time_series_names)
+    @test issubset((get_name(x) for x in forecasts), expected_time_series_names)
 
     @test get_time_series(SingleTimeSeries, electric_load, "max_active_power") isa
           SingleTimeSeries
-    @test get_time_series(AbstractDeterministic, electric_load, "max_active_power") isa
+    @test get_time_series(Deterministic, electric_load, "max_active_power") isa
```
daniel-thom, line 77: Just calling out the loss of this functionality. Like I said in the IS PR, I'm open to restoring it but would prefer not to. I am hoping that users can view what's available (show_time_series, list_time_series_info) and then ask for exactly what they want.

src/base.jl
jd-lara: if at this rate we haven't used it, it isn't needed.

jd-lara (approved):

src/PowerSystems.jl
```diff
@@ -321,18 +323,19 @@ export has_supplemental_attributes
...
+export show_time_series
+#export get_forecast_total_period
```
GabrielKS, line 338: Should this be removed?

test/test_hybrid_system.jl
GabrielKS: Let's restore later along with the restoration in IS we discussed

test/test_system.jl
```diff
@@ -315,13 +315,13 @@ end
...
-    @test get_time_series_resolution(sys) == resolution
+    @test list_time_series_resolutions(sys)[1] == resolution
     @test get_forecast_horizon(sys) == horizon
     @test get_forecast_initial_timestamp(sys) == initial_time
     @test get_forecast_interval(sys) == Dates.Millisecond(second_time - initial_time)
     @test get_forecast_window_count(sys) == 2
-    @test Dates.Hour(get_forecast_total_period(sys)) ==
-          Dates.Hour(second_time - initial_time) + Dates.Hour(resolution * horizon)
+    #@test Dates.Hour(get_forecast_total_period(sys)) ==
+    #Dates.Hour(second_time - initial_time) + Dates.Hour(resolution * horizon)
```
GabrielKS, line 324: Should this be uncommented/removed?

GabrielKS (approved):
Looks good aside from a few minor cleanup things

PR #1101, opened 2024-05-08
Fuel Cost Interface, Including Time Series

GabrielKS:
This PR implements an interface to get at the fuel cost fields within the new cost structs. I took inspiration from the existing variable cost time series interface for `MarketBidCost`. Depends on https://github.com/NREL-Sienna/PowerSystems.jl/pull/1100, that should be merged first.

### DIFF
--- a/src/models/cost_function_timeseries.jl
+++ b/src/models/cost_function_timeseries.jl
@@ -158,6 +158,50 @@ function set_variable_cost!(
     return
 end
 
+function _process_fuel_cost(::Component, fuel_cost::Float64, start_time, len)
+    isnothing(start_time) && isnothing(len) && return fuel_cost
+    throw(ArgumentError("Got time series start_time and/or len, but fuel cost is a scalar"))
+end
+
+function _process_fuel_cost(component::Component, ts_key::TimeSeriesKey, start_time, len)
+    ts = get_time_series(component, ts_key, start_time, len)
+    return get_time_series_array(component, ts, start_time; len = len)
+end
+
+"Get the fuel cost of the component's variable cost, which must be a `FuelCurve`."
+function get_fuel_cost(component::StaticInjection;
+    start_time::Union{Nothing, Dates.DateTime} = nothing,
+    len::Union{Nothing, Int} = nothing)
+    op_cost = get_operation_cost(component)
+    var_cost = get_variable(op_cost)
+    !(var_cost isa FuelCurve) && throw(ArgumentError("Variable cost of type $(typeof(var_cost)) cannot represent a fuel cost, use FuelCurve instead"))
+    return _process_fuel_cost(component, get_fuel_cost(var_cost), start_time, len)
+end
+
+function _set_fuel_cost!(component::StaticInjection, fuel_cost)
+    op_cost = get_operation_cost(component)
+    var_cost = get_variable(op_cost)
+    !(var_cost isa FuelCurve) && throw(ArgumentError("Variable cost of type $(typeof(var_cost)) cannot represent a fuel cost, use FuelCurve instead"))
+    new_var_cost = FuelCurve(get_value_curve(var_cost), get_power_units(var_cost), fuel_cost)
+    set_variable!(op_cost, new_var_cost)
+end
+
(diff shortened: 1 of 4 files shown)


GabrielKS:
@jd-lara the existing `get_variable_cost` function in `cost_function_timeseries.jl` for `MarketBidCost` calls `get_time_series_array`, whereas it makes more sense to me to just return the time series itself, since that's what the user must pass to the setter. I have copied this interface for the fuel cost interface, but if you agree we can change both getters to return the time series.

jd-lara:
> @jd-lara the existing `get_variable_cost` function in `cost_function_timeseries.jl` for `MarketBidCost` calls `get_time_series_array`, whereas it makes more sense to me to just return the time series itself, since that's what the user must pass to the setter. I have copied this interface for the fuel cost interface, but if you agree we can change both getters to return the time series.

What do you mean `the time series itself`?

GabrielKS:
> What do you mean `the time series itself`?

A (subtype of) `TimeSeriesData` rather than a `TimeArray`.

After our DM discussion, I think the best course of action is to keep this as it is (asymmetric but convenient interface) for now and then when I update the market bid cost time series interface I can make a more informed decision. If it ends up making sense to make that one symmetric, I'll change this one to match.

src/models/cost_function_timeseries.jl
```diff
@@ -158,6 +158,59 @@ function set_variable_cost!(
     return
 end
 
+function _process_fuel_cost(::Component, fuel_cost::Float64, start_time, len)
```
jd-lara, line 161:
```suggestion
function _process_fuel_cost(::Component, fuel_cost::Float64, start_time::Dates.DateTime, len::Int)
```

src/models/cost_functions/variable_cost.jl
```diff
@@ -47,8 +47,8 @@ The default units for the x-axis are megawatts and can be specified with `power_
     value_curve::T
     "The units for the x-axis of the curve; defaults to natural units (megawatts)"
     power_units::UnitSystem = UnitSystem.NATURAL_UNITS
-    "Either a fixed value for fuel cost or the name of a fuel cost time series name"
-    fuel_cost::Union{Float64, String}
+    "Either a fixed value for fuel cost or the name of a fuel cost time series"
```
jd-lara, line 50:
```suggestion
    "Either a fixed value for fuel cost or the key to a fuel cost time series"
```

test/test_cost_functions.jl
```diff
@@ -167,16 +168,15 @@ test_costs = Dict(
     end
 end
 
-@testset "Test MarketBidCost with PiecewiseLinearData Cost Timeseries" begin
+@testset "Test MarketBidCost with PiecewiseLinearData Cost Timeseries with Service Forecast" begin
```
jd-lara, line 171:
```suggestion
@testset "Test MarketBidCost with PiecewiseLinearData Cost Timeseries with Service Bid Forecast" begin
```

jd-lara (commented):
@rodrigomha for visibility into the changes in PSI take a look at this PR

jd-lara (approved):
LGTM we can merge to start testing in PSI

jd-lara:
> > What do you mean `the time series itself`?
> 
> A (subtype of) `TimeSeriesData` rather than a `TimeArray`.
> 
> After our DM discussion, I think the best course of action is to keep this as it is (asymmetric but convenient interface) for now and then when I update the market bid cost time series interface I can make a more informed decision. If it ends up making sense to make that one symmetric, I'll change this one to match.

Lets add this to the list of pending items

src/models/cost_function_timeseries.jl
GabrielKS: Addressed, though I think in a more flexible manner

src/models/cost_functions/variable_cost.jl
GabrielKS: Done

test/test_cost_functions.jl
GabrielKS: Done

PR #1103, opened 2024-05-09
Renovate time series functionality for `MarketBidCost`

GabrielKS:
Depends on https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/367. Here we:

- Add a validation function `is_market_bid_curve` and factory function `make_market_bid_curve` to deal with market bid curves, which are `CostCurve{PiecewiseIncrementalCurve}`s with `NaN` first x-coordinate and initial input.
- Make the market bid cost time series return time arrays of these objects, even though they must be constructed with time series of `FunctionData`.
- Rewrite `cost_function_timeseries.jl` so that less logic has to be duplicated for every new field that works this way
- Make `no_load_cost` and `start_up` fields of `MarketBidCost` work this way


jd-lara (approved):
@GabrielKS we can merge and add methods as needed during the PSI integration

PR #1111, opened 2024-05-24
Docstrings

kdayday:
This is the second part of PSY docstrings updates:

- Changes are focused on src/descriptors/power_system_structs.json to fill gaps and update the top-level docstrings and argument descriptions in the Model Library
- Optional and "do not modify" arguments are identified
- A glossary has been added in the Modeler Guide and some cross-references have been added. There's more we could do there or reformat, particularly with the dynamics acronyms, and some terms are still undefined. 
- ~~HybridSystem isn't added to the documentation yet (I can't remember if someone else was working on that)~~ Added HybridSystem docs
- Made cost documentation visible in the Modeler Guide and Model Library, but haven't resolved any to-dos left by Gabriel
- Docs build is attached
- Edit: Includes renames to RenewableNonDispatch, ProductionVariableCostCurve, ConstantReserve, ConstantReserveGroup, ConstantReserveNonSpinning, and moves ControllableLoad to be a StaticLoad

[build_PSY.zip](https://github.com/NREL-Sienna/PowerSystems.jl/files/15440293/build_PSY.zip)


src/models/generated/StandardLoad.jl
```diff
@@ -28,25 +28,29 @@ This file is auto-generated. Do not edit.
...
-- `impedance_reactive_power::Float64`: (optional)
-- `current_active_power::Float64`: (optional)
-- `current_reactive_power::Float64`: (optional)
-- `max_constant_active_power::Float64`
-- `max_constant_reactive_power::Float64`: (optional)
-- `max_impedance_active_power::Float64`: (optional)
-- `max_impedance_reactive_power::Float64`: (optional)
-- `max_current_active_power::Float64`: (optional)
-- `max_current_reactive_power::Float64`: (optional)
+- `constant_active_power::Float64`: (optional) Constant active power demand in MW (P_P)
```
jd-lara, line 42: None of this are optional. The value 0.0 is a valid quantity

kdayday: Optional isn't meant to mean that 0.0 isn't valid. We're using "optional" for everything that has a valid default value, but the user could change it to some other valid value if they want. It's just not something you have to define, because you could leave it at 0.0.

If there's no default, it's not optional and you have to define it. 

If it has a default value, but it really shouldn't be changed (like an expression calculated from other input parameters), we've labeled it "Do not modify."

daniel-thom: `Optional` can be a confusing term. In some programming languagues `Optional` in the context of a type means that a value can be null. Julia doesn't have that, but we use the concept in this package.

We could sidestep this question by auto-generating a string with the default value. For example, this string could be generated as
```
- `constant_active_power::Float64`: (default=0.0) Constant active power demand in MW (P_P)
```

kdayday: Jose and I had also talked about eventually auto-generating the default value. If we can do that now, I think that would be the best solution, and then we can just delete all the optionals.

jd-lara: @daniel-thom  It's a change in the template in IS we can do to always place the default.

docs/src/model_library/cost_curves.md
```diff
@@ -0,0 +1,18 @@
+### Cost Curves
+```@docs
+CostCurve
+FuelCurve
+LinearCurve
```
GabrielKS, line 5: Since `CostCurve` and `FuelCurve` are composed of `LinearCurve`, `QuadraticCurve`, etc., it might be confusing to put them under the same heading.

src/models/generated/StandardLoad.jl
kdayday: Rodrigo and I added a print out of the default value in IS. https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/374 

Example for StandardLoad:
Screenshot 2024-06-04 at 2 45 14 PM

docs/src/model_library/cost_curves.md
kdayday: Resolved by https://github.com/NREL-Sienna/PowerSystems.jl/pull/1111/commits/212fb92018dc8cedb85f02c8654dcb9a4990a6ee

src/models/generated/StandardLoad.jl
kdayday: Removed manual optional in https://github.com/NREL-Sienna/PowerSystems.jl/pull/1111/commits/a23061991537da8536d887fed5ae169bdc76e05b

rodrigomha (approved):

Issue #1081, opened 2024-03-19
Support a system serialization to JSON when MarketBidCost has services


amirmm11:
Error description:

```
compression=CompressionSettings(false, CompressionTypes.DEFLATE = 1, 3, true)
ERROR: LoadError: MethodError: Cannot `convert` an object of type Dict{String, Any} to an object of type Service

Closest candidates are:
  convert(::Type{T}, ::T) where T
   @ Base Base.jl:64

Stacktrace:
  [1] setindex!(A::Vector{Service}, x::Dict{String, Any}, i1::Int64)
    @ Base ./array.jl:969
  [2] _unsafe_copyto!(dest::Vector{Service}, doffs::Int64, src::Vector{Any}, soffs::Int64, n::Int64)
    @ Base ./array.jl:250
  [3] unsafe_copyto!
    @ ./array.jl:304 [inlined]
  [4] _copyto_impl!
    @ ./array.jl:327 [inlined]
  [5] copyto!
    @ ./array.jl:314 [inlined]
  [6] copyto!
    @ ./array.jl:339 [inlined]
  [7] copyto_axcheck!
    @ ./abstractarray.jl:1180 [inlined]
  [8] Vector{Service}(x::Vector{Any})
    @ Base ./array.jl:621
  [9] convert
    @ ./array.jl:613 [inlined]
 [10] MarketBidCost
    @ ~/.julia/packages/PowerSystems/mjN6j/src/models/generated/MarketBidCost.jl:27 [inlined]
 [11] #MarketBidCost#16
    @ ~/.julia/packages/PowerSystems/mjN6j/src/models/generated/MarketBidCost.jl:41 [inlined]
 [12] deserialize_struct(#unused#::Type{MarketBidCost}, data::Dict{String, Any})
    @ InfrastructureSystems ~/.julia/packages/InfrastructureSystems/9btGb/src/serialization.jl:197
 [13] deserialize(#unused#::Type{MarketBidCost}, data::Dict{String, Any})
    @ InfrastructureSystems ~/.julia/packages/InfrastructureSystems/9btGb/src/serialization.jl:159
 [14] deserialize_uuid_handling(field_type::Type{MarketBidCost}, val::Dict{String, Any}, component_cache::Dict{Base.UUID, Component})
    @ PowerSystems ~/.julia/packages/PowerSystems/mjN6j/src/models/serialization.jl:118
[...]

PR #1114, opened 2024-05-30
Fix issue 1081, fix bug from PR 1112

GabrielKS:
This small PR:

- Fixes https://github.com/NREL-Sienna/PowerSystems.jl/issues/1081 by using existing `Component` serialization code to ensure that the vector of `Service`s in `MarketBidCost` gets serialized as UUIDs, adds a test for serialization of `MarketBidCost` with `Service`s
- Fixes a variable cost constructor ambiguity introduced in https://github.com/NREL-Sienna/PowerSystems.jl/pull/1112 (@rodrigomha for visibility; can you verify that this solution works for you?)

### DIFF
--- a/src/models/cost_functions/operational_cost.jl
+++ b/src/models/cost_functions/operational_cost.jl
@@ -1,5 +1,6 @@
-""" Super type for operational cost representation in the model"""
+""" Supertype for operational cost representation in the model"""
 abstract type OperationalCost <: DeviceParameter end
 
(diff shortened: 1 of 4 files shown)


jd-lara (changes requested):
The changes look ok to me. Please just run the formatter

daniel-thom (approved):

jd-lara (approved):

PR #1117, opened 2024-06-03
Jd/parser speed tests

jd-lara:
These are some speed improvements to the PSSe parser. This PR brings down the time to parse the EasterInterconnection raw file from ~500 seconds to ~30 seconds.


src/parsers/pm_io/psse.jl
```diff
@@ -21,29 +21,33 @@ function _init_bus!(bus::Dict{String, Any}, id::Int)
...
+    for bus in pm_bus_data
+        if bus["index"] == bus_i
+            return bus[field]
+        end
+    end
+    @info("Could not find bus $bus_i, returning 0 for field $field")
+    return 0
+end
+
+function _get_bus_value(bus_i::Int, field::String, pm_bus_data::Dict{String, Any})
```
daniel-thom, line 34: The name `pm_bus_data` seems misleading. Should it be `pm_data` for consistency?

src/parsers/pm_io/psse.jl
```diff
@@ -21,29 +21,33 @@ function _init_bus!(bus::Dict{String, Any}, id::Int)
...
-            if bus["index"] == bus_i
-                return bus[field]
-            end
-        end
-    end
-
-    @info("Could not find bus $bus_i, returning 0 for field $field")
-    return 0
+function _get_bus_value(bus_i::Int, field::String, pm_data)
+    return _get_bus_value(bus_i, field, pm_data["bus"])
```
daniel-thom, line 50: This only calls the Array version, right?

src/parsers/power_models_data.jl
```diff
@@ -273,52 +287,55 @@ function read_loads!(sys::System, data, bus_number_to_bus::Dict{Int, ACBus}; kwa
...
+    zone_bus_map = Dict{Int, Vector}()
+    for (_, bus) in data["bus"]
         push!(zones, bus["zone"])
+        push!(get!(zone_bus_map, bus["zone"], Vector()), bus)
     end
 
+    load_zone_map =
+        Dict{Int, Dict{String, Float64}}(i => Dict("pd" => 0.0, "qd" => 0.0) for i in zones)
+    for (key, load) in data["load"]
+        zone = data["bus"][load["load_bus"]]["zone"]
```
daniel-thom, line 316: You could remove quite a few dictionary lookups.
```suggestion
        zone = load_zone_map[data["bus"][load["load_bus"]]["zone"]]
```
And then perform the sums together.

daniel-thom (commented):
Nice work!

src/parsers/pm_io/psse.jl
jd-lara: I just named it like that since it is only the bus dict not the whole data

src/parsers/pm_io/psse.jl
jd-lara: this dispatches depending on wether `pm_data["bus"]` is an Array or a Dict.

jd-lara:
looks like I broke several tests so still need to fix those before merging

src/parsers/pm_io/psse.jl
daniel-thom: It is the full pm_data, right? You have `pm_bus_data["bus"]`.

src/parsers/pm_io/psse.jl
daniel-thom: Perhaps you could make two different function names: one that takes the full pm_data and one that dispatches on the type of the bus container.

jd-lara:
image

src/parsers/pm_io/psse.jl
jd-lara: Changes to `_find_bus_value`

jd-lara:
image

src/parsers/power_models_data.jl
```diff
@@ -98,10 +98,23 @@ function _get_pm_dict_name(device_dict)
...
 
+function _get_pm_bus_name(device_dict::Dict, unique_names::Bool)
+    if haskey(device_dict, "name")
+        if unique_names
+            name = strip(device_dict["name"])
+        else
+            name = strip(device_dict["name"]) * "_" * string(device_dict["bus_i"])
+        end
+    else
+        name = strip(join(string.(device_dict["source_id"]), "-"))
```
daniel-thom, line 109: Is it possible and practical to always use this name? I assume that it will be easier for everyone to understand if this parser always produces bus names in the same way.

src/parsers/power_models_data.jl
```diff
@@ -170,17 +189,31 @@ function read_bus!(sys::System, data::Dict; kwargs...)
...
-    _get_name = get(kwargs, :bus_name_formatter, _get_pm_dict_name)
+    default_bus_naming = x -> _get_pm_bus_name(x, unique_bus_names)
+
+    _get_name = get(kwargs, :bus_name_formatter, default_bus_naming)
     for (i, (d_key, d)) in enumerate(bus_data)
         # d id the data dict for each bus
         # d_key is bus key
-        d["name"] = get(d, "name", string(d["bus_i"]))
         bus_name = strip(_get_name(d))
         bus_number = Int(d["bus_i"])
```
daniel-thom, line 219: The error message on line 230 could be confusing. If this code and its assumptions are correct, the only way for the error to occur is if the user provided a custom bus_formatter function that did not produce unique names.

src/parsers/power_models_data.jl
daniel-thom: Why not take this suggestion?

src/parsers/power_models_data.jl
jd-lara: I just updated the error message

daniel-thom (approved):

PR #1121, opened 2024-06-11
Regenerate structs

GabrielKS:
The changes to the template in https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/374 imply these changes to the docstrings.


jd-lara:
isn't this done in #1119

GabrielKS:
> isn't this done in #1119

It does appear so. Oops.

PR #1122, opened 2024-06-11
add branch name to logger

jd-lara:


src/utils/IO/branchdata_checks.jl
```diff
@@ -88,7 +88,7 @@ function calculate_thermal_limits!(branch, basemva::Float64)
         is_valid = false
 
     elseif get_rate(branch) == 0.0
-        @warn "Data for line rating is not provided, PowerSystems will infer a rate from line parameters" maxlog =
+        @warn "Data for branch $(get_name(branch)) rating is not provided, PowerSystems will infer a rate from line parameters" maxlog =
```
daniel-thom, line 91: Would it be better to use `summary(branch)` instead? You'll get the branch type and name.

daniel-thom (approved):

PR #1124, opened 2024-06-12
Make `vom_cost` a `LinearCurve` in `ProductionVariableCost`s, add tests

GabrielKS:
As discussed with @rodrigomha and @jd-lara. Straightforward here, but let's get a companion branch in PSI before merging?


rodrigomha (approved):
This looks OK for me. It will break some minor stuff in PSI, but we can fix that eventually, by just replacing the Float by reading the `proportional_term` of the LinearCurve

GabrielKS:
Are you going to do anything with the `constant_term`? If not, should it be enforced that that is always zero?

rodrigomha:
> Are you going to do anything with the `constant_term`? If not, should it be enforced that that is always zero?

I don't think we have plans to use it. At least at this point, but I wouldn't force it to be zero. Probably we will find a weird case in the future with a constant VOM (yeah constant variable) term

jd-lara (approved):

Issue #1048, opened 2024-02-06
Implement pretty print REPL methods for operation cost


jd-lara:

jd-lara:
@GabrielKS We need to implement print methods for the cost function data.

PR #1125, opened 2024-06-12
Add Pretty Cost Printing

GabrielKS:
Fixes https://github.com/NREL-Sienna/PowerSystems.jl/issues/1048. Depends on https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/377.

Here's how the new cost struct printouts look at their most complicated (printing some operational costs):
Screenshot 2024-06-12 at 5 11 50 PM

Next is to figure out a way of abbreviating this to stick it into the components printout.


GabrielKS:
Okay, now components show more cost information than before but still a very abbreviated summary to more or less fit on one line:
Screenshot 2024-06-14 at 1 09 32 PM
Screenshot 2024-06-14 at 1 10 01 PM
Screenshot 2024-06-14 at 1 10 19 PM

kdayday (commented):
Is "(a type of InputOutputCurve)" necessary? We haven't focused on InputOutputCurve elsewhere in documentation, so I think it could be extraneous and confusing

Also, could shorten "composed of" to "with" or something, but otherwise, looks good

jd-lara (approved):

PR #1126, opened 2024-06-13
fixing tabular parser

claytonpbarrows:
I've replicated the PSY3.0 objective function created by the tabular parsing of the RTS using the `FuelCurve` functionality.  However, I haven't developed tests. In the interest of time, I'm opening this PR now.

### DIFF
--- a/src/parsers/power_system_table_data.jl
+++ b/src/parsers/power_system_table_data.jl
@@ -934,7 +934,7 @@ function create_pwl_cost(
         var_cost = PiecewisePointCurve(PiecewiseLinearData(cost_pairs))
     elseif length(cost_pairs) == 1
         # if there is only one point, use it to determine the constant $/MW cost
-        var_cost = LinearCurve(cost_pairs[1].y/cost_pairs[1].x)
+        var_cost = LinearCurve(cost_pairs[1].y / cost_pairs[1].x)
     else
         @warn "$(gen.name) has no costs defined, using 0.0" cost_pairs maxlog = 5
         var_cost = LinearCurve(0.0)
--- a/src/parsers/power_system_table_data.jl
+++ b/src/parsers/power_system_table_data.jl
@@ -1051,7 +1051,12 @@ function make_reactive_params(
     return reactive_power, reactive_power_limits
 end
 
-function make_thermal_generator(data::PowerSystemTableData, gen, cost_colnames::_CostPointColumns, bus)
+function make_thermal_generator(
+    data::PowerSystemTableData,
+    gen,
+    cost_colnames::_CostPointColumns,
+    bus,
+)
     @debug "Making ThermaStandard" _group = IS.LOG_GROUP_PARSING gen.name
     active_power_limits =
         (min = gen.active_power_limits_min, max = gen.active_power_limits_max)
--- a/src/parsers/power_system_table_data.jl
+++ b/src/parsers/power_system_table_data.jl
@@ -934,9 +934,10 @@ function create_pwl_cost(
         var_cost = PiecewisePointCurve(PiecewiseLinearData(cost_pairs))
     elseif length(cost_pairs) == 1
         # if there is only one point, use it to determine the constant $/MW cost
-        var_cost = LinearCurve(cost_pairs[1].y)
+        var_cost = LinearCurve(cost_pairs[1].y/cost_pairs[1].x)
     else
-        @warn "Unable to calculate variable cost for $(gen.name)" cost_pairs maxlog = 5
+        @warn "$(gen.name) has no costs defined, using 0.0" cost_pairs maxlog = 5
+        var_cost = LinearCurve(0.0)
     end
 
     return var_cost
--- a/src/parsers/power_system_table_data.jl
+++ b/src/parsers/power_system_table_data.jl
@@ -953,7 +953,7 @@ function create_pwinc_cost(
             first(y_points) * first(x_points),
             x_points,
             y_points[2:end],
-            )
+        )
     elseif length(cost_pairs) == 1
         # if there is only one point, use it to determine the constant $/MW cost
         var_cost = LinearCurve(cost_pairs[1].y)
--- a/src/parsers/power_system_table_data.jl
+++ b/src/parsers/power_system_table_data.jl
@@ -12,23 +12,23 @@ const INPUT_CATEGORY_NAMES = [
 ]
 struct PowerSystemTableData
     base_power::Float64
-    category_to_df::Dict{InputCategory, DataFrames.DataFrame}
-    timeseries_metadata_file::Union{String, Nothing}
+    category_to_df::Dict{InputCategory,DataFrames.DataFrame}
+    timeseries_metadata_file::Union{String,Nothing}
     directory::String
     user_descriptors::Dict
     descriptors::Dict
-    generator_mapping::Dict{NamedTuple, DataType}
+    generator_mapping::Dict{NamedTuple,DataType}
 end
 
 function PowerSystemTableData(
-    data::Dict{String, Any},
+    data::Dict{String,Any},
     directory::String,
-    user_descriptors::Union{String, Dict},
-    descriptors::Union{String, Dict},
-    generator_mapping::Union{String, Dict};
-    timeseries_metadata_file = joinpath(directory, "timeseries_pointers"),
+    user_descriptors::Union{String,Dict},
+    descriptors::Union{String,Dict},
+    generator_mapping::Union{String,Dict};
+    timeseries_metadata_file=joinpath(directory, "timeseries_pointers"),
 )
-    category_to_df = Dict{InputCategory, DataFrames.DataFrame}()
+    category_to_df = Dict{InputCategory,DataFrames.DataFrame}()
 
     if !haskey(data, "bus")
         throw(DataFormatError("key 'bus' not found in input data"))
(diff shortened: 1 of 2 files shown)


jd-lara:
@claytonpbarrows I changed the target branch to the doc strings changes PR so that we can avoid name clashed and @rodrigomha can test PSB with all the changes in the same place

src/parsers/power_system_table_data.jl
```diff
@@ -807,99 +807,160 @@ function make_generator(data::PowerSystemTableData, gen, cost_colnames, bus, gen
...
-        )
-        var_cost = [(var_cost[i][1] - fixed, var_cost[i][2]) for i in 1:length(var_cost)]
-        var_cost = PiecewiseLinearData([(x, y) for (y, x) in var_cost])
-    elseif length(var_cost) == 1
-        var_cost = LinearFunctionData(var_cost[1][1])
+    if length(cost_pairs) > 1
+        x_points = getfield.(cost_pairs, :x)
+        y_points = getfield.(cost_pairs, :y)
+        b = y_points[1] - y_points[2] * x_points[1]
+        fixed = (y_points[1] - y_points[2]) * x_points[1]
```
GabrielKS, line 952: What quantity does *this* represent?

Zooming out — my understanding was that the `fixed` field in `ThermalGenerationCost` was for when the data naturally comes with a fixed cost separate from the variable cost. For data where this is not the case and it's just that the cost/fuel curve has, like, a y-intercept, `PiecewiseIncrementalCurve` et al. are capable of representing that themselves. So here, rather than contriving a `fixed` cost from data that doesn't come that way and then subtracting(?) it out from the variable part, I think we should just let the `PiecewiseIncrementalCurve` represent the data in the form that it was provided to us in. Hopefully that should also reduce the amount of fiddly math required here.

src/parsers/power_system_table_data.jl
```diff
@@ -1001,15 +1062,8 @@ function make_thermal_generator(data::PowerSystemTableData, gen, cost_colnames,
     fuel = parse_enum_mapping(ThermalFuels, gen.fuel)
```
GabrielKS, line 1062: GitHub doesn't let me leave a comment on the correct line, but a few lines up, at
```function make_thermal_generator(data::PowerSystemTableData, gen, cost_colnames, bus)```,
do we want to add the type annotation `cost_colnames::_CostPointColumns` to make sure this only gets called with cost curve-type data?

File: src/parsers/power_system_table_data.jl [hunk 3]
```diff
@@ -807,99 +807,160 @@ function make_generator(data::PowerSystemTableData, gen, cost_colnames, bus, gen
...
+end
+
+function make_cost(
+    ::Type{T},
+    data,
+    gen,
+    cost_colnames::_CostPointColumns,
+) where {T <: ThermalGen}
+    cost_pairs = get_cost_pairs(gen, cost_colnames)
+    var_cost = create_pwl_cost(cost_pairs)
```
GabrielKS, line 839: Is it the case that when the input data is a fuel curve, it's given as incremental/marginal, whereas when the input data is a cost curve, it's given as input-output? That seems plausible, just making sure. If so, maybe put in a comment to that effect.

src/parsers/power_system_table_data.jl
```diff
@@ -807,99 +807,160 @@ function make_generator(data::PowerSystemTableData, gen, cost_colnames, bus, gen
...
+function create_pwl_cost(
+    cost_pairs,
+)
+    if length(cost_pairs) > 1
+        var_cost = PiecewisePointCurve(PiecewiseLinearData(cost_pairs))
+    elseif length(cost_pairs) == 1
         # if there is only one point, use it to determine the constant $/MW cost
-        var_cost = LinearFunctionData(var_cost[1][1] * fuel_cost + vom)
-        fixed = 0.0
+        var_cost = LinearCurve(cost_pairs[1].y)
```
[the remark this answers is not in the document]
GabrielKS: Adding to my suspicion is the fact that we're doing the same thing below in `create_pwinc_cost`, where the `y` values are supposed to have different units

src/parsers/power_system_table_data.jl
```diff
@@ -807,99 +807,160 @@ function make_generator(data::PowerSystemTableData, gen, cost_colnames, bus, gen
...
+    return op_cost
+end
+
+function make_cost(
+    ::Type{T},
+    data,
+    gen,
+    cost_colnames::_HeatRateColumns,
+) where {T <: RenewableGen}
+    @warn "Heat rate parsing not valid for RenewableGen replacing with zero cost"
```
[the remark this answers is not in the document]
claytonpbarrows: I don't want to throw an error because the RTS data has zeros in some of these columns, that would render PSY4 incompatible with the RTS data, which is largely unmaintained.

src/parsers/power_system_table_data.jl
```diff
@@ -807,99 +807,160 @@ function make_generator(data::PowerSystemTableData, gen, cost_colnames, bus, gen
...
+
+        if !in(nothing, [x, y])
+            push!(vals,
+                (x = tryparse(Float64, string(x)) * base_power,
+                    y = tryparse(Float64, string(y))))
         end
-        var_cost = PiecewiseLinearData([(x, y) for (y, x) in var_cost])
+    end
+
+    last_increasing_point = findfirst(x -> x < 0.0, [diff(getfield.(vals, :x))..., -Inf])
```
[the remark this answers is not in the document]
claytonpbarrows: some of the RTS data has valid values with some trailing zeros.

src/parsers/power_system_table_data.jl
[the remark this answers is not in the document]
claytonpbarrows: The non-fuel option is more flexible since we don't really have data in that format. So, I agree with this point and made the suggested change.

src/parsers/power_system_table_data.jl
claytonpbarrows: Is there a way to achieve the desired behavior (valid costs for the range 0.0-Pmax) with a more explicit interpretation in `PiecewiseIncrementalCurve`?

If we can figure out how to do that, I'm fine with implementing the changes, especially now that we know the rest of the parsing works to get the expected objective function values in PSY4.

src/parsers/power_system_table_data.jl
claytonpbarrows: yes. the cost-point data option is less defined and less exercised.

src/parsers/power_system_table_data.jl
GabrielKS: Yeah — concretely, these two two `ThermalGenerationCost`s represent the same total cost across the same power range of 0-1*device base, it's just that `a` has a `PiecewisePointCurve` with a nonzero y-intercept and `b` has converted that to a `fixed` term:
```julia
julia> a = ThermalGenerationCost(;
           variable = CostCurve(PiecewisePointCurve([(0.0, 20.0), (0.5, 24.0), (1.0, 30.0)]), UnitSystem.DEVICE_BASE),
           fixed = 0.0,
           start_up = 0.0,
           shut_down = 0.0)

julia> b = ThermalGenerationCost(;
           variable = CostCurve(PiecewisePointCurve([(0.0, 0.0), (0.5, 4.0), (1.0, 10.0)]), UnitSystem.DEVICE_BASE),
           fixed = 20.0,
           start_up = 0.0,
           shut_down = 0.0)
```
and my claim is that we should parse to whichever form the data naturally come in, which here I think is `a`. If downstream there is a formulation that requires a cost that starts at y=0, I think that math should be done there.

jd-lara: We need to add a check for this logic in the parsing. This is similar to what @rodrigomha did in EI.jl

GabrielKS: For the record, Clayton and I met and figured out that the raw input data here is a set of x-coordinates, the average heat rate at the first x-coordinate, and the marginal heat rates (i.e., slopes) between the rest of the x-coordinates. Therefore, we agree that a simple parse like
```julia
PiecewiseIncrementalCurve(average_at_first_point*first(x_coords), x_coords, slopes)
```
and setting `fixed = 0` in the `ThermalGenerationCost` makes the most sense, assuming this results in the same objective function as before (which it should). No other math should need to be done here.

src/parsers/power_system_table_data.jl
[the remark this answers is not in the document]
GabrielKS: Am I understanding correctly that we get to make up the spec here since there's no existing data we're trying to match? The implementation as it now stands seems at least internally consistent, though I wonder why we decided that cost curves should be input-output curves and fuel curves should be incremental. Is there a rationale for that?

src/parsers/power_system_table_data.jl
GabrielKS: (see above my question regarding why we decided to make cost curves input-output when fuel curves are incremental)

src/parsers/power_system_table_data.jl
```diff
@@ -807,99 +807,160 @@ function make_generator(data::PowerSystemTableData, gen, cost_colnames, bus, gen
...
+function get_cost_pairs(gen::NamedTuple, cost_colnames)
+    base_power = gen.base_mva * gen.active_power_limits_max
+    vals = []
+    for (c, pt) in cost_colnames.columns
+        x = getfield(gen, pt)
+        y = getfield(gen, c)
+
+        if !in(nothing, [x, y])
+            push!(vals,
+                (x = tryparse(Float64, string(x)) * base_power,
```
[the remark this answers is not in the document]
GabrielKS: In the interest of getting the major release done, I'm okay keeping it as-is for now (assuming the as-is math is correct), then changing the implementation later.

jd-lara:
@claytonpbarrows we will merge this to enable a better parsing in psy4 but as discussed with @GabrielKS, there needs to be one more step of cleaning up to be done with the Plexos2Sienna effort.

PR #1130, opened 2024-06-21
Add function to add time series in bulk

daniel-thom:
Companion PR to https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/381

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -1272,9 +1272,23 @@ series arrays because of the overhead in writing the time series to the underlyi
 - `associations`: Iterable of TimeSeriesAssociation instances. Using a Vector is not
   recommended. Pass a Generator or Iterator to avoid loading all time series data into
   system memory at once.
-- `batch_size::Int`: Number of time series to add per batch.
-```
+- `batch_size::Int`: Number of time series to add per batch. Defaults to 100.
 
+# Examples
+```julia
+resolution = Dates.Hour(1)
+associations = (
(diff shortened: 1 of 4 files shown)


src/base.jl
```diff
@@ -829,33 +829,45 @@ function IS.add_time_series_from_file_metadata_internal!(
...
+            area_ts = IS.make_time_series!(cache, file_metadata)
+            IS.add_assignment!(cache, file_metadata)
+            push!(associations, TimeSeriesAssociation(component, area_ts))
+        finally
+            file_metadata.scaling_factor_multiplier = orig_sf
         end
-        file_metadata.scaling_factor_multiplier =
-            replace(file_metadata.scaling_factor_multiplier, "max" => "peak")
-        area_ts = IS.make_time_series!(cache, file_metadata)
-        key = IS.add_time_series!(data, component, area_ts; skip_if_present = true)
```
daniel-thom, line 854: @claytonpbarrows Can you review this change? I flipped the logic around checking for duplicates. Instead of `skip_if_present`, the parsing process tracks the time series that have been added and then skips duplicates. That simplified the time series management code in IS. This line was setting skip = true for Area/LoadZone. I don't think that was correct. Can you confirm?

Also, the PSY tests don't use this feature. Can you point me to a system file (time_series_pointers.json) that does use it?

jd-lara (approved):

PR #1131, opened 2024-06-21
Add Cost Alias Getters

GabrielKS:
The main goal here, discussed with @rodrigomha, is to allow a user of the `ValueCurve` cost aliases to get everything they need without unwrapping to the `FunctionData`. I also cleaned up some docs and tests that were outdated due to a VOM cost refactor. I didn't fix this failure, which I assume is due to changes elsewhere:

Screenshot 2024-06-21 at 11 35 57 AM

---
Adding @kdayday as a reviewer of these added semantics:

- `get_slopes(::PiecewisePointwiseCurve)`  _calculates_ the slopes from the underlying pointwise `FunctionData`
- `get_slopes(::PiecewiseIncrementalCurve)` _fetches_ the slopes == the y-coordinates of the underlying step `FunctionData`
- `get_average_rates(::PiecewiseAverageCurve)` _fetches_ the average rates == the y-coordinates of the underlying step `FunctionData`

### DIFF
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -49,13 +49,15 @@ export get_proportional_term
 export get_quadratic_term
 export get_constant_term
 export get_slopes
+export get_average_rates
 export get_x_lengths
 export is_convex
 export get_points
 export get_x_coords
 export get_y_coords
 
-export ValueCurve, InputOutputCurve, IncrementalCurve, AverageRateCurve
+export ValueCurve
+export InputOutputCurve, IncrementalCurve, AverageRateCurve
 export LinearCurve, QuadraticCurve
 export PiecewisePointCurve, PiecewiseIncrementalCurve, PiecewiseAverageCurve
 export ProductionVariableCostCurve, CostCurve, FuelCurve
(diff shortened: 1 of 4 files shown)


rodrigomha:
I think we should have an interface directly to the CostCurve and FuelCurve (in addition to the interface to the value_curve), for example:
```
get_constant_term(cost_curve::CostCurve{LinearCurve}) = get_constant_term(get_value_curve(cost_curve))
```

What do you think @jd-lara @GabrielKS ?

kdayday (approved):
Comments look good. Have you re-made to docs to see if these are showing up in the API or on the ValueCurve page? If not, maybe we can merge this into the docs branch when you're done where I'll also make that change to have a separate page for each valuecurve

GabrielKS:
> I think we should have an interface directly to the CostCurve and FuelCurve (in addition to the interface to the value_curve), for example:
> 
> ```
> get_constant_term(cost_curve::CostCurve{LinearCurve}) = get_constant_term(get_value_curve(cost_curve))
> ```
> 
> What do you think @jd-lara @GabrielKS ?

Mildly oppose, having so many methods only work on `CostCurve` if it is parameterized a certain way seems like bad design. Technically that is what I'm doing here too, but only because the user will be treating each cost alias as if it is a completely separate struct. But I could be convinced.

GabrielKS:
> Comments look good. Have you re-made to docs to see if these are showing up in the API or on the ValueCurve page? If not, maybe we can merge this into the docs branch when you're done where I'll also make that change to have a separate page for each valuecurve

No, haven't rebuilt the docs. Hopefully all the changes here should be independent of your efforts over there, should be minimal merge conflicts.

jd-lara (approved):

PR #1132, opened 2024-06-21
Add some more PSY 3 -> PSY 4 system conversions

GabrielKS:
Not because we've committed to providing these, but I wanted the converted version of a system anyway. Briefly ran the storage math by @rodrigomha but would appreciate a more formal review.

### DIFF
--- a/src/data_format_conversions.jl
+++ b/src/data_format_conversions.jl
@@ -128,13 +128,24 @@ function _convert_data!(
             component["storage_technology_type"] = IS.serialize(StorageTech.OTHER_CHEM)
             soc_min = component["state_of_charge_limits"]["min"]
             soc_max = component["state_of_charge_limits"]["max"]
-            # Derive storage_capacity from old state of charge limits and normalize new
-            # state of charge limits, initial capacity accordingly
-            component["storage_capacity"] = soc_max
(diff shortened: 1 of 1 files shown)


src/data_format_conversions.jl
```diff
@@ -107,7 +125,30 @@ function _convert_data!(
...
+                component["storage_level_limits"] = Dict("min" => 0.0, "max" => 1.0)
+                (component["initial_energy"] != 0.0) && throw(
+                    ArgumentError(
+                        "Maximum state of charge is zero but initial energy is not; cannot parse",
+                    ),
+                )
+                component["initial_storage_capacity_level"] = 0.0
+            else
+                # Derive storage_capacity from old state of charge limits and normalize new
+                # state of charge limits, initial capacity accordingly
```
jd-lara, line 142: This is going to be dependent on the data that the user had and we have no way of knowing the prior assumptions. We should just throw a warning that we are doing the conversion.

jd-lara (approved):
This is good. Some of the conversions are not possible to get 100% right since the previous implementation had less information.

src/data_format_conversions.jl
GabrielKS: Added a warning.

PR #1140, opened 2024-07-01
Add `input_at_zero` to cost functions

GabrielKS:
Discussing with @jd-lara et al., it sounds like in certain cases we need to be able to explicitly represent a no-load cost as part of a cost curve instance, even if that cost is not part of the representation of the actual… cost curve. For instance, some input formats define a no-load cost and *do not* define a cost at minimum generation, and (roughly) you're supposed to infer the latter from the former by assuming the data is quadratic. Here we:

- Add `input_at_zero` to all `ValueCurve` subtypes to maintain the isomorphic design, but only expose it to the user through the simplified cost alias interface as part of `PiecewiseIncrementalCurve` because that is the only place we expect it to be used. It defaults to `nothing` if not specified and is preserved under all conversions.
- Add relevant getters, alter relevant printouts.
- Make `initial_input` a `Union{Float64, Nothing}` to represent the case where we have a no-load cost but have not yet calculated a minimum generation cost. It is the user's responsibility to figure out how to fill in `initial_input`; if they want it to be `nothing` they have to specify that.
- Have the conversions error out if `initial_input` is `nothing`.
- Make the docstrings clearer about the difference between `initial_input` and `input_at_zero`.

If these changes look good, I'll next:

- Edit the `Variable Costs` modeler guide page (the one with the table of all the cost aliases) to match.
- Add hand-drawn graphs to that page to illustrate all the cost aliases — this will further disambiguate `initial_input` vs. `input_at_zero`.
- Add an `initial_input` time series to `MarketBidCost`.

### DIFF
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -63,7 +63,8 @@ export PiecewisePointCurve, PiecewiseIncrementalCurve, PiecewiseAverageCurve
 export ProductionVariableCostCurve, CostCurve, FuelCurve
 export OperationalCost, MarketBidCost, LoadCost, StorageCost
 export HydroGenerationCost, RenewableGenerationCost, ThermalGenerationCost
-export get_function_data, get_initial_input, get_value_curve, get_power_units
+export get_function_data, get_initial_input, get_input_at_zero
+export get_value_curve, get_power_units
 export get_fuel_cost, set_fuel_cost!, get_vom_cost
 export is_market_bid_curve, make_market_bid_curve
 export get_no_load_cost, set_no_load_cost!, get_start_up, set_start_up!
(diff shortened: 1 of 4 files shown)


jd-lara:
@GabrielKS I think we need to support some de-serialization case not in PSB. Take a look at the PSI integration test fails

GabrielKS:
> @GabrielKS I think we need to support some de-serialization case not in PSB. Take a look at the PSI integration test fails

It was a manifestation of https://github.com/JuliaLang/julia/issues/35053, fixed it by adding some more constructors

jd-lara:
> > @GabrielKS I think we need to support some de-serialization case not in PSB. Take a look at the PSI integration test fails
> 
> It was a manifestation of [JuliaLang/julia#35053](https://github.com/JuliaLang/julia/issues/35053), fixed it by adding some more constructors

such a difference with good testing

jd-lara (approved):

PR #1146, opened 2024-07-07
Add `is_component_in_aggregation_topology`

GabrielKS:
For PowerAnalytics/`ComponentSelector`-related reasons, it is useful to be able to test whether a given `Component` is within a given `AggregationTopology` without having a reference to the relevant `System`. This exposes that functionality. I also made the existing `get_component_in_aggregation_topology` test more thorough.


jd-lara (approved):
@rodrigomha this would be useful later for the NTP EI repository

PR #1158, opened 2024-07-16
Add Honduras structs

rodrigomha:


src/parsers/psse_dynamic_data.jl
```diff
@@ -61,10 +61,15 @@ function _parse_input_types(_v, val)
         return NaN
         # If the parameter is a tuple (as a string), then construct the tuple directly.
     elseif isa(_v, String)
+        #TODO: Generalize n-length tuple
         m = match(r"^\((\d+)\s*,\s*(\d+)\)$", _v)
+        m2 = match(r"^\((\d+)\s*,\s*(\d+),\s*(\d+)\)$", _v)
```
rodrigomha, line 66: @jd-lara Any idea how to generalize this Regex for a tuple of n-length? The PIDGOV model and WPIDHY models have a tuple with three terms.

src/models/generated/DEGOV1.jl
```diff
@@ -0,0 +1,176 @@
...
+"""Get [`DEGOV1`](@ref) `R`."""
+get_R(value::DEGOV1) = value.R
+"""Get [`DEGOV1`](@ref) `Te`."""
+get_Te(value::DEGOV1) = value.Te
+"""Get [`DEGOV1`](@ref) `P_ref`."""
+get_P_ref(value::DEGOV1) = value.P_ref
+"""Get [`DEGOV1`](@ref) `ext`."""
+get_ext(value::DEGOV1) = value.ext
+"""Get [`DEGOV1`](@ref) `states`."""
+get_states(value::DEGOV1) = value.states
```
jd-lara, line 143:
```suggestion
get_states(value::DEGOV1) = get_degov1_states(value)
```

src/parsers/psse_dynamic_data.jl
jd-lara: Let's do this for another time. For now this should be fine

jd-lara (approved):

PR #1159, opened 2024-07-19
Add option to specify equality predicate for compare_values and provide a more sensible default

GabrielKS:
Companion to https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/387, see discussion (including of backwards compatibility) there.


GabrielKS:
Tests are failing in CI due to the dependency on the IS companion branch.

src/base.jl
```diff
@@ -2266,6 +2266,7 @@ function IS.compare_values(
     y::T;
     compare_uuids = false,
     exclude = Set{Symbol}(),
+    match_fn = IS.isequivalent,
```
jd-lara, line 2269: This function probably needs a version that takes less arguments and forwards to the version with the `match_fn`

jd-lara (changes requested):
I think the better implementation for sophisticated users is to implement 

```
IS.compare_values(
     match_fn::Function,
    x::T,
     y::T;
     compare_uuids = false,
     exclude = Set{Symbol}(),
```

and forward 

```
IS.compare_values(
    x::T,
     y::T;
     compare_uuids = false,
     exclude = Set{Symbol}(),
```

to 

```
IS.compare_values(
 IS.isequivalent,
    x::T,
     y::T;
     compare_uuids = false,
     exclude = Set{Symbol}(),
```

src/base.jl
GabrielKS: Done

src/base.jl
```diff
@@ -2341,7 +2341,11 @@ function get_bus_numbers(sys::System)
     return sort(collect(sys.bus_numbers))
 end
 
+_fetch_match_fn(match_fn::Function) = match_fn
```
jd-lara, line 2344: Can't you call the IS method instead of re define?

GabrielKS: Almost did, but I was thinking it didn't make sense to add `_fetch_match_fn` to IS's public interface and didn't want to rely on private stuff

jd-lara: Not a fan of cross repo duplication but it's a minimal thing

GabrielKS:
@jd-lara checks are passing now that the IS companion is merged

PR #1163, opened 2024-07-27
Quick fix to make valuecurves visible in model library

kdayday:
Value curves page in the model library was showing missing docs due to duplication with the API. Quick fix to at least make them all visible (if a bit out of order), pending final docs organization


jd-lara:
@GabrielKS  we are planning to move value curves to IS anyway. We should make this change part of that transition.

PR #1170, opened 2024-08-05
Update asset directory for picture links

kdayday:
This breaks the docs pictures for my local build, but matches Rodrigo's suggestions and the previous format.


rodrigomha (approved):
Temp merge

PR #1173, opened 2024-08-05
add base power default to Source

rodrigomha:


jd-lara (approved):
LGTM. We need to open an issue in PowerFlows to implement this logic correctly.

PR #1178, opened 2024-08-26
fix: Adding support for quadratic functions using the table data parser

### LINKED ISSUE #1177: Table data parser does not support Quadratic heat rates

pesap:
Currently only works for heat rate and not cost curves.

### Implementation details:
- Added new fields for generators to capture quadratic heat rate curve. For simplicity, the names are `heat_rate_a0`, `heat_rate_a1`, `heat_rate_a2`. All of the new fields are `null` by default. Open to suggestions on the nameing,
- If any `heat_rate_a*` column are defined, then we call `create_poly_cost` for `ThermalGen`,
- `create_poly_cost` will return 
  - `QuadraticCostCurve` if all the coefficients are passed
  - `LinearCurve` if `a1` and `a0` is passed,
  - `LinearCurve` if `a1`  is passed,.
  - raise an Error if `a2` is passed but `a1` or `a0` is nothing,

Closes #1177

### DIFF
--- a/src/parsers/power_system_table_data.jl
+++ b/src/parsers/power_system_table_data.jl
@@ -823,7 +823,7 @@ function make_cost(
 ) where {T <: ThermalGen}
     fuel_price = gen.fuel_price / 1000.0
 
-    # We check if there is any Quadratic or Linear Data defined. If not we fall back to create PowerLoad
+    # We check if there is any Quadratic or Linear Data defined. If not we fall back to create PiecewiseIncrementalCurve
     quadratic_fields = (gen.heat_rate_a0, gen.heat_rate_a1, gen.heat_rate_a2)
 
     if any(field -> field != nothing, quadratic_fields)
--- a/src/parsers/power_system_table_data.jl
+++ b/src/parsers/power_system_table_data.jl
@@ -964,39 +964,43 @@ function create_pwl_cost(
     return var_cost
 end
 
+"""
+    create_poly_cost(gen, cost_colnames)
+
+Return a Polynomial function cost based on the coeffiecients provided on gen.
+
+Three supported cases,
+  1. If three values are passed then we have data looking like: `a2 * x^2 + a1 * x + a0`,
+  2. If `a1` and `a0` are passed then we have data looking like: `a1 * x + a0`,
+  3. If only `a1` is passed then we have data looking like: `a1 * x`.
+"""
 function create_poly_cost(
     gen, cost_colnames,
 )
-    vals = []
-    for coeff in cost_colnames
-        a = getfield(gen, Symbol(coeff))
-        if a != nothing
-            push!(vals, tryparse(Float64, a))
-        end
-    end
+    fixed_cost = 0.0
+    parse_maybe_nothing(x) = isnothing(x) ? nothing : tryparse(Float64, x)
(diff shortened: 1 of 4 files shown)


src/parsers/power_system_table_data.jl
```diff
@@ -545,6 +545,7 @@ function gen_csv_parser!(sys::System, data::PowerSystemTableData)
         throw(IS.ConflictingInputsError("Heat rate and cost points are both defined"))
     elseif length(heat_rate_fields) > 0
         cost_colnames = _HeatRateColumns(zip(heat_rate_fields, output_point_fields))
+        @warn cost_colnames
```
rodrigomha, line 548: I guess we should give a proper warning? Or remove it?

src/parsers/power_system_table_data.jl
```diff
@@ -823,8 +824,17 @@ function make_cost(
 ) where {T <: ThermalGen}
     fuel_price = gen.fuel_price / 1000.0
 
-    cost_pairs = get_cost_pairs(gen, cost_colnames)
-    var_cost, fixed = create_pwinc_cost(cost_pairs)
+    # We check if there is any Quadratic or Linear Data defined. If not we fall back to create PowerLoad
+    @warn typeof(gen)
```
rodrigomha, line 828: Ditto

src/parsers/power_system_table_data.jl
```diff
@@ -956,6 +966,40 @@ function create_pwl_cost(
...
+        a = getfield(gen, Symbol(coeff))
+        if a != nothing
+            push!(vals, tryparse(Float64, a))
+        end
+    end
+
+    # Three supported cases,
+    #   1. If three values are passed then we have data looking like: a^2 + a + c,
+    #   2. If two then data is looking like: a + c
+    #   3. If two then data is looking like: c
```
rodrigomha, line 983: This note should say something like:
```
# Three supported cases,
#   1. If three values are passed then we have data looking like: a * x^2 + b * x + c,
#   2. If two then data is looking like: a * x + b
#   3. If two then data is looking like: a * x
```

rodrigomha (changes requested):
Remove or update the warnings and update briefly the docs

pesap:
> Remove or update the warnings and update briefly the docs

This has been addressed.

GabrielKS (changes requested):

rodrigomha (approved):

GabrielKS (approved):

rodrigomha (approved):

test/test_power_system_table_data.jl
```diff
@@ -181,3 +181,51 @@ end
     g = get_components(ThermalStandard, sys)
     @test get_variable.(get_operation_cost.(g)) == get_variable.(get_operation_cost.(g))
 end
+
+@testset "Test create_poly_cost function" begin
+
+    cost_colnames = ["heat_rate_a0", "heat_rate_a1", "heat_rate_a2"]
+
+    # Coefficients for a CC using natura gas
```
GabrielKS, line 188:
```suggestion
    # Coefficients for a CC using natural gas
```

GabrielKS (changes requested):
Looks good aside from the typo

GabrielKS (approved):

PR #1079, opened 2024-03-18
Move ComponentSelector from PowerAnalytics to InfrastructureSystems: PSY portion

GabrielKS:
This draft PR will get filled in as @tengis-nrl and I move the ComponentSelector features from PowerAnalytics to InfrastructureSystems.

PR #1197, opened 2024-09-18
Add `ComponentSelector` feature: PSY portion (take two)

GabrielKS:
Successor to https://github.com/NREL-Sienna/PowerSystems.jl/pull/1079 and companion to/depends on https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/342. Adds an immutable, lazy, system-independent representation of a grouped set of components. For more details and a demonstration, see https://github.nrel.gov/gkonars/PowerAnalytics-demos/blob/main/component_selector_pr_demo.ipynb.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -1262,9 +1262,6 @@ get_available_components(
         subsystem_name = subsystem_name,
     )
 
-get_available_components(selector::ComponentSelector, sys::System) =
-    filter(get_available, get_components(selector, sys))
-
 """
 Like [`get_component`](@ref) but also returns `nothing` if the component is not `get_available`.
 """
--- a/src/base.jl
+++ b/src/base.jl
@@ -99,7 +99,7 @@ sys = System(100.0; compression = CompressionSettings(
 sys = System(100.0; time_series_in_memory = true)
 ```
 """
-struct System <: IS.InfrastructureSystemsType
+struct System <: IS.SystemLike
     data::IS.SystemData
     frequency::Float64 # [Hz]
     bus_numbers::Set{Int}
--- a/src/base.jl
+++ b/src/base.jl
@@ -99,7 +99,7 @@ sys = System(100.0; compression = CompressionSettings(
 sys = System(100.0; time_series_in_memory = true)
 ```
 """
-struct System <: IS.SystemLike
+struct System <: IS.ComponentContainer
     data::IS.SystemData
     frequency::Float64 # [Hz]
     bus_numbers::Set{Int}
@@ -1220,55 +1220,10 @@ function get_components_by_name(
     return IS.get_components_by_name(T, sys.data, name)
 end
 
-"""
-Like `get_components` but returns only those components `c` for which `get_available(c)`.
-"""
-IS.get_available_components(
-    ::Type{T},
-    sys::System;
-    subsystem_name = nothing,
-) where {T <: Component} =
-    IS.get_components(get_available, T, sys; subsystem_name = subsystem_name)
-
-IS.get_available_components(
-    filter_func::Function,
-    ::Type{T},
-    sys::System;
-    subsystem_name = nothing,
-) where {T <: Component} =
-    IS.get_components(
-        x -> get_available(x) && filter_func(x),
-        T,
-        sys;
-        subsystem_name = subsystem_name,
-    )
-
-function _get_available_component(args...; kwargs...)
-    the_component = IS.get_component(args...; kwargs...)
-    return get_available(the_component) ? the_component : nothing
-end
-
-"""
-Like `get_component` but also returns `nothing` if the component is not `get_available`.
-"""
-IS.get_available_component(sys::System, args...; kwargs...) =
-    _get_available_component(sys, args...; kwargs...)
-
-IS.get_available_component(
-    arg1::Union{Base.UUID, String},
-    sys::System,
-    args...;
-    kwargs...,
-) =
-    _get_available_component(arg1, sys, args...; kwargs...)
-
-IS.get_available_component(
-    ::Type{T},
-    sys::System,
-    args...;
-    kwargs...,
-) where {T <: Component} =
-    _get_available_component(T, sys, args...; kwargs...)
+# PSY availability is a pure function of the component and the system is not needed; here we
+# implement the required IS.ComponentContainer interface
+IS.get_available(::System, component::Component) =
+    get_available(component)
 
 """
 Return true if the component is attached to the system.
--- a/src/base.jl
+++ b/src/base.jl
@@ -1238,14 +1238,39 @@ function get_components_by_name(
 end
 
 """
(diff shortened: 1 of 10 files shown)


GabrielKS:
This is now ready for in-depth code review! Note the dependency on https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/342. For a demonstration, see https://github.nrel.gov/gkonars/PowerAnalytics-demos/blob/main/component_selector_pr_demo.ipynb.

src/component_selector.jl
```diff
@@ -0,0 +1,124 @@
...
+    get_components(filterby, selector, sys)
+Get the components of the `System` that make up the `ComponentSelector`.
+ - `filterby`: optional filter function to apply after evaluating the `ComponentSelector`
+"""
+get_components(
+    filterby::Union{Nothing, Function},
+    selector::ComponentSelector,
+    sys::System,
+) =
+    get_components(selector, sys; filterby = filterby)
```
daniel-thom, line 22: I think this would be slightly cleaner.
```
get_components(selector::ComponentSelector, sys::System) = IS.get_components(selector, sys)
get_components(filter by::Function, selector::ComponentSelector, sys::System) = get_components(selector, sys; filterby = filterby)
```

daniel-thom: `filter_func` instead of `filterby`? Even if the second is a better term, it's better to keep consistency in the docstrings.

daniel-thom: Related to our offline discussion, the fact that you can create the component selector with a filter function and then pass a separate filter function here adds complexity and possibly confusion to the user interface.

src/component_selector.jl
```diff
@@ -0,0 +1,124 @@
...
+IS.get_components(selector::ComponentSelector, sys::System; filterby = nothing) =
+    IS.get_components(selector, sys.data; filterby = filterby)
+
+"""
+    get_component(selector, sys; filterby = nothing)
+Get the component of the `System` that makes up the `SingularComponentSelector`; `nothing`
+if there is none.
+ - `filterby`: optional filter function to apply after evaluating the `ComponentSelector`
+"""
+get_component(selector::SingularComponentSelector, sys::System; filterby = nothing) =
```
daniel-thom, line 35: Similar to above, these methods could be defined without kwargs or unions.

src/component_selector.jl
GabrielKS: Renamed this to `scope_limiter` to try to reduce confusion — see [this](https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/342#discussion_r1779178830) thread in the IS PR.

src/component_selector.jl
GabrielKS: I'm fine to add the method where the function is the first argument, but I'm hoping to keep the kwarg version as well — it makes more sense to me in the context where this would be used. Is that objectionable?

src/component_selector.jl
GabrielKS: `scope_limiter` is now no longer part of the user interface (see the IS PR).

src/component_selector.jl
GabrielKS: I was outvoted; the `filterby` kwarg is no more (see the IS PR)

src/base.jl
```diff
@@ -1100,36 +1104,15 @@ See also: [`iterate_components`](@ref), [`get_components` with a filter](@ref ge
...
-function get_components(
+function IS.get_components(
     ::Type{T},
     sys::System;
     subsystem_name = nothing,
 ) where {T <: Component}
     return IS.get_components(T, sys.data; subsystem_name = subsystem_name)
 end
 
-"""
```
GabrielKS, line 1111: Moved to `get_components_interface` (though there's an argument it should appear here too)

src/base.jl
```diff
@@ -1184,16 +1167,10 @@ function get_components_by_name(
     return IS.get_components_by_name(T, sys.data, name)
 end
 
-"""
```
GabrielKS, line 1187: Moved to `get_components_interface` (though there's an argument it should appear here too)

GabrielKS:
@daniel-thom Ready for another review. Besides the changes in the IS PR, the main change here is discussed at the top of `get_components_interface.jl`.

src/base.jl
GabrielKS: Not quite. `IS.get_components` will still be distinct from `PSY.get_components`, it's just that both will now be defined on `System` and the latter will just call the former. See the top of `get_components_interface.jl` for a detailed explanation.

jd-lara:
@GabrielKS we will scope this PR for PSY 5.0

src/get_components_interface.jl
```diff
@@ -0,0 +1,260 @@
...
+"""
+get_available_component(
+    selector::IS.SingularComponentSelector,
+    sys::System,
+) =
+    IS.get_available_component(selector, sys)
+
+# get_groups
+"""
+Get the groups that make up the `ComponentSelector`. Optionally specify a filter function
```
daniel-thom, line 227: The functionality in get_groups combined with groupby is very powerful and clever. However, the documentation is a bit lacking.

1. I understand how these capabilities are used in PowerAnalytics.jl. Let's take an example where you want to sum active_power for generators grouped by fuel type. We need this functionality merged ASAP. 100% agree and let's do it.
2. If this is to be a public interface that we expect users to use, the documentation needs improvement. At minimum, there should be a complete workflow example. Common questions will be, "what is the difference between `groupby = :each` and `groupby = :all`? What is an example of a function that should be passed? This leaves me with three conflicting opinions. (1) Leave it alone with the understanding that very few people will ever consider this. (2) Don't export this method. (3) Go all in and give complete documentation.
3. Once a power user understands what's going on here, they are likely to ask why we did all this. They might say, why not rely on existing packages, like this example
```
using DataFrames
julia> combine(groupby(DataFrame(get_components(ThermalStandard, sys)), :fuel), :active_power => sum)
5×2 DataFrame
 Row │ fuel                               active_power_sum 
     │ ThermalF…                          Float64          
─────┼─────────────────────────────────────────────────────
   1 │ ThermalFuels.NATURAL_GAS = 7              29.7472
   2 │ ThermalFuels.COAL = 1                     13.6381
   3 │ ThermalFuels.DISTILLATE_FUEL_OIL…          6.5
   4 │ ThermalFuels.OTHER = 14                    0.0
   5 │ ThermalFuels.NUCLEAR = 9                   0.849257
```
If we go all in on documentation, we should anticipate this question and address it.

daniel-thom (approved):
I think it's ready to merge unless @jd-lara or @GabrielKS want to change something about public exports. Documentation improvements can be made in future work.

GabrielKS:
@daniel-thom I definitely agree that documentation needs elaboration (and that that should be done in a subsequent PR). I'm good with the public exports as they are right now if others are.

jd-lara (approved):

PR #1202, opened 2024-10-13
feat: Adding capability to create PiecewiseLinear from csv parser

pesap:
This enable to create a PiecewiseLinear or Increment from the CSV parser. Before everything was parsed as PiecewiseIncr. I will fix the test on the following days. Do not merge.


GabrielKS:
FYI, [last time](https://github.com/NREL-Sienna/PowerSystems.jl/pull/1126) we upgraded the tabular data parser we explicitly decided not to support creating input-output fuel curves.

pesap:
That is ok. We can remove this branch once the Nevada project is done.

PR #1205, opened 2024-10-21
Add LCC and VSC Data Model, including parser

rodrigomha:


rodrigomha:
@jd-lara this is ready for review

src/models/generated/TwoTerminalHVDCDetailedLine.jl
```diff
@@ -0,0 +1,164 @@
+#=
+This file is auto-generated. Do not edit.
+=#
+
+#! format: off
+
+"""
+    mutable struct TwoTerminalHVDCDetailedLine <: ACBranch
```
jd-lara, line 8: Why not re-use the VSC existing structure?

src/PowerSystems.jl
```diff
@@ -35,6 +35,7 @@ export MonitoredLine
 export DCBranch
 export TwoTerminalHVDCLine
 export TwoTerminalVSCDCLine
```
jd-lara, line 37: Add a comment that this will be dropped in PSY 5.0 please.

rodrigomha:
This is ready for review @jd-lara 
We need to merge this to get testing passing in our PSI branch of 2T physical models.

jd-lara (changes requested):
Add code here that reads from PSSe the VSC data and puts into the struct

PR #1214, opened 2024-11-21
Vpp and flexibleload

akody:


src/models/generated/FlexiblePowerLoad.jl
```diff
@@ -0,0 +1,146 @@
...
+    mutable struct FlexiblePowerLoad <: ControllableLoad
+        name::String
+        available::Bool
+        bus::ACBus
+        active_power::Float64
+        reactive_power::Float64
+        max_active_power::Float64
+        max_reactive_power::Float64
+        base_power::Float64
+        balance_time_period::Union{Nothing, Dates.Period}
```
jd-lara, line 17: I suggest to use an Integer and specify that the value is hours.

test/test_vpp_system.jl
```diff
@@ -0,0 +1,31 @@
+@testset "VPP System tests" begin
+    test_sys = PSB.build_system(PSB.PSITestSystems, "c_sys14"; add_forecasts = false)
+
+    #include("/Users/akody/Library/CloudStorage/OneDrive-NREL/Sienna/VPP_Project/PowerSystems.jl/src/models/VPPSystem.jl")
```
GabrielKS, line 4:
```suggestion
```

PR #1216, opened 2024-11-23
PERF: dispatch on `UnitSystem` value in `_get_multiplier`

GabrielKS:
Before:
Screenshot 2024-11-22 at 4 20 21 PM

After:
Screenshot 2024-11-22 at 5 28 35 PM

On the current `@scoped_enum` implementation, `UnitSystem.NATURAL_UNITS` isn't a constant, it's a dictionary lookup — one that incurs 80 bytes of memory allocation. Thus you can see the performance impact piling up in the "before" case in the order in which the unit systems are checked — `DEVICE_BASE`, then `SYSTEM_BASE`, then `NATURAL_UNITS`. There are a few ways one might fix this:

1. Alter `@scoped_enum` to make `UnitSystem.NATURAL_UNITS` behave more like an immutable struct lookup, either by actually making it one or doing something to clue the compiler into the fact that it will never change. This would be the best, most holistic solution, but I couldn't immediately figure out how we might do it without completely overhauling `@scoped_enum`. I'll keep it in the back of my head though.
2. Instead of testing `setting.unit_system == IS.UnitSystem.NATURAL_UNITS`, do `setting.unit_system.value == 2`, etc. No dictionary lookup, but too hardcoded for me.
3. The multiple dispatch approach given here.

Most of the cost savings are from dispatching on the `UnitSystem` value. Adding another layer of dispatch on the nothing-ness of `get_internal(c).units_info` seemed to make it a bit faster on top of that, so I've done that too.

### DIFF
--- a/src/models/components.jl
+++ b/src/models/components.jl
@@ -7,23 +7,38 @@ Default behavior of a component. If there is no base_power field, assume is in t
 """
 get_base_power(c::Component) = get_system_base_power(c)
 
-function _get_multiplier(c::T) where {T <: Component}
-    setting = get_internal(c).units_info
-    if isnothing(setting)
-        return 1.0
-    elseif setting.unit_system == IS.UnitSystem.DEVICE_BASE
-        return 1.0
-    elseif setting.unit_system == IS.UnitSystem.SYSTEM_BASE
-        numerator = get_base_power(c)
-        denominator = setting.base_value
-    elseif setting.unit_system == IS.UnitSystem.NATURAL_UNITS
-        numerator = get_base_power(c)
-        denominator = 1.0
-    else
-        error("Undefined Conditional")
-    end
-    return numerator / denominator
-end
+_get_multiplier(c::T) where {T <: Component} =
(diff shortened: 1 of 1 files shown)


src/models/components.jl
```diff
@@ -7,23 +7,38 @@ Default behavior of a component. If there is no base_power field, assume is in t
...
+    ::Val{IS.UnitSystem.SYSTEM_BASE},
+) where {T <: Component} =
+    get_base_power(c) / setting.base_value
+_get_multiplier(
+    c::T,
+    ::IS.SystemUnitsSettings,
+    ::Val{IS.UnitSystem.NATURAL_UNITS},
+) where {T <: Component} =
+    get_base_power(c)
+_get_multiplier(::T, ::IS.SystemUnitsSettings, ::Val) where {T <: Component} =
```
daniel-thom, line 40: Do we need this one or can let Julia do its normal behavior?

jd-lara (approved):

PR #1230, opened 2024-12-17
Parse missing sw shunt data in psse.jl file

mcllerena:
Adding missing parameters "VSWHI" (upper_limit), "VSWLO" (lower limit), Bi (admittance increment) & Ni (number of steps), to the function _psse2pm_shunt!.

### DIFF
--- a/src/models/generated/SwitchedAdmittance.jl
+++ b/src/models/generated/SwitchedAdmittance.jl
@@ -10,8 +10,9 @@ This file is auto-generated. Do not edit.
         available::Bool
         bus::ACBus
         Y::Complex{Float64}
-        number_of_steps::Int
-        Y_increase::Complex{Float64}
+        number_of_steps::Union{Int, Vector{Int}}
+        Y_increase::Union{Complex{Float64}, Vector{Complex{Float64}}}
+        admittance_limits::MinMax
         dynamic_injector::Union{Nothing, DynamicInjection}
         services::Vector{Service}
         ext::Dict{String, Any}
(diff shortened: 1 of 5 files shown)


jd-lara (changes requested):
We need to add a new key to distinguish between fixed and switch shunts too

src/parsers/pm_io/psse.jl
```diff
@@ -344,7 +344,7 @@ for Switched Shunts, as given by the PSS(R)E Fixed and Switched Shunts
 specifications.
 """
 function _psse2pm_shunt!(pm_data::Dict, pti_data::Dict, import_all::Bool)
-    @info "Parsing PSS(R)E Shunt data into a PowerModels Dict..."
+    @info "Parsing PSS(R)E Shunt data into a PowerxModels Dict..."
```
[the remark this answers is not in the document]
mcllerena: Sorry, a misstep on the keyboard. I've already fix it.

jd-lara:
@mcllerena you need to run the formatter

jd-lara (changes requested):
@mcllerena this looks good to me. We need to add a case for testing this code and also add the code to read this data into the SwtichedShunt structure.

jd-lara (changes requested):
We need to add here the code that puts the PM data into the SW shunt struct too.

jd-lara:
@GabrielKS please confirm that we can export to PSSe too please

src/models/generated/SwitchedAdmittance.jl
```diff
@@ -10,8 +10,9 @@ This file is auto-generated. Do not edit.
         available::Bool
         bus::ACBus
         Y::Complex{Float64}
-        number_of_steps::Int
-        Y_increase::Complex{Float64}
+        number_of_steps::Union{Int, Vector{Int}}
```
GabrielKS, line 13: Why do we need both `Int` and  `Vector{Int}`?

GabrielKS (commented):
Reviewing at the newer version https://github.com/NREL-Sienna/PowerSystems.jl/pull/1252

src/models/generated/SwitchedAdmittance.jl
jd-lara: I think I originally added this to enable contingencies x time steps. We can bring it to Vector Int.

jd-lara: I think that PSSe has both options

rodrigomha:
Closing as being addressed with 3W-XFRM in #1252

Issue #1229, opened 2024-12-16
Move `with_units` here from PowerFlows.jl


GabrielKS:
In a recent pull request to PowerFlows.jl, I've prototyped this function [`with_units`](https://github.com/NREL-Sienna/PowerFlows.jl/blob/465e92fcf921c460f1bd9e0c042c2cc335897c2f/src/psse_export.jl#L47) that can be combined with the Julia `do` block to perform some action with a `System` under a particular units setting without permanently changing the units setting:

```julia
with_units(sys, PSY.UnitSystem.SYSTEM_BASE) do
  # things in this block will evaluate as if the System's unit system was SYSTEM_BASE no matter what it was before
end
# the System's unit system is back to normal
```

I think it should be moved to PSY and exported. Also

- A less heavy-handed approach to suppressing the logging should probably be taken
- Tests need to be written

Adding this as an issue here in case someone wants to take it on to learn about writing tests, etc. — otherwise I'll do it myself later this week.

jd-lara:
We should add this to the power flow project

PR #1232, opened 2024-12-23
Add `with_units_base` feature

GabrielKS:
Closes https://github.com/NREL-Sienna/PowerSystems.jl/issues/1229 by adding a new exported function `with_units_base` that serves as a "context manager" to execute a block of code as if the system's units base were set to a particular value.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -523,8 +523,6 @@ _set_units_base!(system::System, settings::String) =
     _set_units_base!(system::System, UNIT_SYSTEM_MAPPING[uppercase(settings)])
 
 """
-`set_units_base_system!(system::System, units::Union{UnitSystem, String})`
-
 Sets the units base for the getter functions on the devices. It modifies the behavior of all getter functions
 
 # Examples
@@ -551,10 +549,8 @@ function get_units_base(system::System)
 end
 
 """
-`with_units_base(function, system, units::Union{UnitSystem, String})`
-
-A "context manager" similar to `Logging.with_logger` that sets the system's units base to
-the given value, executes the function, then sets the units base back.
+A "context manager" that sets the [`System`](@ref)'s [units base](@ref per_unit) to the
+given value, executes the function, then sets the units base back.
 
 # Examples
 ```julia
--- a/src/base.jl
+++ b/src/base.jl
@@ -513,36 +513,48 @@ function set_units_setting!(
     return
 end
 
+function _set_units_base!(system::System, settings::UnitSystem)
+    to_change = (system.units_settings.unit_system != settings)
+    to_change && (system.units_settings.unit_system = settings)
+    return (to_change, settings)
+end
+
+_set_units_base!(system::System, settings::String) =
+    _set_units_base!(system::System, UNIT_SYSTEM_MAPPING[uppercase(settings)])
+
 """
+`set_units_base_system!(system::System, units::Union{UnitSystem, String})`
+
 Sets the units base for the getter functions on the devices. It modifies the behavior of all getter functions
+
+# Examples
+```julia
+set_units_base_system!(sys, "NATURAL_UNITS")
+```
+```julia
+set_units_base_system!(sys, UnitSystem.SYSTEM_BASE)
+```
 """
-function set_units_base_system!(system::System, settings::String)
-    set_units_base_system!(system::System, UNIT_SYSTEM_MAPPING[uppercase(settings)])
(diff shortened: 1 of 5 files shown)


src/base.jl
```diff
@@ -536,6 +536,42 @@ function get_units_base(system::System)
...
+active_power_mw = with_units_base(sys, UnitSystem.NATURAL_UNITS) do
+    get_active_power(gen)
+end
+# now active_power_mw is in natural units no matter what units base the system is in
+```
+"""
+function with_units_base(f::Function, sys::System, units::Union{UnitSystem, String};
+    suppress_units_info = true)
+    units_logger = if suppress_units_info
+        x -> Logging.with_logger(x, Logging.SimpleLogger(Logging.Warn))
```
GabrielKS, line 558: Something like this is needed to prevent getting spammed by `[ Info: Unit System changed to...` messages, but the current implementation has the undesired consequence of suspending any fancy logging the user has configured, directing any warning and error logs from `set_units_base_system!` (of which, granted, there should be none) only to the console. @daniel-thom do you know of a simple way to temporarily filter out all logs below a given level without replacing existing loggers? `Logging.disable_logging` is almost there, but there doesn't seem to be a public way to query the state to return to…. If not, maybe we change the interface to "suppresses _all_ logging from internal calls….

GabrielKS: Or maybe I've overcomplicated this and should just bypass `set_units_base_system!` altogether.

daniel-thom: If I understood your last point, you would have a separate code path that does not issue the log message, right? If so, I like that.

daniel-thom (approved):
LGTM

src/base.jl
GabrielKS: Done

src/base.jl
```diff
@@ -513,27 +513,65 @@ function set_units_setting!(
...
 """
 Get the system's [unit base](@ref per_unit))
 """
 function get_units_base(system::System)
-    return string(system.units_settings.unit_system)
+    return string(_get_units_base(system))
+end
+
+"""
+`with_units_base(function, system, units::Union{UnitSystem, String})`
```
kdayday, line 554: Delete -- already taken care of with DocStringExtensions

Screenshot 2024-12-23 at 1 31 17 PM

src/base.jl
```diff
@@ -513,27 +513,65 @@ function set_units_setting!(
...
+
+A "context manager" similar to `Logging.with_logger` that sets the system's units base to
```
kdayday, line 556:
```suggestion
A "context manager" that sets the [`System`](@ref)'s [units base](@ref per_unit) to
```

src/base.jl
```diff
@@ -513,27 +513,65 @@ function set_units_setting!(
...
+    to_change = (system.units_settings.unit_system != settings)
+    to_change && (system.units_settings.unit_system = settings)
+    return (to_change, settings)
+end
+
+_set_units_base!(system::System, settings::String) =
+    _set_units_base!(system::System, UNIT_SYSTEM_MAPPING[uppercase(settings)])
+
 """
+`set_units_base_system!(system::System, units::Union{UnitSystem, String})`
```
kdayday, line 526: Delete -- redundant

https://nrel-sienna.github.io/PowerSystems.jl/previews/PR1232/api/public/#PowerSystems.set_units_base_system!-Tuple%7BSystem,%20Union%7BUnitSystem,%20String%7D%7D

kdayday (commented):
Could you add an example at the end of this tutorial section showing `with_units_base`? https://nrel-sienna.github.io/PowerSystems.jl/stable/tutorials/creating_system/#Changing-System-Per-Unit-Settings
and mention in https://nrel-sienna.github.io/PowerSystems.jl/stable/explanation/per_unit/ with the other functions (halfway down). Thanks!

src/base.jl
GabrielKS: Done

src/base.jl
GabrielKS: Done

src/base.jl
GabrielKS: Done

GabrielKS:
@kdayday fixed the docstrings and added to that tutorial and that explanation page — how do those look?

jd-lara (approved):

jd-lara:
@GabrielKS can you rebase this branch to update it?

GabrielKS:
> @GabrielKS can you rebase this branch to update it?

Done

PR #1247, opened 2025-01-10
[DRAFT] [PSY5] Add 3W Transformer Struct

rodrigomha:
Initial idea for 3W transformer based on PSSE.

We are not giving support to any tap voltage control (i.e. `COD1 = COD2 = COD3 = 0`), and we are assuming that all the data must come in system per unit. These requires the flags `CW = CZ = CM = 1`.

### DIFF
--- a/src/models/generated/Transformer3W.jl
+++ b/src/models/generated/Transformer3W.jl
@@ -36,6 +36,12 @@ This file is auto-generated. Do not edit.
         primary_turns_ratio::Float64
         secondary_turns_ratio::Float64
         tertiary_turns_ratio::Float64
+        available_primary::Bool
+        available_secondary::Bool
+        available_tertiary::Bool
+        rating_primary::Float64
+        rating_secondary::Float64
+        rating_tertiary::Float64
         services::Vector{Service}
         ext::Dict{String, Any}
         internal::InfrastructureSystemsInternal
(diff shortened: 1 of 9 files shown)


src/models/generated/Transformer3W.jl
```diff
@@ -0,0 +1,326 @@
...
+        active_power_flow_tertiary::Float64
+        reactive_power_flow_tertiary::Float64
+        r_primary::Float64
+        x_primary::Float64
+        r_secondary::Float64
+        x_secondary::Float64
+        r_tertiary::Float64
+        x_tertiary::Float64
+        rating::Union{Nothing, Float64}
+        r_12::Float64
```
jd-lara, line 28: are these for the Y equivalent?

jd-lara (changes requested):
We need to add the code that parses the PM data into the transformer object in this PR

jd-lara:
@GabrielKS can you confirm this has everything we need to export the 3W transformer as well please

src/models/generated/Transformer3W.jl
rodrigomha: Yes, the r_primary and those are the ones that are the equivalent, the r12 and those, are the direct params from PSSE

GabrielKS (commented):
Reviewing these changes at the new PR, https://github.com/NREL-Sienna/PowerSystems.jl/pull/1252

PR #1252, opened 2025-01-14
Add parser and struct of 3W Transformer and Switched Shunt

rodrigomha:


rodrigomha:
@jd-lara We had to make a rebase from main, since tests were failing due to the bug in convert component.

Tests are now passing locally with the `InfrastructureSystems#main`

image

src/parsers/power_models_data.jl
```diff
@@ -808,6 +873,33 @@ function read_branch!(
...
+    data::Dict,
+    bus_number_to_bus::Dict{Int, ACBus};
+    kwargs...,
+)
+    @info "Reading 3W transformer data"
+    if !haskey(data, "3w_transformer")
+        @info "There is no 3W transformer data in this file"
+        return
+    end
+    _get_name = get(kwargs, :xfrm_3w_name_formatter, _get_pm_3w_name)
```
GabrielKS, line 887: Shouldn't the default be to use the `name` field?

test/test_dynamic_generator.jl
```diff
@@ -482,6 +482,7 @@ end
     @test dynamics[1] == Gen1AVR
     @test get_dynamic_injector(static_gen) == Gen1AVR
     @test get_base_power(static_gen) == get_base_power(Gen1AVR)
+    @test PSY.compare_values(static_gen, deepcopy(static_gen))
```
GabrielKS, line 485: I don't object, but out of curiosity, what is this for?

test/test_parse_psse.jl
```diff
@@ -23,10 +23,11 @@ end
 
 @testset "PSSE Component Parsing" begin
     @info "Testing Load Parsing"
-    sys = build_system(PSYTestSystems, "psse_240_parsing_sys") # current/imedance_power read in natural units during parsing
+    sys = build_system(PSYTestSystems, "psse_240_parsing_sys") # current/impedance_power read in natural units during parsing
     @test get_current_active_power(get_component(StandardLoad, sys, "load10021")) == 223.71
     @test get_impedance_reactive_power(get_component(StandardLoad, sys, "load10021")) ==
           583.546
+    PSB.clear_serialized_systems("psse_Benchmark_4ger_33_2015_sys")
```
GabrielKS, line 30: Given https://github.com/NREL-Sienna/PowerSystemCaseBuilder.jl/pull/115, is this necessary anymore?

test/test_parse_psse.jl
```diff
@@ -37,6 +38,24 @@ end
...
     @test get_available(get_component(ThermalStandard, sys, "generator-2438-EG")) == 1
 
+    sys3 = build_system(PSSEParsingTestSystems, "psse_ACTIVSg2000_sys")
+    sys4 = build_system(PSSEParsingTestSystems, "pti_frankenstein_70_sys")
+
+    base_dir = string(dirname(@__FILE__))
+    file_dir = joinpath(base_dir, "test_data", "5circuit_3w.raw")
+    sys5 = System(file_dir)
+
+    @info "Testing  Three-Winding Transformer Parsing"
```
GabrielKS, line 48:
```suggestion
    @info "Testing Three-Winding Transformer Parsing"
```

test/test_parse_psse.jl
```diff
@@ -37,6 +38,24 @@ end
...
+    @test isnothing(get_component(Transformer3W, sys3, "1"))
+    @test get_available(
+        get_component(Transformer3W, sys5, "FAV SPOT 01-FAV SPOT 02-FAV SPOT 03-i_A"),
+    ) == false
+    @test length(get_components(Transformer3W, sys5)) == 5
```
GabrielKS, line 53: Seems like there is more to test than this? Do we have coverage on all the three winding stuff that was added in this PR?

test/test_parse_psse.jl
```diff
@@ -37,6 +38,24 @@ end
...
+
+    @info "Testing Switched Shunt Parsing"
+    @test get_available(get_component(SwitchedAdmittance, sys3, "1030_9")) == false
+    @test get_available(get_component(SwitchedAdmittance, sys4, "1003_1")) == true
```
GabrielKS, line 57: Same here — in particular, I'd want to test that in a case with multiple blocks, the blocks are parsed correctly

test/test_system.jl
```diff
@@ -289,6 +289,39 @@ end
     end
 end
 
+@testset "Test begin_time_series_update" begin
```
GabrielKS, line 292: I don't think this belongs in this PR — can we merge it separately?

test/test_system.jl
```diff
@@ -407,6 +440,35 @@ end
     end
 end
 
+@testset "Test fast deepcopy of system" begin
```
GabrielKS, line 443: same here, this was a different PR that should be merged into `psy5` separately

test/test_dynamic_generator.jl
GabrielKS: oh, I added this in a different PR. The Git history is a bit messed up, this should be merged to `psy5` separately so it doesn't appear in this diff.

src/descriptors/power_system_structs.json
```diff
@@ -1010,6 +1010,337 @@
             ],
             "supertype": "ACBranch"
         },
+        {
```
GabrielKS, line 1013: I will review the generated models rather than this JSON.

src/models/generated/Transformer3W.jl
```diff
@@ -0,0 +1,380 @@
...
+        rating_secondary::Float64
+        rating_tertiary::Float64
+        services::Vector{Service}
+        ext::Dict{String, Any}
+        internal::InfrastructureSystemsInternal
+    end
+
+A 3-winding transformer.
+
+The model uses an equivalent star model with a star (hidden) bus. The user must transform the data to use `CW = CZ = CM = 1` and `COD1 = COD2 = COD3 = 0` (no voltage control) if taken from a PSS/E 3W transformer model. Three equivalent impedances (connecting each side to the star bus) are required to define the model. Shunt conductance (iron losses) and magnetizing susceptance can be considered from the star bus to ground. The model is described in Chapter 3.6 in J.D. Glover, M.S. Sarma and T. Overbye: Power Systems Analysis and Design.
```
GabrielKS, line 52: What is the expected behavior if the parser encounters a PSS/E transformer record where `CW = CZ = CM = 1` or `COD1 = COD2 = COD3 = 0` is violated? Will it automatically perform this transformation?

GabrielKS (changes requested):
Sorry for the delay. I'm not seeing any glaring issues from the exporter's point of view, but it's probable that there's something I won't notice until I actually start implementing. I've found a few smaller issues and things I have questions about. Also, could we update `psy5` so this diff only shows related code? Some of Dan's and my recent PRs are in here.

rodrigomha:
I don't think we can clean up this PR. All of those files are when we rebase main into this branch. 

I rebase today the `psy5` branch with main, but its still showing the differences, since the rebase happened afterwards I think

src/models/generated/Transformer3W.jl
rodrigomha: Currently it gets ignored. We could fail, give a warning that it will be ignored, or add all the flags to the struct, and we can save them if we want, but will likely not be used anywhere. @jd-lara What do you think we should do here? We should probably add a lot of default fields with all the flags? So at least we can store them always and use them in our exporter.

src/models/generated/Transformer3W.jl
```diff
@@ -0,0 +1,380 @@
...
+
+"""
+    mutable struct Transformer3W <: ACBranch
+        name::String
+        available::Bool
+        primary_secondary_arc::Arc
+        secondary_tertiary_arc::Arc
+        primary_tertiary_arc::Arc
+        star_bus::ACBus
+        active_power_flow_primary::Float64
```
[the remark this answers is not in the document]
rodrigomha: We could made `WindingData` a `DeviceParameter` type. I like this approach. What do you think @jd-lara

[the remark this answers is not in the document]
jd-lara: In a separate call this will be added to the PSY 6.0 scope

src/models/generated/Transformer3W.jl
rodrigomha: Currently being included in the EXT

src/parsers/pm_io/psse.jl
```diff
@@ -368,26 +370,47 @@ function _psse2pm_shunt!(pm_data::Dict, pti_data::Dict, import_all::Bool)
...
+            sub_data["step_number"] = sub_data["step_number"][sub_data["step_number"] .!= 0]
+
+            y_increment = Dict(
+                k => v for
+                (k, v) in switched_shunt if startswith(k, "B") && isdigit(last(k))
+            )
+            y_increment_sorted =
+                sort(collect(keys(y_increment)); by = x -> parse(Int, x[2:end]))
+            sub_data["y_increment"] = [y_increment[k] for k in y_increment_sorted]im
+            sub_data["y_increment"] = sub_data["y_increment"][sub_data["y_increment"] .!= 0]
```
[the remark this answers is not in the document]
rodrigomha: Usually it comes with 8 blocks, and all of the ones not being used are set-up with zero. So, if the data comes in the form like `y_increment = [1.0, 2.0, 3.0, 0.0, 4.0, 0.0, 0.0, 0.0]` I believe its fine if the fourth entry is also discarded. What matters is that the data should end up being `y_increment = [1.0, 2.0, 3.0, 4.0]`.

src/parsers/power_models_data.jl
rodrigomha: No, for branches we usually want the name to use like the buses that connect + name.

test/test_parse_psse.jl
rodrigomha: I can try to remove it. We added it because it was failing for some reason.

rodrigomha:
Tests are passing locally, but they are failing in some obscure `ext` thing in 2w transformers.

I think we should merge to 3WTransformer, update everything there, and merge to psy5 asap.

rodrigomha:
Just merged LCC and VSC to this branch. We should be ready to merge this to psy5.

jd-lara (approved):

PR #1268, opened 2025-01-29
Add RT, XT, ZR, ZX parameters to ext dict for generator.

mcllerena:
Put RT, XT fields into the ext dictionary. Since ZR & ZX where already added.


GabrielKS (changes requested):
Can we get a version of this PR that only includes the relevant commits? If this gets merged it merges a lot of stuff besides just adding those parameters.

Issue #1160, opened 2024-07-25
`System` constructor keyword argument `load_zone_formatter` is ignored


GabrielKS:
There is a kwarg `load_zone_formatter` listed as one of the supported `SYSTEM_KWARGS`:
https://github.com/NREL-Sienna/PowerSystems.jl/blob/71b9bf406ef7001eb5d6b56cc780b73093c9eb54/src/base.jl#L13
but, unlike all the other `SYSTEM_KWARGS`, nothing is ever done with it. This is the only reference to it in the codebase. I'm currently in a situation where I need to remap the names of load zones, so it would be nice to support this behavior as we do for buses, generators, branches, etc.

GabrielKS:
`good first issue` because if all we need is a name formatter I think it could be implemented pretty similarly to the existing ones

rodrigomha:
What do you need to do here? The main problem is that powerworld does not create a zone dictionary. The load zones are created manually based on the information in the `bus` dictionary in `bus["zone"]` key. Would be tricky to pass a function is that we don't have access at the moment of creation of any additional data `x` that we can do something.

GabrielKS:
@rodrigomha my use case is [here](https://github.com/NREL-Sienna/PowerFlows.jl/blob/67c176e0f9d9d59dbdaa7c6fb44e34c9829c958b/src/psse_export.jl#L1011L1018) — I am creating a `System` from a PSS/E .raw file and adjusting a bunch of names based on a JSON that gets written alongside the .raw by the  new Sienna PSS/E exporter. All I need is the ability to pass in a mapping from what Sienna thinks the name should be to what I actually want the name to be and have that be evaluated.

PR #1273, opened 2025-01-31
Read export metadata file for round trip with PowerFlows.jl's PSS/E exporter

GabrielKS:
Here, I implement logic such that when `System("my_filename.raw")` is called, PowerSystems will check for a `my_filename_export_metadata.json` file written by the PowerFlows.jl PSS/E exporter in the same directory and, if it is found, use its contents to undo the system alterations PowerFlows had to make to write to the PSS/E format. The goal is for systems loaded this way to look as similar to the original system that was serialized as possible. The user can opt out of this feature by passing `try_reimport = false`. There is also interface to manually pass a metadata dictionary to perform this same remapping.

In the process of implementation, I also added a `area_name_formatter` System constructor kwarg along the same lines as https://github.com/NREL-Sienna/PowerSystems.jl/issues/1160.

No unit tests are currently included. Up until now, this functionality was implemented in PowerFlows.jl, and there are fairly detailed but not perfectly comprehensive tests of this functionality due to the testing of the round trip fidelity over there. Looking for reviewer input on whether that is sufficient to merge this now or whether dedicated unit tests in PowerSystems are required first.

Exporter is implemented in https://github.com/NREL-Sienna/PowerFlows.jl/pull/53 and that PR now depends on this one.

@daniel-thom don't feel obligated to fully review, just tagging you in case you have opinions on how I'm messing with the System constructor.

### DIFF
--- a/src/parsers/psse_metadata_reimport.jl
+++ b/src/parsers/psse_metadata_reimport.jl
@@ -115,8 +115,10 @@ function parse_export_metadata_dict(md::Dict)
         Int64,
     )
     loadzone_name_map = reverse_dict(md["zone_mapping"])
+    get!(loadzone_name_map, 1, "1")
     loadzone_name_formatter = name -> loadzone_name_map[name]
     area_name_map = reverse_dict(md["area_mapping"])
+    get!(area_name_map, 1, "1")
     area_name_formatter = name -> area_name_map[name]
 
     sys_kwargs = Dict(
--- a/src/parsers/psse_metadata_reimport.jl
+++ b/src/parsers/psse_metadata_reimport.jl
@@ -2,7 +2,7 @@
 
 # Currently the following mappings are used here:
 #   area_mapping: maps Sienna name to PSS/E number for all areas
-#   zone_mapping: maps Sienna name to PSS/E number for all zones
+#   zone_mapping: maps Sienna name to PSS/E number for all load zones
 #   bus_number_mapping: maps Sienna number to PSS/E number for all buses
 #   load_name_mapping: maps (Sienna bus, Sienna name) to PSS/E name for all loads
 #   shunt_name_mapping: maps (Sienna bus, Sienna name) to PSS/E name for all shunts
@@ -117,6 +117,8 @@ function System(raw_path::AbstractString, md::Dict)
     )
     loadzone_name_map = reverse_dict(md["zone_mapping"])
     loadzone_name_formatter = name -> loadzone_name_map[name]
+    area_name_map = reverse_dict(md["area_mapping"])
+    area_name_formatter = name -> area_name_map[name]
 
     sys =
         System(raw_path;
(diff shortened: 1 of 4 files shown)


src/parsers/psse_metadata_reimport.jl
```diff
@@ -0,0 +1,160 @@
...
+        new_name = reversed_name_mapping[(p_bus_n, p_name)]
+        return new_name
+    end
+    return component_id_formatter
+end
+
+function remap_bus_numbers!(sys::System, bus_number_mapping)
+    for bus in collect(get_components(Bus, sys))
+        set_number!(bus, parse(Int, bus_number_mapping[get_number(bus)]))
+    end
```
daniel-thom, line 75: `sys.bus_numbers` is now out-of-date.

daniel-thom (changes requested):

src/parsers/psse_metadata_reimport.jl
```diff
@@ -0,0 +1,160 @@
...
+        (p_bus_n, p_name) = device_dict["source_id"][2:3]
+        (p_bus_n isa Integer) || (p_bus_n = parse(Int64, p_bus_n))
+        new_name = reversed_name_mapping[(p_bus_n, p_name)]
+        return new_name
+    end
+    return component_id_formatter
+end
+
+function remap_bus_numbers!(sys::System, bus_number_mapping)
+    for bus in collect(get_components(Bus, sys))
```
jd-lara, line 73: why are we calling collect here?

jd-lara (commented):
The PR looks fine to me but we need to add an issue to increase test coverage.

src/parsers/psse_metadata_reimport.jl
GabrielKS: The `bus_number_mapping` is a reversed dictionary so that guarantees uniqueness. I can modify `sys.bus_numbers`, but isn't it a bug for public interface to be able to put the system into an inconsistent state?

src/parsers/psse_metadata_reimport.jl
GabrielKS: because https://github.com/NREL-Sienna/InfrastructureSystems.jl/issues/432

src/parsers/psse_metadata_reimport.jl
daniel-thom: We could set `exclude_setter = true` in power_system_structs.json and then add a function for `set_bus!(::System, ::Bus)`, similar to `set_name!`.

jd-lara: I think this solution is ok for now and we can revise the bus renumbering in PSY 6. I think that when we re-design the capability to read/write to the DB this will come back

GabrielKS: I'll fix it as a one-off here for now, principled solution tracked at https://github.com/NREL-Sienna/PowerSystems.jl/issues/1274

GabrielKS: Done

GabrielKS:
> The PR looks fine to me but we need to add an issue to increase test coverage.

-> https://github.com/NREL-Sienna/PowerSystems.jl/issues/1275

daniel-thom (approved):

jd-lara (approved):

PR #1276, opened 2025-02-01
Add set_bus_number!

### LINKED ISSUE #1274: Bug: `set_number!` on a bus puts the system into an inconsistent state
The `System` has a field `bus_numbers` that is not updated by `set_number!`. See discussion at https://github.com/NREL-Sienna/PowerSystems.jl/pull/1273#discussion_r1937478004. I claim that this is a bug.

daniel-thom:
Fixes #1274

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -2619,6 +2619,28 @@ function convert_component!(
     remove_component!(sys, old_load)
 end
 
+"""
+Set the number of a bus.
+"""
+function set_bus_number!(sys::System, bus::ACBus, number::Int)
+    if number in sys.bus_numbers
+        throw(ArgumentError("bus number $number is already stored in the system"))
+    end
+    bus.number = number
+    push!(sys.bus_numbers, number)
+    return
+end
+
+function set_number!(bus::ACBus, number::Int)
+    Base.depwarn(
+        "This method will be removed in v5.0 because its use breaks system consistency" *
+        "checks. Please call `set_bus_number!(::System, bus, number)` instead.",
+        :set_number!,
+    )
+    bus.number = number
+    return
+end
+
 # Use this function to avoid deepcopy of shared_system_references.
 function _copy_internal_for_conversion(component::Component)
     internal = get_internal(component)
--- a/test/test_system.jl
+++ b/test/test_system.jl
@@ -614,3 +614,16 @@ end
     device2 = first(get_components(ThermalStandard, sys2))
     @test_throws ArgumentError add_service!(device2, service1, sys2)
 end
+
+@testset "Test set_bus_number!" begin
+    sys = PSB.build_system(PSITestSystems, "test_RTS_GMLC_sys")
+    buses = collect(get_components(ACBus, sys))
+    bus1 = buses[1]
+    bus2 = buses[2]
+    orig = get_number(bus1)
+    new_number = 9999999
+    @test orig != new_number
+    set_bus_number!(sys, bus1, new_number)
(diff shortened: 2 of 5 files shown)


src/base.jl
```diff
@@ -2619,6 +2619,28 @@ function convert_component!(
...
+    if number in sys.bus_numbers
+        throw(ArgumentError("bus number $number is already stored in the system"))
+    end
+    bus.number = number
+    push!(sys.bus_numbers, number)
+    return
+end
+
+function set_number!(bus::ACBus, number::Int)
+    Base.depwarn(
```
daniel-thom, line 2635: Note that this message is not printed by default. It should show up in tests. I'm debating whether to make it an `@warn` instead.

jd-lara: This is ok with me, I don't think this operation is very uncommon anyway.

daniel-thom: Once we merge to main, I'll merge into `psy5` and delete the deprecated method.

jd-lara (approved):

src/base.jl
```diff
@@ -2619,6 +2619,28 @@ function convert_component!(
...
 
+"""
+Set the number of a bus.
+"""
+function set_bus_number!(sys::System, bus::ACBus, number::Int)
+    if number in sys.bus_numbers
+        throw(ArgumentError("bus number $number is already stored in the system"))
+    end
+    bus.number = number
+    push!(sys.bus_numbers, number)
```
GabrielKS, line 2630: Don't you also have to pop the old number? At least in my use case, `sys.bus_numbers` is populated with all the old bus numbers

src/base.jl
```diff
@@ -2619,6 +2619,28 @@ function convert_component!(
...
 end
 
+"""
+Set the number of a bus.
+"""
+function set_bus_number!(sys::System, bus::ACBus, number::Int)
+    if number in sys.bus_numbers
+        throw(ArgumentError("bus number $number is already stored in the system"))
+    end
+    bus.number = number
```
GabrielKS, line 2629: In `set_name!`, we `throw_if_not_attached` — for consistency, seems like we should do that here

src/base.jl
```diff
@@ -2619,6 +2619,28 @@ function convert_component!(
     remove_component!(sys, old_load)
 end
 
+"""
+Set the number of a bus.
+"""
+function set_bus_number!(sys::System, bus::ACBus, number::Int)
+    if number in sys.bus_numbers
```
GabrielKS, line 2626: Given https://github.com/NREL-Sienna/InfrastructureSystems.jl/issues/397, seems like `set_bus_number!` with the bus's current number should be a no-op

test/test_system.jl
```diff
@@ -614,3 +614,16 @@ end
...
+    sys = PSB.build_system(PSITestSystems, "test_RTS_GMLC_sys")
+    buses = collect(get_components(ACBus, sys))
+    bus1 = buses[1]
+    bus2 = buses[2]
+    orig = get_number(bus1)
+    new_number = 9999999
+    @test orig != new_number
+    set_bus_number!(sys, bus1, new_number)
+    @test get_number(bus1) == new_number
+    @test_throws ArgumentError set_bus_number!(sys, bus1, get_number(bus2))
```
GabrielKS, line 628: Test state of `sys.bus_numbers`

GabrielKS (commented):
This was merged very quickly and I didn't have a chance to review — I think there are some changes worth making.

PR #1277, opened 2025-02-01
Remove workaround for set_number!(::ACBus, ::Int)

daniel-thom:


src/PowerSystems.jl
```diff
@@ -412,7 +412,6 @@ export create_poly_cost
 #export make_time_series
 export get_bus_numbers
 export set_bus_number!
-export set_number!  # Remove this in v5.0.
```
daniel-thom, line 415: This method still exists for DCBus.

jd-lara (commented):
@daniel-thom did you intend to open this PR on main or. PSY5?

daniel-thom:
> @daniel-thom did you intend to open this PR on main or. PSY5?

main. Gabriel's workaround is no longer necessary. After this is merged, I will delete the deprecated `set_number!` function in the psy5 branch.

src/PowerSystems.jl
GabrielKS: Deleting this export shouldn't be on this PR if this PR is to be merged before the next major version, right?

GabrielKS (changes requested):

daniel-thom:
Thanks @GabrielKS. My changes in the other PR were a bit sloppy. This should address them.

src/PowerSystems.jl
daniel-thom: #1276 added this line. I later realized that this function for DCBus was already exported and reverted the change. I suppose you could argue that should have exported the function here in case DCBus is ever changed or deleted. I'm not sure it matters. I did add a test to ensure that it is exported for ACBus.

src/base.jl
```diff
@@ -2623,11 +2623,19 @@ end
 Set the number of a bus.
 """
 function set_bus_number!(sys::System, bus::ACBus, number::Int)
+    throw_if_not_attached(bus, sys)
+
+    orig = get_number(bus)
+    if number == orig
```
GabrielKS, line 2629: nit: untested

test/test_system.jl
```diff
@@ -626,4 +626,10 @@ end
     set_bus_number!(sys, bus1, new_number)
     @test get_number(bus1) == new_number
     @test_throws ArgumentError set_bus_number!(sys, bus1, get_number(bus2))
+    @test !(orig in get_bus_numbers(sys))
```
GabrielKS, line 629: Also test new number in `get_bus_numbers`

src/PowerSystems.jl
GabrielKS: I guess if I were being really nitpicky I'd say it still belongs in this file since (until PSY5) it's still part of the interface even independently of DCBus. But it's not a big deal and I'm okay with it either way.

test/test_system.jl
```diff
@@ -625,5 +625,23 @@ end
...
+
+    # Ensure that the no-op case works.
+    set_bus_number!(sys, bus1, new_number)
+    @test get_number(bus1) == new_number
+    @test new_number in get_bus_numbers(sys)
+
+    # Ensure that duplicate numbers are blocked.
     @test_throws ArgumentError set_bus_number!(sys, bus1, get_number(bus2))
+
+    # Esnure that you can't change an unattached bus.
```
GabrielKS, line 640: typo
```suggestion
    # Ensure that you can't change an unattached bus.
```

GabrielKS (approved):
LGTM aside from typo, thanks

Issue #1279, opened 2025-02-10
FACTS Parsing approach


jd-lara:
1. Throw an error if J not 0.0. I.e., do not support series FACTS. 
2. Only support operating MODE 0 (unavailable), 1 and 2 (check if PSSe allows for 2 with J = 0.0). Throw error.
3. Use defaults when exporting to PSSe files. 
4. Fields that matter: VSET (the Bus has to a PV bus, and the set point stored in the bus), SHMX, VTMIN, RMPCT (needed to redistribute reactive power)

PR #1303, opened 2025-03-18
Add parser for  FACTS Control Devices

mcllerena:
1. Add parsing capability for FACTS Control Devices under the following approach:
- Throw an error if J not 0.0 (Support only for STATCOM operation)
- Only support operating MODE 0 (unavailable), 1 and 2. PSSE throws an error when setting terminal bus (J) to 0.0 and operation mode is 2.
- Fields added: VSET (the Bus has to a PV bus, and the set point stored in the bus), SHMX & RMPCT (needed to redistribute reactive power). All fields related to I (sending bus).
2. This closes issue #1279 
3. New Struct created and tested FACTSControlDevice

Note: The test for the VSC line had to be commented out temporarily since it was using the "frankenstein_70" network, and it is broken for FACTSControlDevices (it has 2 similar IDs) and throws an error at parsing.


src/models/generated/FACTSControlDevice.jl
```diff
@@ -0,0 +1,115 @@
...
+=#
+
+#! format: off
+
+"""
+    mutable struct FACTSControlDevice <: ElectricLoad
+        name::String
+        available::Bool
+        bus::ACBus
+        mode::Int64
```
jd-lara, line 12: Call this control mode and lets make this an enum so we don't need to guess from the integer

src/models/generated/FACTSControlDevice.jl
```diff
@@ -0,0 +1,115 @@
...
+
+Facts control devices.
+
+Most often used in AC power flow studies as a control of voltage and, active and reactive power.
+
+# Arguments
+- `name::String`: Name of the component. Components of the same type (e.g., `PowerLoad`) must have unique names, but components of different types (e.g., `PowerLoad` and `ACBus`) can have the same name
+- `available::Bool`: Indicator of whether the component is connected and online (`true`) or disconnected, offline, or down (`false`). Unavailable components are excluded during simulations
+- `bus::ACBus`: Sending end bus number
+- `mode::Int64`: Control mode. Where 0: out of service, 1: series & shunt links operating (normal operation), 2: series link bypassed & shunt link operates as STATCOM
```
jd-lara, line 29: Make this an enum

src/models/generated/FACTSControlDevice.jl
```diff
@@ -0,0 +1,116 @@
+#=
+This file is auto-generated. Do not edit.
+=#
+
+#! format: off
+
+"""
+    mutable struct FACTSControlDevice <: ElectricLoad
```
jd-lara, line 8: This should be a "StaticInjection"

jd-lara (changes requested):
we should also add a field for a dynamic injector

src/models/supplemental_constructors.jl
```diff
@@ -53,6 +53,32 @@ function ACBus(
     )
 end
 
+"""Allows construction with bus type specified as a string for legacy code."""
+function FACTSControlDevice(
+    name,
+    available,
+    bus,
+    factsmode::String,
```
jd-lara, line 61:
```suggestion
    control_mode::String,
```
For consistency

src/parsers/pm_io/psse.jl
```diff
@@ -1223,6 +1223,62 @@ function _psse2pm_dcline!(pm_data::Dict, pti_data::Dict, import_all::Bool)
...
+
+            sub_data["bus"] = facts["I"] # Sending bus number
+
+            if facts["J"] != 0
+                @warn "Series FACTs not supported."
+            end
+
+            if facts["MODE"] > 3
+                throw(DataFormatError("Operation mode not supported."))
+            end
```
jd-lara, line 1257:
```suggestion
            if facts["J"] != 0
                @warn "Series FACTs not supported."
            end

            if facts["MODE"] > 3
                throw(DataFormatError("Operation mode not supported."))
            end
```
We should throw these warnings and error in the `make_facts` function call so that the parser's scope is limited to parsing as much as possible. I'd rather fail in the `make_facts` function such that if we ever need to support it we tackle it there.

src/parsers/pm_io/psse.jl
```diff
@@ -1223,6 +1223,62 @@ function _psse2pm_dcline!(pm_data::Dict, pti_data::Dict, import_all::Bool)
...
+
+            sub_data["voltage_setpoint"] = facts["VSET"] # Voltage setpoint at f_bus
+            sub_data["max_shunt_current"] = facts["SHMX"] # Max shunt current at f_bus
+
+            # % of MVAr required to hold voltage at f_bus
+            if facts["RMPCT"] < 0
+                throw(DataFormatError("% MVAr required must me positive."))
+            else
+                sub_data["reactive_power_required"] = facts["RMPCT"]
+            end
```
jd-lara, line 1267: You can log a warning here and then fail in the `make_facts` function

jd-lara (changes requested):
overall looks good, some requested changes to the location of the error throwing

mcllerena:
All test are passing locally with PowerSystems psy5 branch

src/models/generated/FACTSControlDevice.jl
jd-lara, line 1: Should this device have some sort of reactive power limits?

src/parsers/pm_io/psse.jl
```diff
@@ -1223,6 +1223,55 @@ function _psse2pm_dcline!(pm_data::Dict, pti_data::Dict, import_all::Bool)
...
+                ["facts", sub_data["bus"], sub_data["name"]]
+            sub_data["index"] = length(pm_data["facts"]) + 1
+
+            if import_all
+                _import_remaining_keys!(sub_data, facts)
+            end
+
+            push!(pm_data["facts"], sub_data)
+        end
+    end
```
jd-lara, line 1272:
```suggestion
    end
    return
```

src/models/generated/FACTSControlDevice.jl
jd-lara: never mind, this is the Max current

jd-lara (approved):

PR #1313, opened 2025-03-25
Additional line ratings for some AC Branches

### LINKED ISSUE #1253: Add multiple rates for AC branches
For PSY5, we need to add nullable fields to allow the AC branch structs to represent multiple ratings, corresponding to PSSE's `RATEA`, `RATEB`, `RATEC`.

m-bossart:
Adds additional ratings for `Line`, `MonitoredLine`, `PhaseShiftingTransformer`, `TapTransformer`, `Transformer2W`

 Questions/Comments:

- Currently uses generic names `rating_b` and `rating_c` for the additional ratings which references the PSSE terminology. The existing `rating` is unchanged. Should we instead make a more descriptive name (e.g. `emergency_rating`, `overload_rating`?). The challenge is that the exact meaning of the additional ratings seems to vary, and often they are not used at all. 
- Did not currently include three winding transformer because the parsed data seems to include only a single rating. I'm not sure if this should be modified.

Fixes #1253


src/models/generated/Line.jl
```diff
@@ -61,6 +65,10 @@ mutable struct Line <: ACBranch
     rating::Float64
     "Minimum and maximum angle limits (radians)"
     angle_limits::MinMax
+    "Second current rating; entered in MVA."
```
rodrigomha, line 68: @jd-lara Are we okay with being `Union{Nothing, Float64}`? If we default to 0.0, and only uses Float64, our code will require to check the ratings with `iszero` to bypass rather than `isnothing`. 

Nothing in here makes sense since it represents that we do not have data, while 0.0 will be something like PSSE, that 0.0 means no data.

jd-lara: I am fixing this in a separate PR and branch

jd-lara (approved):

PR #1314, opened 2025-03-25
line limits improvements

jd-lara:
This PR addresses important changes to the rate parsing from PSSe. Still requires updates to the function to "fix" the rate limits but it is open for comments now


src/utils/IO/branchdata_checks.jl
jd-lara, line 1: @rodrigomha can you check this. I deleted the logic based on SIL and used the max transfer limit calculations.

jd-lara:
This PR still needs some testing but this would simplify the calculations for the proper ratings in lines.

rodrigomha (approved):

src/utils/IO/branchdata_checks.jl
```diff
@@ -59,105 +59,114 @@ function sanitize_angle_limits!(line)
...
+    161.0 => (min = 30.0, max = 276.0),
+    230.0 => (min = 50.0, max = 470.0),
+    345.0 => (min = 160.0, max = 702.0),
+    500.0 => (min = 150.0, max = 1383.0),
+    765.0 => (min = 2200.0, max = 6900.0), # This value is 3x the SIL value from https://neos-guide.org/wp-content/uploads/2022/04/line_flow_approximation.pdf
+)
 
-    return new_rate
-end
+function check_rating_values(line::Union{Line, MonitoredLine}, basemva::Float64)
```
pesap, line 86: Are we sure that at this point the `rating_value` can not be negative?

pesap: I also think that if the function is called `check` we need to return `true` or `false`. It looks like it is always returning `true`

src/utils/IO/branchdata_checks.jl
```diff
@@ -59,105 +59,114 @@ function sanitize_angle_limits!(line)
...
+    theoretical_line_rate_pu = line_rating_calculation(branch)
+    for field in [:rating, :rating_b, :rating_c]
+        rating_value = getfield(branch, field)
+        if isnothing(rating_value)
+            @assert field ∈ [:rating_b, :rating_c]
+            continue
+        end
+        if rating_value < 0.0
+            @error "PowerSystems does not support negative line rates. $(field): $(rating)"
+            return false
```
pesap, line 146: If we are not supporting this, shall we throw the error instead?

pesap (changes requested):
LGTM overall. I think we can just add consistency if we want the `check` functions to create the variable `is_valid` or not. I think the only one that does not follow that convention is `check_rating_values`. Also, I think it is better practice to just throw the error for non-supported data instead of just returning false (see `calculate_rate_limits!).

src/utils/IO/branchdata_checks.jl
jd-lara: How can a rating be negative? What's the interpretation of a negative power value for rating?

src/utils/IO/branchdata_checks.jl
jd-lara: if false is returned it fails after

Issue #1017, opened 2023-09-18
can we have a function to fix the line/circuit breaker thing the raw file 0 -> 99999 ?


jd-lara:

leorese:
There are two different issues described here, as follows:
1. Treatment of 0 MVA branch rates
2. Treatment of zero impedance lines
Suggested treatment for each of those issues is given as follows.

**0 MVA branch rates**
In PSS/E, if RATEX = 0.0, PSS/E will bypass checks for this branch (this branch will not be included in any examination of circuit loading).
In case of Sienna, it is suggested that the branch rate be parsed with the exact same value as in PSS/E (0.0). During the network constraints formulation, if branch rate is equal to 0.0, no branch limit constraint should be assigned to that branch.
Another alternative would be to set the branch rate to Inf. This is equivalent of not enforcing branch limits to the specific branch. However, this option is not recommended.

**Zero impedance lines**
In PSS/E, zero impedance lines are defined as having the following characteristics:
- Its resistance must be zero.
- Its magnitude of reactance must be less than or equal to the zero impedance line threshold tolerance, THRSHZ.
- It must be a non-transformer branch.
During network solutions, buses connected by such lines are treated as the same bus, thus having
identical bus voltages. At the completion of each solution, the loadings on zero impedance lines are
determined. When obtaining power flow solutions, zero impedance line flows, as calculated at the end of the
solution, are preserved with the working case and are available to the power flow solution reporting
activities.
There are two options to model zero impedance branches in Sienna, as follows:
1. Implement a topological pre-processor when formulating the network matrices and replicate the treatment made by PSS/E (merge nodes into a single one and post-process flows from the DC power flow solution)
2. Explicitly model the zero impedance branches using the formulation presented in: https://ieeexplore.ieee.org/document/117003 (a more comprehensive presentation of this formulation, for both AC and DC PF formulations, is given here: https://simoes.sites.ufsc.br/assp/z_zero.pdf)
Attention must be made to the check of circuit breaker status (open/closed) when building the network matrices.

jd-lara:
closed in a PR within PNM

PR #1315, opened 2025-03-26
add new struct for switch/breaker

### LINKED ISSUE #821: PSSe switched shunts parsed as fixed shunts.
When parsing PSSe power flow data switched shunts are added as fixed ones using the `Binit` value as a fixed value. If the control mode is 1, this value might not match the results from the power flow. 

The mismatch in reactive power injections leads to possible power flow failures. For instance when solving the WECC 240 bus case. 

cc. @ccoffrin this is also true when solving a PF with PowerModels.

jd-lara:
closes #821 and partially #1017 

@mcllerena if this design works we can start implementing the parser for switches as described in #1017. 
@GabrielKS please check if this is enough to export to PSSe v33


rodrigomha (approved):
This is good. We need to scope the parsing in a new PR

PR #1325, opened 2025-03-31
Change name for FixedOutage

m-bossart:
should the name be `FixedOutage` or `FixedForcedOutage`?

### DIFF
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -259,7 +259,7 @@ export Contingency
 export Outage
 export GeometricDistributionForcedOutage
 export PlannedOutage
-export TimeSeriesForcedOutage
+export FixedOutage
 
 export get_mean_time_to_recovery
 export get_outage_transition_probability
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -259,7 +259,7 @@ export Contingency
 export Outage
 export GeometricDistributionForcedOutage
 export PlannedOutage
-export FixedOutage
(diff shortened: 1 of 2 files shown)


scdhulipala:
> FixedForcedOutage seems better to me @scdhulipala any thoughts?

I fee like this depends on how this ends up getting modeled maybe? If the model sees this coming in the look ahead and can plan around it FixedOutage is more apt, if we are modeling this like a true forced outage but instead of drawing we have a time series of 0 and 1 to indicate this FixedForcedOutage probably makes more sense.

jd-lara:
> > FixedForcedOutage seems better to me @scdhulipala any thoughts?
> 
> I fee like this depends on how this ends up getting modeled maybe? If the model sees this coming in the look ahead and can plan around it FixedOutage is more apt, if we are modeling this like a true forced outage but instead of drawing we have a time series of 0 and 1 to indicate this FixedForcedOutage probably makes more sense.

Let's go with FixedForcedOutage.

src/PowerSystems.jl
```diff
@@ -259,7 +259,7 @@ export Contingency
 export Outage
 export GeometricDistributionForcedOutage
 export PlannedOutage
-export TimeSeriesForcedOutage
+export FixedOutage
```
jd-lara, line 262:
```suggestion
export FixedForcedOutage
```

jd-lara (changes requested):
Lets go with FixedForcedOutage

jd-lara (approved):

Issue #1040, opened 2024-01-11
PSS/E Parser - Branch Voltage endpoints


amirmm11:
Considering a voltage difference (tolerance) in both ends when parsing branches. 

in 
 ``` julia
 function check_endpoint_voltages(line)
    is_valid = true
    arc = get_arc(line)
    if get_base_voltage(get_from(arc)) != get_base_voltage(get_to(arc))
        is_valid = false
        @error "Voltage endpoints of $(line.name) are different, cannot create Line"
    end
    return is_valid
end
```

to 

``` julia
function check_endpoint_voltages(line, tolerance=0.02)
    is_valid = true
    arc = get_arc(line)
    from_voltage = get_base_voltage(get_from(arc))
    to_voltage = get_base_voltage(get_to(arc))
    percent_difference = abs(from_voltage - to_voltage) / ((from_voltage + to_voltage) / 2)
    if percent_difference > tolerance
        is_valid = false
        @error "Voltage endpoints of $(line.name) have more than $(tolerance*100)% difference, cannot create Line"
    end

    return is_valid
end
```

jd-lara:
@claytonpbarrows any thoughts on this?

PR #1329, opened 2025-04-02
Bug fixes for reading MMWG files

mcllerena:
- The new keys "psw" and "qsw" where added in the parser as the active and reactive power flow respectively, since these in particular are necessary for parsing the file.
-  There were some problems creating the key "ext" when converting a branch to a 2W transformer for bus voltage differences.
     - A small logic was added to keep considering the elements as branches if the threshold of 1kV is not exceeded, otherwise it will be considered as a 2W transformer and an empty "ext" dict is added.
     - Improve of the warning statement to see the voltage difference and in which buses this is happening.

### DIFF
--- a/src/parsers/pm_io/psse.jl
+++ b/src/parsers/pm_io/psse.jl
@@ -1290,6 +1290,8 @@ function _build_switch_breaker_sub_data(
     sub_data["state"] = pop!(branch, "ST")
     sub_data["active_power_flow"] = 0.0
     sub_data["reactive_power_flow"] = 0.0
+    sub_data["psw"] = sub_data["active_power_flow"]
+    sub_data["qsw"] = sub_data["reactive_power_flow"]
     sub_data["rating"] = pop!(branch, "RATEA")
     sub_data["discrete_branch_type"] = discrete_branch_type
     sub_data["source_id"] =
--- a/src/parsers/power_models_data.jl
+++ b/src/parsers/power_models_data.jl
@@ -709,6 +709,7 @@ function make_branch(name::String, d::Dict, bus_f::ACBus, bus_t::ACBus, source_t
     primary_shunt = d["b_fr"]
     alpha = d["shift"]
     branch_type = get_branch_type(d["tap"], alpha, d["transformer"])
+
     if d["transformer"]
         if branch_type == Line
             throw(DataFormatError("Data is mismatched; this cannot be a line. $d"))
@@ -827,9 +828,8 @@ function make_transformer_2w(
        get_bustype(bus_t) == ACBusTypes.ISOLATED
         available_value = false
     end
-    # println(name)
+
     ext = source_type == "pti" ? d["ext"] : Dict{String, Any}()
-    # println(d["ext"])
 
     return Transformer2W(;
         name = name,
--- a/src/parsers/power_models_data.jl
+++ b/src/parsers/power_models_data.jl
@@ -83,11 +83,12 @@ end
 
 function correct_pm_transformer_status!(pm_data::PowerModelsData)
     for (k, branch) in pm_data.data["branch"]
-        if !branch["transformer"] &&
-           pm_data.data["bus"][branch["f_bus"]]["base_kv"] !=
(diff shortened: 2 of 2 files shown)


src/parsers/pm_io/psse.jl
```diff
@@ -1290,6 +1290,8 @@ function _build_switch_breaker_sub_data(
     sub_data["state"] = pop!(branch, "ST")
     sub_data["active_power_flow"] = 0.0
     sub_data["reactive_power_flow"] = 0.0
+    sub_data["psw"] = sub_data["active_power_flow"]
+    sub_data["qsw"] = sub_data["reactive_power_flow"]
```
m-bossart, line 1294: How are these different than the two lines above?

m-bossart: Which are set to zero

m-bossart:
@mcllerena Can you comment more on why the additional logic is needed and why the threshold would be 1kV?

src/parsers/pm_io/psse.jl
```diff
@@ -1290,6 +1290,8 @@ function _build_switch_breaker_sub_data(
     sub_data["state"] = pop!(branch, "ST")
     sub_data["active_power_flow"] = 0.0
```
jd-lara, line 1291: There is something off here. Aren't we supposed to be reading the data not setting to 0.0

src/parsers/power_models_data.jl
```diff
@@ -83,11 +83,12 @@ end
 
 function correct_pm_transformer_status!(pm_data::PowerModelsData)
     for (k, branch) in pm_data.data["branch"]
-        if !branch["transformer"] &&
-           pm_data.data["bus"][branch["f_bus"]]["base_kv"] !=
-           pm_data.data["bus"][branch["t_bus"]]["base_kv"]
+        f_bus_bvolt = pm_data.data["bus"][branch["f_bus"]]["base_kv"]
+        t_bus_bvolt = pm_data.data["bus"][branch["t_bus"]]["base_kv"]
+        if !branch["transformer"] && !isapprox(f_bus_bvolt, t_bus_bvolt; atol = 1.0)
```
jd-lara, line 88: This tolerance needs to be added as a constant

src/parsers/pm_io/psse.jl
mcllerena: They are the same, is just that the keys "psw" and "qsw" are used entirely in the data.jl file. It is to avoid changing it all over that file. There are two possibilities:
- Updating all appearances on data.jl
- Fix the struct replacing active_power_flow and reactive_power_flow with the original "psw" and "qsw"

src/parsers/pm_io/psse.jl
mcllerena: There are no fields on the original data indicating the active or reactive power flow. They are set with 0.0 as initial values. Otherwise, it may be set to the rating of the line itself as an initial value.

src/parsers/pm_io/psse.jl
jd-lara: ok I understand this is to be used in the switches that do have those keys.

mcllerena:
> @mcllerena Can you comment more on why the additional logic is needed and why the threshold would be 1kV?

- When reading the EI file there are small and huge differences between the voltages of buses connected by a branch. To handle the small differences, a threshold can be defined to avoid converting them into a 2W transformer.
- For the huge differences (i.e. 999kV and 23.8kV) the branch correction is executed and an empty "ext" dict key is created since and error is thrown otherwise.

jd-lara:
> @mcllerena Can you comment more on why the additional logic is needed and why the threshold would be 1kV?

this is related to #1040

src/parsers/power_models_data.jl
mcllerena: Set on commit https://github.com/NREL-Sienna/PowerSystems.jl/pull/1329/commits/d059500c51e7bf146fbcbd14598052657b3a7e2d

jd-lara (approved):

PR #1330, opened 2025-04-03
Initial handling and parsing for Multi-Section Lines

mcllerena:
There are some considerations for parsing & building the system regarding Multi-section lines:
- Dummy buses are considered passive nodes forming a multi-section line grouping.
- Multi-section lines are treated as a single entity regarding its service status. So the availability of each branch connecting a dummy bus is done, and if any of them is disconnected, then the entire multi-section line is disconnected.
- FACTs devices must not be connected to dummy buses.
- A dummy bus must not be a converter bus (LCC, VSC).
- The field `section_number` represents the segments a the line is divided into (i.e. if section_number=1, the multi-section line is treated as a single element in the power flow)


jd-lara (commented):
@mcllerena in these cases I rather just ignore the "dummy" buses and throw a warning that those are ignored and the line is a added as a regular line. The reason for this is to avoid having to add formulations to PSI.

mcllerena:
In these commits:
- The MultiSectionLine struct was removed.
- Each multi-section line group was added as a regular line.
- To be defined as a regular line, the equivalent of r, x and b using each line section is calculated.
- The ratings considered are the max value of the ratings in each line section. 
- The line sections connected to a dummy bus were not considered when creating the system.

jd-lara (approved):

PR #1345, opened 2025-04-17
Fix matpower files parsing bug and update tests

mcllerena:
- There's were some issues with the tests due to the parsing of the area interchanges, since matpower files does not define these properly.
- Some tests were updated to match the switch shunts naming.

### DIFF
--- a/src/parsers/power_models_data.jl
+++ b/src/parsers/power_models_data.jl
@@ -309,7 +309,7 @@ function read_bus!(sys::System, data::Dict; kwargs...)
             name = "$(area_from_name)_$(area_to_name)"
             available = true
             active_power_flow = d["power_transfer"]
-            flow_limits = (from_to = Inf, to_from = Inf)
+            flow_limits = (from_to = -INFINITE_BOUND, to_from = INFINITE_BOUND)
 
             ext = Dict{String, Any}(
                 "index" => d["index"],
--- a/src/parsers/power_models_data.jl
+++ b/src/parsers/power_models_data.jl
@@ -271,18 +271,18 @@ function read_bus!(sys::System, data::Dict; kwargs...)
             "PTOL" => "",
         )
         if data["source_type"] == "pti" && haskey(data, "area_interchange")
-        for (_, area_data) in data["area_interchange"]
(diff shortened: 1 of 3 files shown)


mcllerena:
@jd-lara in commit https://github.com/NREL-Sienna/PowerSystems.jl/pull/1345/commits/e6609f2b82cd04ca1f217c32971bb32c4b2078f1 I commented out the line where checking for ac transmission values was throwing an error, wasn't sure exactly where the problem lied.
With these all tests are passing in psy5

src/parsers/power_models_data.jl
```diff
@@ -295,35 +297,37 @@ function read_bus!(sys::System, data::Dict; kwargs...)
...
+            to_area = get_component(Area, sys, area_to_name)
 
-        name = "$(area_from_name)_$(area_to_name)"
-        available = true
-        active_power_flow = d["power_transfer"]
-        flow_limits = (from_to = Inf, to_from = Inf)
+            name = "$(area_from_name)_$(area_to_name)"
+            available = true
+            active_power_flow = d["power_transfer"]
+            flow_limits = (from_to = Inf, to_from = Inf)
```
jd-lara, line 312: I am not sure we can serielize/deserialize `Inf` so we need to use the `INFINITE_BOUND` constant 
```suggestion
            flow_limits = (from_to = INFINITE_BOUND, to_from = INFINITE_BOUND)
```

jd-lara (changes requested):
use `INFINITE_BOUND` instead of INF. We need a test that parses the file, writes to JSON and reads back.

mcllerena:
> use `INFINITE_BOUND` instead of INF. We need a test that parses the file, writes to JSON and reads back.

Change is done in commit https://github.com/NREL-Sienna/PowerSystems.jl/pull/1345/commits/1fba9f8b4954fce06d5eca9eb0e9150bfeb8fece.
An the serialization/deserialization test is done in https://github.com/NREL-Sienna/PowerSystems.jl/pull/1347/commits/e22be2a902a2faf3aaf5a7985fc3b3aa7d0af76b

jd-lara (approved):

PR #1347, opened 2025-04-22
Handle Transformer Impedance Correction Tables parsing from PSS/e

mcllerena:


jd-lara (changes requested):
We don't need to change the transformer structs, the map to the correction table number would be stored in the Supplement Attribute

mcllerena:
> We don't need to change the transformer structs, the map to the correction table number would be stored in the Supplement Attribute

Sorry about that, it is reverted back, his code was done before the meeting. I'll work on the supplemental attribute now

src/impedance_correction.jl
```diff
@@ -0,0 +1,41 @@
...
+
+# Arguments
+- `ict_row::Int64`: Row number of the impedance correction table to linked it with a specific Transformer component.
+- `internal::InfrastructureSystemsInternal`: power system internal reference, do not modify
+"""
+struct ImpedanceCorrectionTable <: SupplementalAttribute
+    table_number::String
+    function_data::PiecewiseLinearData
+    subcategory::String
+    type::String
```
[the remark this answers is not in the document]
GabrielKS: Okay it looks like `type` is for turns ratio vs. phase shift angle. Let's use an `@ScopedEnum` for that.

src/parsers/power_models_data.jl
```diff
@@ -162,6 +162,95 @@ function _get_pm_3w_name(
...
+        ]
+
+        if length(x) == length(y)
+            lookup[table_number] =
+                PiecewiseLinearData([(x[i], y[i]) for i in eachindex(x)])
+            table_type =
+                (x[1] >= 0.5 && x[1] <= 1.5) ? "turns_ratio" : "phase_shift_angle"
+            type_lookup[string(table_number)] = table_type
+        else
+            @warn "Impedance correction mismatch at table $table_number: tap and scale count differs"
```
GabrielKS, line 204: I'm not super familiar with the conventions on warning vs. erroring in this parser, but shouldn't this at least be an `@error`, if not a `throw(error)`?

src/parsers/power_models_data.jl
```diff
@@ -162,6 +162,95 @@ function _get_pm_3w_name(
...
+function _attach_impedance_correction_tables!(
+    sys::System,
+    value::Union{Transformer2W, Transformer3W},
+    name::String,
+    d::Dict,
+    ict_lookup::Dict{String, PiecewiseLinearData},
+    type_lookup::Dict{String, String},
+)
+    # Match for windings in Transformer 2W and 3W
+    keys_and_categories = if isa(value, Transformer3W)
```
GabrielKS, line 223: do this with multiple dispatch

src/parsers/power_models_data.jl
```diff
@@ -162,6 +162,95 @@ function _get_pm_3w_name(
...
+        zip(("",), ("correction_table",))
+    end
+
+    for (subcategory, key) in keys_and_categories
+        if haskey(d, key)
+            table_number = string(d[key])
+            if haskey(ict_lookup, table_number)
+                pwl_data = ict_lookup[table_number]
+                table_type = get(type_lookup, table_number, "phase_shift_angle")
+                ict = ImpedanceCorrectionTable(;
```
GabrielKS, line 240: It looks like we're making a new `ImpedanceCorrectionTable` instance for every transformer? I think the design we had settled on was to make one `ImpedanceCorrectionTable` instance for each row of the table (though see my note about renaming) and then if multiple transformers share the same ICT entry, associate them with that same supplemental attribute.

src/impedance_correction.jl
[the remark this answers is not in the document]
GabrielKS: and it looks like `subcategory` is `primary_correction_table`, etc. — that can also be an `@ScopedEnum`.

src/impedance_correction.jl
```diff
@@ -0,0 +1,41 @@
+"""
+Attribute that contains information regarding the Transformer Impedance Correction Table.
+
+# Arguments
+- `ict_row::Int64`: Row number of the impedance correction table to linked it with a specific Transformer component.
+- `internal::InfrastructureSystemsInternal`: power system internal reference, do not modify
+"""
+struct ImpedanceCorrectionTable <: SupplementalAttribute
```
GabrielKS, line 8: I think this represents just one row of the impedance correction table, not the entire table, right? Maybe the name should reflect that

src/parsers/power_models_data.jl
```diff
@@ -162,6 +162,95 @@ function _get_pm_3w_name(
...
+                    ict_row = table_number,
+                    function_data = pwl_data,
+                    subcategory = subcategory,
+                    type = table_type,
+                )
+                add_supplemental_attribute!(sys, value, ict)
+            else
+                @warn "Transformer $name references missing correction table $table_number"
```
GabrielKS, line 248: is this an `@error`?

test/test_parse_psse.jl
```diff
@@ -201,3 +201,32 @@ end
     @test isapprox(get_inverter_extinction_angle_limits(lcc).max, pi / 2)
     @test get_power_mode(lcc)
 end
+
+@testset "PSSE Impedance Correction Table Parsing" begin
+    sys = build_system(PSSEParsingTestSystems, "pti_frankenstein_70_sys")
+
+    tr2w = first(get_components(Transformer2W, sys))
+
+    t_dict = Dict("T3" => 1.02, "T1" => 0.88, "T2" => 1.0, "T4" => 1.17)
```
GabrielKS, line 210: nitpick: present the keys in sorted order for readability

test/test_parse_psse.jl
```diff
@@ -201,3 +201,32 @@ end
...
+        type = all(0.5 .<= xs .<= 1.5) ? "turns_ratio" : "phase_shift",
+    )
+
+    add_supplemental_attribute!(sys, tr2w, ict2w)
+
+    tr2w_supp_attr = get_supplemental_attributes(tr2w)
+
+    @test PSY.get_table_number(tr2w_supp_attr[1]) == "1"
+    @test PSY.get_function_data(tr2w_supp_attr[1])[1] == (x = 0.88, y = 1.12)
+    @test PSY.get_type(tr2w_supp_attr[1]) == "turns_ratio"
```
GabrielKS, line 231: This doesn't really seem to be testing the parsing, just some getters? A full test of the parsing would involve parsing a system with an ICT, ensuring that the correct table row supplemental attributes are added, ensuring that the associations between transformers and table rows are correct, etc.

GabrielKS: (and I think full testing is pretty important in this case)

src/parsers/power_models_data.jl
jd-lara: yes, it should be

src/impedance_correction.jl
jd-lara: Yes, this should be `ImpedanceCorrectionData`

src/impedance_correction.jl
```diff
@@ -0,0 +1,41 @@
...
+    table_number::String
+    function_data::PiecewiseLinearData
```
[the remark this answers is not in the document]
jd-lara: `impedance_correction_function_data`

jd-lara:
@mcllerena looks like you might also need to run the formatter

test/test_parse_psse.jl
jd-lara, line 1: We should add to this test a serialize/deserialize call.

src/impedance_correction.jl
```diff
@@ -0,0 +1,46 @@
...
+# Arguments
+- `table_number::Int64`: Row number of the ICT to be linked with a specific Transformer component.
+- `impedance_correction_function_data::PiecewiseLinearData`: Function to define intervals (tap ratio/angle shift) in the Transformer component.
+- `transformer_winding::String`: Indicates the winding to which the ICT is linked to for a Transformer component.
+- `transformer_control_mode::String`: Defines the control modes of the Transformer, whether is for off-nominal turns ratio or phase angle shifts.
+- `internal::InfrastructureSystemsInternal`: power system internal reference, do not modify
+"""
+struct ImpedanceCorrectionData <: SupplementalAttribute
+    table_number::Int64
+    impedance_correction_function_data::PiecewiseLinearData
```
GabrielKS, line 13: thoughts on this?
```suggestion
    impedance_correction_curve::PiecewiseLinearData
```

src/definitions.jl
```diff
@@ -148,3 +162,11 @@ const START_COST = 1e8
 const INFINITE_COST = 1e8
 const INFINITE_BOUND = 1e6
 const BRANCH_BUS_VOLTAGE_DIFFERENCE_TOL = 0.01
+
+const TAP_RATIO_UBOUND = 1.5
```
GabrielKS, line 166: Things that are specific to PSS/E parsing should be named accordingly

src/definitions.jl
```diff
@@ -148,3 +162,11 @@ const START_COST = 1e8
...
+const TAP_RATIO_LBOUND = 0.5
+
+const PHASE_SHIFT_ANGLE = 1
+const TAP_RATIO = 2
```
GabrielKS, line 170: Remove these and just use the enum values directly

test/test_parse_psse.jl
```diff
@@ -201,3 +201,75 @@ end
...
+    tr2w = first(get_components(Transformer2W, sys))
+
+    x = [0.88, 1.0, 1.020, 1.170]
+    y = [1.12, 1.0, 0.979, 0.895]
+
+    pwl2w = PiecewiseLinearData([(x[i], y[i]) for i in eachindex(x)])
+    ict2w = PSY.ImpedanceCorrectionData(;
+        table_number = 1,
+        impedance_correction_function_data = pwl2w,
+        transformer_winding = 0,
```
GabrielKS, line 217: use enum value directly

src/models/supplemental_constructors.jl
```diff
@@ -113,6 +113,23 @@ function FACTSControlDevice(
...
+    table_number,
+    impedance_correction_function_data,
+    transformer_winding::String,
+    transformer_control_mode::String,
+    internal = InfrastructureSystemsInternal(),
+)
+    return ImpedanceCorrectionData(
+        table_number,
+        impedance_correction_function_data,
+        get_enum_value(WindingCategory, transformer_winding),
```
GabrielKS, line 127: It looks like this only gets used with a non-constant once? Feel free to disagree but it seems better to convert the winding number to a `WindingCategory` in that one place rather than to define a new constructor for it. I think the motivating principle here is we should convert from not inherently meaningful numbers to very inherently meaningful enum values as soon as possible.

src/parsers/pm_io/psse.jl
```diff
@@ -1443,6 +1449,49 @@ function _psse2pm_multisection_line!(pm_data::Dict, pti_data::Dict, import_all::
     return
 end
 
+function sort_values_by_key_prefix(imp_correction::Dict{String, Any}, prefix::String)
+    sorted_values = [
+        v for (_, v) in sort(
```
GabrielKS, line 1454:
```suggestion
        last(pair) for pair in sort(
```

src/parsers/pm_io/psse.jl
```diff
@@ -1443,6 +1449,49 @@ function _psse2pm_multisection_line!(pm_data::Dict, pti_data::Dict, import_all::
...
+            [
+                (parse(Int, k[2:end]), v) for
+                (k, v) in imp_correction if startswith(k, prefix)
+            ];
+            by = x -> x[1],
```
GabrielKS, line 1459:
```suggestion
            by = first,
```

src/parsers/pm_io/psse.jl
```diff
@@ -1443,6 +1449,49 @@ function _psse2pm_multisection_line!(pm_data::Dict, pti_data::Dict, import_all::
     return
 end
 
+function sort_values_by_key_prefix(imp_correction::Dict{String, Any}, prefix::String)
```
GabrielKS, line 1452: Don't constrain yourself to a particular `keytype`:
```suggestion
function sort_values_by_key_prefix(imp_correction::Dict{String, <:Any}, prefix::String)
```

src/parsers/pm_io/psse.jl
```diff
@@ -1443,6 +1449,49 @@ function _psse2pm_multisection_line!(pm_data::Dict, pti_data::Dict, import_all::
...
+        v for (_, v) in sort(
+            [
+                (parse(Int, k[2:end]), v) for
+                (k, v) in imp_correction if startswith(k, prefix)
+            ];
+            by = x -> x[1],
+        )
+    ]
+
+    while !isempty(sorted_values) && sorted_values[end] == 0.0
```
GabrielKS, line 1463: Could this be done more simply with `findfirst` or similar?

src/parsers/power_models_data.jl
```diff
@@ -162,6 +162,105 @@ function _get_pm_3w_name(
...
+        x = table_data["tap_or_angle"]
+        y = table_data["scaling_factor"]
+
+        if length(x) == length(y)
+            pwl_data = PiecewiseLinearData([(x[i], y[i]) for i in eachindex(x)])
+            pwl_lookup[table_number] = pwl_data
+
+            table_type =
+                if (x[1] >= TAP_RATIO_LBOUND && x[1] <= TAP_RATIO_UBOUND)
+                    TAP_RATIO
```
GabrielKS, line 191: yeah there's no reason to use numbers rather than enum values here, that's what enums are for

src/parsers/power_models_data.jl
```diff
@@ -162,6 +162,105 @@ function _get_pm_3w_name(
...
+            cache_key = (table_number, idx)
+            if haskey(ict_instances, cache_key)
+                ict = ict_instances[cache_key]
+                add_supplemental_attribute!(sys, transformer, ict)
+            else
+                @error "No correction table associated with transformer $name."
+            end
+        end
+    end
+end
```
GabrielKS, line 262: nitpick: reduce duplication between the two methods here

src/parsers/power_models_data.jl
```diff
@@ -1060,18 +1159,29 @@ function read_branch!(
     end
 
     _get_name = get(kwargs, :branch_name_formatter, _get_pm_branch_name)
+    ict_instances = _impedance_correction_table_lookup(data)
 
-    for (d_key, d) in data["branch"]
+    for (_, d) in data["branch"]
```
GabrielKS, line 1164:
```suggestion
    for d in values(data["branch"])
```

test/test_parse_psse.jl
```diff
@@ -201,3 +201,75 @@ end
...
+    tr2w_supp_attr = get_supplemental_attributes(tr2w)
+
+    @test PSY.get_table_number(tr2w_supp_attr[1]) == 1
+    @test PSY.get_impedance_correction_function_data(tr2w_supp_attr[1])[1] ==
+          (x = 0.88, y = 1.12)
+    @test PSY.get_transformer_control_mode(tr2w_supp_attr[1]) ==
+          TransformerControlMode.TAP_RATIO
+end
+
+@testset "PSSE System Serialize/Desearialized - Initial Test" begin
```
GabrielKS, line 238:
```suggestion
@testset "PSSE System Serialize/Deserialize - Initial Test" begin
```

test/test_parse_psse.jl
```diff
@@ -201,3 +201,75 @@ end
...
+    sa1_y = get_Y.(get_components(SwitchedAdmittance, original_sys))
+    sa2_y = get_Y.(get_components(SwitchedAdmittance, deserialized_sys))
+    @test sa1_y == sa2_y
+
+    original_data = original_sys.data.components.data
+    deserialized_data = deserialized_sys.data.components.data
+    original_keys = Set(keys(original_data))
+    deserialized_keys = Set(keys(deserialized_data))
+
+    @test original_keys == deserialized_keys
```
GabrielKS, line 262: We have `IS.compare_values` for this

test/test_parse_psse.jl
```diff
@@ -201,3 +201,75 @@ end
...
+        deser_components = deserialized_data[component_type]
+
+        @test length(orig_components) == length(deser_components)
+
+        orig_names = Set(keys(orig_components))
+        deser_names = Set(keys(deser_components))
+
+        @test orig_names == deser_names
+    end
+end
```
GabrielKS, line 275: Correct me if I'm wrong, it doesn't look like you're testing anything related to ICT parsing here? Also, my earlier comment about the lack of testing for ICT parsing still stands. Happy to advise on what that would look like if you want.

GabrielKS (changes requested):
The implementation is looking pretty good now, by far the largest issue at this point is the lack of thorough testing. Enum usage could also still be improved.

test/test_parse_psse.jl
mcllerena: I understand, I do know we are not really testing. I'm just adding this to be upgraded later when I have a system with a ICT integrated. I already made a PR with this PSTD

mcllerena: This test was just for testing the serialization and deserialization of a System, and compare values of both inside PSY.

src/impedance_correction.jl
mcllerena: Changed this to "impedance_correction_curve", I think fits it better in this context

src/definitions.jl
mcllerena: Didn't find a name for these constants in the manuals. I renamed it to min/max threshold since these are not really bounds in real tap ratios.

jd-lara (approved):
I am ok with this implementation. More testing will come when we also get more cases.

src/parsers/power_models_data.jl
```diff
@@ -162,6 +162,124 @@ function _get_pm_3w_name(
...
+        if length(x) == length(y)
+            pwl_data = PiecewiseLinearData([(x[i], y[i]) for i in eachindex(x)])
+            pwl_lookup[table_number] = pwl_data
+
+            table_type =
+                if (
+                    x[1] >= PSSE_PARSER_TAP_RATIO_LBOUND &&
+                    x[1] <= PSSE_PARSER_TAP_RATIO_UBOUND
+                )
+                    TransformerControlMode.TAP_RATIO.value
```
GabrielKS, line 194: Currently your code:

1. Refers to the `TAP_RATIO` enum value
2. Turns it into a raw number with `.value`
3. In your supplemental constructor, turns it back into the `TAP_RATIO` enum value

That doesn't make much sense. Also I don't know that we want `.value` to be part of the `@ScopedEnum` public interface (I don't think `ENUM_ITEM.value` is done anywhere else in Sienna).

In general, my recommendation would be: whenever you encounter the raw numbers in PSS/E input, _immediately_ turn them into an enum value, then only ever work with the enum values from that point forward. So all your functions will take enum values except for that very first level of parsing, whenever you print out something to debug it will always be an enum value, etc. — seems much easier to work with.

src/parsers/power_models_data.jl
```diff
@@ -162,6 +162,124 @@ function _get_pm_3w_name(
...
+    d::Dict,
+    ict_instances::Dict{Tuple{Int64, Int64}, ImpedanceCorrectionData},
+)
+    _attach_single_ict!(
+        sys,
+        transformer,
+        name,
+        d,
+        "correction_table",
+        WindingCategory.TR2W_WINDING.value,
```
GabrielKS, line 260: see above wrt avoiding `.value`

src/parsers/power_models_data.jl
```diff
@@ -162,6 +162,124 @@ function _get_pm_3w_name(
...
+function _attach_impedance_correction_tables!(
+    sys::System,
+    transformer::Transformer3W,
+    name::String,
+    d::Dict,
+    ict_instances::Dict{Tuple{Int64, Int64}, ImpedanceCorrectionData},
+)
+    winding_names = ["primary", "secondary", "tertiary"]
+
+    for idx in 1:(WindingCategory.TERTIARY_WINDING.value)
```
GabrielKS, line 277: see above, use `instances`

src/parsers/power_models_data.jl
GabrielKS: nice helper function

GabrielKS (changes requested):
A few more enum (non-)usage comments, let me know if it would be useful to chat synchronously about that.

I'm not super comfortable approving this with no real testing. I'm worried that if we merge this now, testing will be forgotten about and then down the line someone will discover a bug. Is there some reason we can't wait to merge this until there is a system available to test with? It wouldn't even require a release of PSTD+PSB yet, just a merge into the relevant `psy5` branches….

As a second best option, the data we'd need to test with is so small that you could just have a string in the PSY tests with a few transformers and a pair of ICT entries in PSS/E .raw format and try calling the relevant functions on those for now.

src/parsers/power_models_data.jl
jd-lara: yes @GabrielKS you are right that `.value` should be avoided. @mcllerena lets try complete this fix this week

mcllerena:
> A few more enum (non-)usage comments, let me know if it would be useful to chat synchronously about that.
> 
> I'm not super comfortable approving this with no real testing. I'm worried that if we merge this now, testing will be forgotten about and then down the line someone will discover a bug. Is there some reason we can't wait to merge this until there is a system available to test with? It wouldn't even require a release of PSTD+PSB yet, just a merge into the relevant `psy5` branches….
> 
> As a second best option, the data we'd need to test with is so small that you could just have a string in the PSY tests with a few transformers and a pair of ICT entries in PSS/E .raw format and try calling the relevant functions on those for now.

A .raw file was added for testing this and the .value's and some references were fixed as well.

test/test_parse_psse.jl
```diff
@@ -203,36 +203,57 @@ end
...
+    @test get_impedance_correction_curve(suppl_attr_tr2w_1).points[end] ==
+          (x = 24.1, y = 1.27)
+    @test get_transformer_winding(suppl_attr_tr2w_1) == WindingCategory.TR2W_WINDING
+    @test get_transformer_control_mode(suppl_attr_tr2w_1) ==
+          TransformerControlMode.PHASE_SHIFT_ANGLE
+
+    tr2w_2 = get_component(Transformer2W, sys, "BUS 109-BUS 104-i_1")
+    suppl_attr_tr2w_2 = get_supplemental_attributes(tr2w_2)[1]
+    @test get_table_number(suppl_attr_tr2w_2) == 1
+    @test get_impedance_correction_curve(suppl_attr_tr2w_2).points[1] ==
```
GabrielKS, line 224: dot access is not public interface, use getters here and elsewhere

test/test_parse_psse.jl
```diff
@@ -203,36 +203,57 @@ end
...
-    tr2w_supp_attr = get_supplemental_attributes(tr2w)
-
-    @test PSY.get_table_number(tr2w_supp_attr[1]) == 1
-    @test PSY.impedance_correction_curve(tr2w_supp_attr[1])[1] ==
+    base_dir = string(dirname(@__FILE__))
+    file_dir = joinpath(base_dir, "test_data", "modified_14bus_system.raw")
+    sys = System(file_dir)
+
+    tr2w_1 = get_component(Transformer2W, sys, "BUS 110-BUS 109-i_1")
+    suppl_attr_tr2w_1 = get_supplemental_attributes(tr2w_1)[1]
```
GabrielKS, line 211:
```suggestion
    suppl_attr_tr2w_1 = only(get_supplemental_attributes(tr2w_1))
```
here and elsewhere

src/parsers/power_models_data.jl
```diff
@@ -257,7 +252,7 @@ function _attach_impedance_correction_tables!(
         name,
         d,
         "correction_table",
-        WindingCategory.TR2W_WINDING.value,
+        0,
```
GabrielKS, line 255: Avoid hard-coded magic numbers by proper use of the enum

src/parsers/power_models_data.jl
```diff
@@ -162,6 +162,119 @@ function _get_pm_3w_name(
...
+                    TransformerControlMode.TAP_RATIO
+                else
+                    TransformerControlMode.PHASE_SHIFT_ANGLE
+                end
+
+            for winding_index in 0:(length(instances(WindingCategory)) - 1)
+                ict_instances[(table_number, winding_index)] = ImpedanceCorrectionData(;
+                    table_number = table_number,
+                    impedance_correction_curve = pwl_data,
+                    transformer_winding = winding_index,
```
GabrielKS, line 199: Again I'm not super happy about iterating over numbers then converting them to enum values — why not iterate over the enum values (`instances(WindingCategory)`) directly?

GabrielKS (changes requested):
If we're okay with putting that data file there, this testing approach works for me, thanks!

Enum usage has gotten better but is still not quite there. Still happy to provide more detailed guidance on that if it would be useful.

test/test_parse_psse.jl
```diff
@@ -201,3 +201,79 @@ end
...
+    @test get_table_number(suppl_attr_tr2w_2) == 1
+    @test get_impedance_correction_curve(suppl_attr_tr2w_2).points[1] ==
+          (x = 0.88, y = 1.12)
+    @test get_impedance_correction_curve(suppl_attr_tr2w_2).points[end] ==
+          (x = 1.17, y = 0.895)
+    @test get_transformer_winding(suppl_attr_tr2w_2) == WindingCategory.TR2W_WINDING
+    @test get_transformer_control_mode(suppl_attr_tr2w_2) ==
+          TransformerControlMode.TAP_RATIO
+
+    tr3w_1 = get_component(Transformer3W, sys, "BUS 109-BUS 104-BUS 107-i_1")
```
GabrielKS, line 232: Can we test the case where one three-winding transformer has ICT entries for multiple of its windings?

GabrielKS: Also

- Multiple transformers have the same ICT entry on the same winding
- Multiple transformers have the same ICT entry on different windings

PR #1349, opened 2025-04-24
New Update to Import/Export structs

rodrigomha:
Redoing it another branch since the rebase of json files is a mess lol.

### DIFF
--- a/src/models/cost_functions/ImportExportCost.jl
+++ b/src/models/cost_functions/ImportExportCost.jl
@@ -34,8 +34,8 @@ end
 ImportExportCost(;
     import_offer_curves = nothing,
     export_offer_curves = nothing,
-    energy_import_weekly_limit = INFINITE_COST, # TODO: change to INFINITE_BOUND after rebase
-    energy_export_weekly_limit = INFINITE_COST, # TODO: change to INFINITE_BOUND after rebase
+    energy_import_weekly_limit = INFINITE_BOUND,
+    energy_export_weekly_limit = INFINITE_BOUND,
     ancillary_service_offers = Vector{Service}(),
 ) = ImportExportCost(
     import_offer_curves,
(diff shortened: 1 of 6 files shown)


src/models/cost_functions/ImportExportCost.jl
```diff
@@ -0,0 +1,288 @@
...
+    }
+    "Sell Price Curves data to export power, which can be a time series of `PiecewiseStepData` or a
+    [`CostCurve`](@ref) of [`PiecewiseIncrementalCurve`](@ref)"
+    export_offer_curves::Union{
+        Nothing,
+        TimeSeriesKey,
+        CostCurve{PiecewiseIncrementalCurve},
+    }
+    "Weekly limit on the amount of energy that can be imported, defined in system base p.u-hours."
+    energy_import_weekly_limit::Float64
```
GabrielKS, line 27: Are we comfortable baking "weekly" into the struct itself? Do any other PSY struct fields specify a time duration in this way? What if someone wants to use something other than a week?

src/models/cost_functions/ImportExportCost.jl
```diff
@@ -0,0 +1,288 @@
...
+        throw(
+            ArgumentError(
+                "Import Curve does not have incremental slopes. Check slopes.",
+            ),
+        )
+    end
+end
+
+"""
+Make a CostCurve{PiecewiseIncrementalCurve} from suitable for inclusion in a ImportExportCost from a
```
GabrielKS, line 191:
```suggestion
Make a `CostCurve{PiecewiseIncrementalCurve}` suitable for inclusion in an `ImportExportCost` from a
```

GabrielKS: same elsewhere

GabrielKS (commented):
We're just about there. Only major thing here is I'm not sure we want to put `weekly` in the struct field itself.

src/models/cost_functions/ImportExportCost.jl
jd-lara: we probably need to leave this for now since we can't really make models with uneven resolutions.

PR #1354, opened 2025-04-28
Update hydro structs with height levels

rodrigomha:


jd-lara (changes requested):
we discussed using elevation.

PR #1356, opened 2025-04-29
add fixed term to renewable cost

### LINKED ISSUE #1316: Add Fixed Term to RenewableGenerationCost
As discussed in Sienna Invest meetings

rodrigomha:
Closes #1316

### DIFF
--- a/src/models/cost_functions/RenewableGenerationCost.jl
+++ b/src/models/cost_functions/RenewableGenerationCost.jl
@@ -17,7 +17,7 @@ The `variable` cost is a required parameter, but `zero(CostCurve)` can be used t
     variable::CostCurve
     "(default of 0) Cost of curtailing power represented as a [`CostCurve`](@ref)"
     curtailment_cost::CostCurve = zero(CostCurve)
-    "Fixed cost of keeping the unit online. For some cost representations this field can be duplicative"
+    "Fixed cost of keeping the unit online. For some cost representations this field can be duplicative with respect to the data in the VOM field."
     fixed::Float64 = 0.0
 end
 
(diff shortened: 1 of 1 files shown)


GabrielKS (changes requested):
I'm not in principle opposed, but I'd request some more documentation on when someone should use this versus a `variable` cost with a nonzero y-intercept, whether we guarantee that these two representations will always be simulated the same way or whether certain formulations will do different things with them, and what the units are. (And not to cause scope creep but if it would be convenient to make those same updates to `ThermalGenerationCost` at the same time that would be great….) Having just dealt with a bug caused by multiple different ways to specify things related to this in the `MarketBidCost` case, it's at the top of my mind that we should be unambiguous about these kinds of things.

jd-lara (commented):
I think we can merge this one and deal with the Thermal one separately and add documentation on handling costs as a separate PRs

src/models/cost_functions/RenewableGenerationCost.jl
```diff
@@ -17,6 +17,8 @@ The `variable` cost is a required parameter, but `zero(CostCurve)` can be used t
     variable::CostCurve
     "(default of 0) Cost of curtailing power represented as a [`CostCurve`](@ref)"
     curtailment_cost::CostCurve = zero(CostCurve)
+    "Fixed cost of keeping the unit online. For some cost representations this field can be duplicative"
```
jd-lara, line 20:
```suggestion
    "Fixed cost of keeping the unit online. For some cost representations this field can be duplicative with respect to the data in the VOM field. "
```

PR #1358, opened 2025-05-01
Adding load conforming field

mcllerena:
An is_conforming Bool field was added to the structs of StaticLoad subtypes. All tests are passing in psy5

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -2705,6 +2705,7 @@ function convert_component!(
         constant_reactive_power = get_reactive_power(old_load),
         max_constant_active_power = get_max_active_power(old_load),
         max_constant_reactive_power = get_max_active_power(old_load),
+        is_conforming = get_is_conforming(old_load),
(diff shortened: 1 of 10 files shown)


jd-lara (commented):
I don't like the name nomenclature conforming or non-conforming. It isn't clear what that means @scdhulipala we need a better semantics for what this is.

jd-lara (approved):
Ok, this is ok. We just need to document what this means better

rodrigomha:
This PR destroyed every case builder stuff. We need to make is_conforming optional
@jd-lara @mcllerena

rodrigomha:
@mcllerena We need to revert this to put the `is_conforming` at the end with an optional term (I guess true?)

jd-lara:
> @mcllerena We need to revert this to put the `is_conforming` at the end with an optional term (I guess true?)

we also need to use an enum instead

PR #1367, opened 2025-05-02
Add conformity field to all StaticLoad subtypes

mcllerena:


src/parsers/pm_io/matpower.jl
```diff
@@ -461,6 +461,13 @@ function _matpower_to_powermodels!(mp_data::Dict{String, <:Any})
         pm_data["switch"] = []
     end
 
+    # Add conformity key to bus data if not present
+    for bus in pm_data["bus"]
+        if !haskey(bus, "conformity")
+            bus["conformity"] = 1
```
jd-lara, line 467: Add a logging info message to reflect that this is being set by default

jd-lara (approved):

PR #1368, opened 2025-05-03
Jd/update ext docstring

jd-lara:


rodrigomha (approved):
I did not check all the files, but statistically, looks good

PR #1370, opened 2025-05-05
Add base power for transformers

rodrigomha:
Requires: https://github.com/NREL-Sienna/PowerSystemsTestData/pull/71

Tests are passing locally with exception of a serialization/deserialization minor issue with the ext that we have to fix later.

I have not tested extensively CaseBuilder, so still TBD to fix all the build systems.

### DIFF
--- a/src/descriptors/power_system_structs.json
+++ b/src/descriptors/power_system_structs.json
@@ -882,6 +882,17 @@
                     "needs_conversion": true
                 },
                 {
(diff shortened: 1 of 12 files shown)


jd-lara (approved):

jd-lara:
@mcllerena for some reason unrelated to this PR looks like the de-serialization of the impedance correction tables isn't working as expected.

PR #1376, opened 2025-05-06
Fix parse psse test

rodrigomha:
We were getting problems with serializing named tuples.

Do you think this approach is okay @jd-lara ? We could also add in the ext, `r_source` and `x_source` rather than `r` and `x`?


jd-lara:
@rodrigomha run the formatter

Issue #1335, opened 2025-04-03
Validation range for voltage angle


m-bossart:
The validation range for voltage angle is currently -pi/2 to pi/2. This can be exceeded for valid data from large systems. We need a mechanism to either loosen this requirement in general, modify the requirement for large systems, or bypass just this check when parsing large systems.

rbolgaryn:
Transformers can also shift angles by 180 degrees or other numbers of degrees, and you can end up with parts of the grid with bus angles in a different sector outside of that range. This kind of check we cannot have.

jd-lara:
It looks like the validation code is meant for OPF problems and not that useful for general use datasets. The quantity that matters is that angles across lines isn't more than 90 degrees.

PR #1377, opened 2025-05-06
Remove angle limits checks

rodrigomha:
Closes https://github.com/NREL-Sienna/PowerSystems.jl/issues/1335


jd-lara (commented):
I am more in favor to remove this check because the values relative to the ref bus can be larger than pi.

PR #1389, opened 2025-05-13
Handle unit conversion for branches

rodrigomha:
It goes with https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/466

TODO: 
- [x] Change json struct file to include `mva` to each existing need conversion.
- [x] Change json struct file to include `ohm` or `siemens` to new components that need them
- [x] Implement `get_multiplier_methods` to work with new units
- [x] Implement custom methods for Transformer3W.
- [x] Update parser of PSS/E for Transformers to now store the data in device base rather than system base.


src/models/components.jl
```diff
@@ -67,8 +67,7 @@ function _get_multiplier(
 ) where {T <: Branch}
     base_voltage = get_base_voltage(get_arc(c).from)
     if isnothing(base_voltage)
-        @warn "Base voltage is not defined for $(c.name). Returning in DEVICE_BASE units."
-        return 1.0
+        error("Base voltage is not defined for $(c.name).")
```
jd-lara, line 70:
```suggestion
        error("Base voltage is not defined for $(summary(c).")
```

rodrigomha:
Tests are passing locally with the IS branch:

image

PR #1398, opened 2025-05-23
Remove bus modifications during parsing 

m-bossart:
Removes the modifications of bus types depending on the connected generators during parsing.
Instead, logs errors during the system checks.

@jd-lara Should this definitely be an error and not a warning? This causes errors to be logged in various systems used in the tests:
image

### DIFF
--- a/src/utils/IO/system_checks.jl
+++ b/src/utils/IO/system_checks.jl
@@ -3,10 +3,13 @@
 
 ## Check that all the buses have a type defintion and that bus types are consistent with generator connections ##
 
-function buscheck(sys)
+function buscheck(sys::System)
     buses = get_components(ACBus, sys)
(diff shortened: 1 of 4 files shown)


src/utils/IO/system_checks.jl
```diff
@@ -1,13 +1,32 @@
 
 ### Utility Functions needed for the construction of the Power System, mostly used for consistency checking ####
 
-## Check that all the buses have a type defintion ##
-
-function buscheck(buses)
+## Check that all the buses have a type defintion and that bus types are consistent with generator connections ##
+
+function buscheck(sys)
```
jd-lara, line 6:
```suggestion
function buscheck(sys::System)
```

jd-lara (approved):

PR #1404, opened 2025-05-30
Add available to ACBus 

jd-lara:
This is another change that needs to be documented in the migration guide

### DIFF
--- a/src/descriptors/power_system_structs.json
+++ b/src/descriptors/power_system_structs.json
@@ -229,6 +229,12 @@
                     "data_type": "String"
                 },
                 {
+                    "null_value": "false",
+                    "name": "available",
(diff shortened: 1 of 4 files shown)


src/descriptors/power_system_structs.json
```diff
@@ -228,6 +228,12 @@
                     "comment": "Name of the component. Components of the same type (e.g., `PowerLoad`) must have unique names, but components of different types (e.g., `PowerLoad` and `ACBus`) can have the same name",
                     "data_type": "String"
                 },
+                {
+                    "null_value": "false",
```
m-bossart, line 232: @jd-lara should this be true?

jd-lara: I rather make the null value false to avoid that `ACBus(nothing)` gets added to a system and works.

m-bossart:
The constructor for ACBus will fail in some of the parsing code. Should these constructors always default to `available=true`?  Or is there data on the availability of buses that was previously ignored?

PR #1414, opened 2025-06-10
add defaults which get bus voltages

m-bossart:
Adds defaults to Transformer constructors for base voltages which grabs the base voltage from the connected bus.


rodrigomha (approved):
@jd-lara I think this is a better approach. In the parsing we can just populate it directly, but we can reuse the previous constructors without any issue

PR #1415, opened 2025-06-11
Add missing fields here and there: tests now pass

luke-kiernan:
A few miscellaneous fixes to make the tests pass.

### DIFF
--- a/src/parsers/pm_io/common.jl
+++ b/src/parsers/pm_io/common.jl
@@ -130,3 +130,17 @@ UNIT_SYSTEM_MAPPING = Dict(
     "NATURAL_UNITS" => IS.UnitSystem.NATURAL_UNITS,
     "NA" => nothing,
 )
+
+"""Add the \"base_voltage_primary\" and \"base_voltage_secondary\" keys to transformers that
+don't already have those fields."""
+function set_base_voltages!(pm_data::Dict{String, <:Any})
+    base_voltages = Dict{Int64, Float64}(
+        bus_ind => bus_data["base_kv"] for (bus_ind, bus_data) in pm_data["bus"]
+    )
+    for transf in values(pm_data["branch"])
+        if transf["transformer"] == true && !haskey(transf, "base_voltage_primary")
+            transf["base_voltage_primary"] = base_voltages[transf["f_bus"]]
+            transf["base_voltage_secondary"] = base_voltages[transf["t_bus"]]
+        end
+    end
+end
--- a/src/parsers/pm_io/common.jl
+++ b/src/parsers/pm_io/common.jl
@@ -130,17 +130,3 @@ UNIT_SYSTEM_MAPPING = Dict(
     "NATURAL_UNITS" => IS.UnitSystem.NATURAL_UNITS,
     "NA" => nothing,
 )
-
-"""Add the \"base_voltage_primary\" and \"base_voltage_secondary\" keys to transformers that
-don't already have those fields."""
-function set_base_voltages!(pm_data::Dict{String, <:Any})
-    base_voltages = Dict{Int64, Float64}(
-        bus_ind => bus_data["base_kv"] for (bus_ind, bus_data) in pm_data["bus"]
-    )
-    for transf in values(pm_data["branch"])
-        if transf["transformer"] == true && !haskey(transf, "base_voltage_primary")
-            transf["base_voltage_primary"] = base_voltages[transf["f_bus"]]
-            transf["base_voltage_secondary"] = base_voltages[transf["t_bus"]]
-        end
-    end
-end
--- a/src/parsers/pm_io/matpower.jl
+++ b/src/parsers/pm_io/matpower.jl
@@ -486,7 +486,16 @@ function _matpower_to_powermodels!(mp_data::Dict{String, <:Any})
 
     # use once available
     arrays_to_dicts!(pm_data)
-    set_base_voltages!(pm_data)
+
+    base_voltages = Dict{Int64, Float64}(
+        bus_ind => bus_data["base_kv"] for (bus_ind, bus_data) in pm_data["bus"]
+    )
+    for transf in values(pm_data["branch"])
+        if transf["transformer"] == true && !haskey(transf, "base_voltage_from")
+            transf["base_voltage_from"] = base_voltages[transf["f_bus"]]
+            transf["base_voltage_to"] = base_voltages[transf["t_bus"]]
+        end
+    end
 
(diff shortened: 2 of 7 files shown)


test/test_parse_psse.jl
```diff
@@ -353,8 +353,10 @@ end
     @test IS.compare_values(original_sys, deserialized_sys)
 end
 
+#TODO - add system to PSB to make this test functional
```
jd-lara: @luke-kiernan this system should be already in PSB can you uncomment this code please?

src/parsers/power_models_data.jl
```diff
@@ -1059,8 +1059,8 @@ function make_transformer_2w(
         rating_b = _get_rating("Transformer2W", name, d, "rate_b"),
         rating_c = _get_rating("Transformer2W", name, d, "rate_c"),
         base_power = d["base_power"],
-        base_voltage_primary = d["base_voltage_from"],
-        base_voltage_secondary = d["base_voltage_to"],
+        base_voltage_primary = get_base_voltage(bus_f),
+        base_voltage_secondary = get_base_voltage(bus_t),
```
m-bossart: We can't use the base voltage of the buses in the parser. We need to use the values from the dicts because they can be different

test/test_parse_psse.jl
```diff
@@ -353,8 +353,10 @@ end
...
+#=
 @testset "PSSE isolated bus handling (unavailable vs topologically isolated)" begin
-    #sys = build_system(PSSEParsingTestSystems, "isoltaed_bus_test_system"; force_build = true) #TODO - add system to PSB to make this test functional
+    #sys = build_system(PSSEParsingTestSystems, "isolated_bus_test_system"; force_build = true)
```
m-bossart: @luke-kiernan can you please open a PR to add this new system to PSB and then uncomment this test? The raw file is added to PowerSystemsTestData here: 
https://github.com/NREL-Sienna/PowerSystemsTestData/pull/78

m-bossart (changes requested):

test/test_parse_psse.jl
m-bossart: It needs to be added but the raw file is in PowerSystemsTestData

src/parsers/power_models_data.jl
luke-kiernan: I was getting "key not found" errors, which is why I added this. So use the value if it's present in the dictionary, otherwise do `get_base_voltage`?

test/test_parse_psse.jl
luke-kiernan: Sure, I can do that.

src/parsers/power_models_data.jl
m-bossart: We need to add an entry into the matpower parsing dicts with the base voltages. Since matpower doesn't seem to include this data in a transformer entry, we can get the base voltages from the buses at the matpower parsing level. We can't change it here because this code is common to both the psse and matpower parsing, so we need to use the data from the dicts.

m-bossart: The changes should come in this file: 
https://github.com/NREL-Sienna/PowerSystems.jl/blob/psy5/src/parsers/pm_io/matpower.jl

test/test_parse_psse.jl
luke-kiernan: done: [PR](https://github.com/NREL-Sienna/PowerSystemCaseBuilder.jl/pull/131) submitted to PSB

luke-kiernan:
Okay finished: should be unit-safe now.

src/parsers/power_models_data.jl
```diff
@@ -1059,8 +1059,9 @@ function make_transformer_2w(
         rating_b = _get_rating("Transformer2W", name, d, "rate_b"),
         rating_c = _get_rating("Transformer2W", name, d, "rate_c"),
         base_power = d["base_power"],
-        base_voltage_primary = d["base_voltage_from"],
-        base_voltage_secondary = d["base_voltage_to"],
+        # for psse inputs, these numbers may be different than the buses' base voltages
+        base_voltage_primary = d["base_voltage_primary"],
+        base_voltage_secondary = d["base_voltage_secondary"],
```
m-bossart: This should be changed back to `d["base_voltage_from"]` and `d["base_voltage_from"]`. This is necessary because it is possible that an object that is parsed from psse as a line will be converted to a transformer if the voltages are different. In psse, the primary corresponds to from and the secondary corresponds to to

src/parsers/pm_io/psse.jl
```diff
@@ -1695,6 +1695,7 @@ function _pti_to_powermodels!(
     _psse2pm_multisection_line!(pm_data, pti_data, import_all)
     _psse2pm_branch!(pm_data, pti_data, import_all)
     _psse2pm_transformer!(pm_data, pti_data, import_all)
+    set_base_voltages!(pm_data)
```
m-bossart: With the change above, we should be able to remove this line

src/parsers/pm_io/common.jl
```diff
@@ -130,3 +130,17 @@ UNIT_SYSTEM_MAPPING = Dict(
     "NATURAL_UNITS" => IS.UnitSystem.NATURAL_UNITS,
     "NA" => nothing,
 )
+
+"""Add the \"base_voltage_primary\" and \"base_voltage_secondary\" keys to transformers that
+don't already have those fields."""
+function set_base_voltages!(pm_data::Dict{String, <:Any})
```
m-bossart:
```suggestion
function set_transformer_base_voltages!(pm_data::Dict{String, <:Any})
```

src/parsers/pm_io/matpower.jl
```diff
@@ -486,6 +486,7 @@ function _matpower_to_powermodels!(mp_data::Dict{String, <:Any})
 
     # use once available
     arrays_to_dicts!(pm_data)
+    set_base_voltages!(pm_data)
```
m-bossart:
```suggestion
    set_transformer_base_voltages!(pm_data)
```

src/parsers/power_models_data.jl
luke-kiernan: > This should be changed back to d["base_voltage_from"] and d["base_voltage_from"].

I'm guessing you meant `d["base_voltage_from"]` and `d["base_voltage_to"]`. Edit: okay, yeah: looking at the diff, it's clear. I didn't realize I had changed the key, because I was only looking at the diff with my most recent commit, not with the base branch.

PR #1419, opened 2025-06-16
Update current pti parser to handle V35 PSSE files

mcllerena:
- Current parser is extracting correctly substation location data for parse_pti and parse_psse
- Handle of multi-line Re/Im of IC data for v35.
- Add to ext new parameters of version v35.
- Update Load and Switched Shunt structs to handle old and current file versions.

### DIFF
--- a/src/descriptors/power_system_structs.json
+++ b/src/descriptors/power_system_structs.json
@@ -2584,6 +2584,12 @@
                     "comment": "[`OperationalCost`](@ref) of interrupting load"
                 },
                 {
+                    "name": "interruptible",
+                    "comment": "Interruptible load flag, one for an interruptible load for zero for a non interruptible load.",
+                    "data_type": "Int64",
+                    "default": "0"
+                },
+                {
                     "name": "services",
                     "data_type": "Vector{Service}",
                     "comment": "Services that this device contributes to",
@@ -2900,6 +2906,13 @@
                     "data_type": "Complex{Float64}"
                 },
                 {
+                    "name": "initial_status",
+                    "comment": "Vector of initial switched shunt status, one for in-service and zero for out-of-service for block i (1 through 8)",
+                    "null_value": "Int[]",
+                    "data_type": "Vector{Int}",
+                    "default": "Int[]"
+                },
+                {
                     "name": "number_of_steps",
                     "comment": "Vector with number of steps for each adjustable shunt block. For example, `number_of_steps[2]` are the number of available steps for admittance increment at block 2.",
                     "null_value": "Int[]",
@@ -3022,6 +3035,12 @@
                     "comment": "Indicator of scalability of the load. Indicates whether the specified load is conforming or non-conforming.",
                     "data_type": "LoadConformity",
                     "default": "LoadConformity.UNDEFINED"
+                },
+                {
+                    "name": "interruptible",
+                    "comment": "Interruptible load flag, one for an interruptible load for zero for a non interruptible load.",
+                    "data_type": "Int64",
+                    "default": "0"
                 },
                 {
                     "name": "services",
@@ -3204,6 +3223,12 @@
                     "default": "LoadConformity.UNDEFINED"
                 },
                 {
+                    "name": "interruptible",
+                    "comment": "Interruptible load flag, one for an interruptible load for zero for a non interruptible load.",
+                    "data_type": "Int64",
+                    "default": "0"
+                },
+                {
                     "name": "services",
                     "data_type": "Vector{Service}",
                     "comment": "Services that this device contributes to",
@@ -3328,6 +3353,12 @@
                     "comment": "Indicator of scalability of the load. Indicates whether the specified load is conforming or non-conforming.",
                     "data_type": "LoadConformity",
                     "default": "LoadConformity.UNDEFINED"
+                },
+                {
+                    "name": "interruptible",
+                    "comment": "Interruptible load flag, one for an interruptible load for zero for a non interruptible load.",
+                    "data_type": "Int64",
+                    "default": "0"
                 },
                 {
                     "name": "services",
--- a/src/parsers/pm_io/pti.jl
+++ b/src/parsers/pm_io/pti.jl
@@ -5,7 +5,7 @@
 #####################################################################
 
 """
-A list of data file sections in the order that they appear in a PTI v33 file
+A list of data file sections in the order that they appear in a PTI v33/v35 file
 """
 const _pti_sections = [
     "CASE IDENTIFICATION",
@@ -30,6 +30,13 @@ const _pti_sections = [
     "INDUCTION MACHINE",
 ]
 
+const _pti_sections_v35 = vcat(
+    _pti_sections[1:6],
+    ["SWITCHING DEVICE"],
+    _pti_sections[7:end],
+    "SUBSTATION DATA",
+)
+
 const _transaction_dtypes = [
     ("IC", Int64),
     ("SBASE", Float64),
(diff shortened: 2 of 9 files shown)


src/descriptors/power_system_structs.json
jd-lara, line 1: we have a type interruptible load for this. We should just handle this using the correct type instead of the flag. We could add this field to the ext in those cases to know it comes from a PSSe interruptible load.

src/descriptors/power_system_structs.json
```diff
@@ -2583,6 +2583,12 @@
                     "data_type": "Union{LoadCost, MarketBidCost}",
                     "comment": "[`OperationalCost`](@ref) of interrupting load"
                 },
+                {
```
jd-lara, line 2600: See above, use the correct type of load for these cases

src/descriptors/power_system_structs.json
```diff
@@ -3023,6 +3036,12 @@
                     "data_type": "LoadConformity",
                     "default": "LoadConformity.UNDEFINED"
                 },
+                {
```
jd-lara: Same here.

mcllerena:
Added these fixes on the last commits, did the tests changing the interruptible flag on the raw file and the load types were created correctly after creating the system.

mcllerena:
Just finished adding the GI to the buses https://github.com/NREL-Sienna/PowerSystems.jl/pull/1419/commits/69a0aec607f2cb61f23d567d26a46f9e7b1d370f

jd-lara (changes requested):
Add a new structure that combines interruptible power + standard load to support PSSe interruptible flag.

src/parsers/power_models_data.jl
```diff
@@ -166,6 +170,53 @@ function _get_pm_3w_name(
...
+
+    bus_coords_lookup = Dict{Int, Tuple{Float64, Float64}}()
+
+    for (_, substation) in substation_data
+        if haskey(substation, "nodes") && haskey(substation, "latitude") &&
+           haskey(substation, "longitude")
+            lat, lon = substation["latitude"], substation["longitude"]
+            for node in substation["nodes"]
+                if haskey(node, "I")
+                    bus_coords_lookup[node["I"]] = (lat, lon)
```
daniel-thom: Could you create the GeographicInfo object here and share references across components? It could matter a lot if there are many instances of components at the same location.

src/parsers/power_models_data.jl
```diff
@@ -163,6 +167,53 @@ function _get_pm_3w_name(
...
+                    ),
+                )
+                add_supplemental_attribute!(sys, bus, geo_info)
+                buses_with_coords += 1
+            else
+                buses_without_coords += 1
+            end
+        end
+
+        @info "Added coordinates to $(buses_with_coords[]) buses, $(buses_without_coords[]) buses without coordinates"
```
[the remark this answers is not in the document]
daniel-thom: I agree with Copilot.

src/parsers/power_models_data.jl
```diff
@@ -512,13 +616,21 @@ function read_loads!(sys::System, data, bus_number_to_bus::Dict{Int, ACBus}; kwa
...
             has_component(StandardLoad, sys, get_name(load)) && throw(
                 DataFormatError(
                     "Found duplicate load names of $(get_name(load)), consider formatting names with `load_name_formatter` kwarg",
                 ),
             )
+        elseif data["source_type"] == "pti" && is_interruptible && d["interruptible"] == 1
+            load = make_interruptible_standardload(d, bus, sys_mbase; kwargs...)
+            has_component(InterruptibleStandardLoad, sys, get_name(load)) && throw(
+                DataFormatError(
+                    "Found duplicate interruptible load names of $(get_name(load)), consider formatting names with `load_name_formatter` kwarg",
```
daniel-thom: Prefer `summary(load)` over `get_name(load)` because it includes the struct name.

src/parsers/pm_io/pti.jl
```diff
@@ -1258,6 +1776,140 @@ function _get_line_elements(line::AbstractString)
...
+
+        if haskey(pti_data, section)
+            push!(pti_data[section], section_data)
+        else
+            pti_data[section] = [section_data]
+        end
+
+        return updated_line_index
+    catch message
+        throw(@error("Parsing failed at line $line_index: $(sprint(showerror, message))"))
```
daniel-thom: This doesn't look right. Do you mean to call `error()`?

mcllerena: Yes, my bad. I just fix it on the last commits.

daniel-thom: Looks like the pattern preceded you. It's odd, but still kinda works.
```
julia> throw(@error("asdfs"))
┌ Error: asdfs
└ @ Main REPL[15]:1
ERROR: nothing
Stacktrace:
 [1] top-level scope
   @ logging/logging.jl:418
```

daniel-thom: The `ERROR: nothing` is confusing though.

mcllerena: Yeah, better to use the error() or @error

src/parsers/pm_io/pti.jl
```diff
@@ -1267,83 +1919,247 @@ file (typically given by default by `get_pti_sections()`.
...
+        is_v35 = true
+    end
+
+    header_line_start = is_v35 ? 2 : 1
+    bus_data_start = is_v35 ? 25 : 4
+
+    current_dtypes = is_v35 ? _pti_dtypes_v35 : _pti_dtypes
+
+    line_index = 1
+    while line_index <= length(data_lines)
```
daniel-thom, line 1945: There is a lot of logic here and I can't say that I've reviewed all of it. I will say that I'm unsure about your change from a for loop (built-in index) to a while loop (managed index). It will be very easy for someone to miss incrementing `line_index` in some conditional. What is your reasoning here?

mcllerena: Yes, I was aware of this risk in using a while loop instead of a for loop. Is just that because I had to keep track of the line index for each components, where there were line/multi-line structures like in the Impedance Correction Data for v35 files, also for the complex parsing of the Substation Data Blocks structure in the file to extract the latitude/longitude fields; I couldn't find a solution to handle this with a for loop when I wrote the code.

daniel-thom: Ok thanks, I see that now. Can you point me at an example? I'd like to take a look at one of these files before approving.

mcllerena: Sure, I'll share it internally since the only one I have with all these characteristics is really heavy.

src/parsers/power_models_data.jl
jd-lara: Agree with this

src/parsers/pm_io/pti.jl
jd-lara: @mcllerena @error is for logging an error not throwing

src/parsers/pm_io/pti.jl
jd-lara: @mcllerena we can merge once Dan Approves but we still haven't implemented extensive testing. We should add an issue for increasing testing before release.

mcllerena: I’ll be building small tests with ICD and substation data specially, since these were the most annoying ones

src/parsers/power_models_data.jl
mcllerena: Fixed on last commits

daniel-thom (approved):

PR #1422, opened 2025-06-17
Mb/matpower transformer shunts

m-bossart:
Reflect the shunt for a matpower transformer to the primary side.


rodrigomha (approved):
This seems to be OK. In Matpower, a branch is always a transformer if it has tap different than zero right?

m-bossart:
> This seems to be OK. In Matpower, a branch is always a transformer if it has tap different than zero right?

Yes I believe so. Transformers always have non-zero tap and lines always have zero tap

Issue #1424, opened 2025-06-20
Missing scalings for some r, x, g, b, and maybe other params


hannahchubin:
As I've been going through the documentation for the new and updated objects in the model library for psy5, I noticed via the getter functions that a lot of parameters are not being scaled. Here's the list of params I've already discussed with Rodrigo, plus some others I noticed that look like they also need to be scaled:

`DiscreteControlledACBranch`: `r`, `x`
`Line`: `g`
`MonitoredLine`: `r`, `x`, `g`, `b`
`TModelHVDCLine`: `r`
`TwoTerminalLCCLine`: `r` (and maybe also `rectifier_rc`, `rectifier_xc`, `inverter_rc`, `inverter_xc`, `compounding_resistance`, `rectifier_capacitor_reactance`, `inverter_capacitor_reactance`)
`TwoTerminalVSCLine`: `g`

PR #1425, opened 2025-06-20
hc/psy5_missing_scalings

hannahchubin:
updated all r, x, g, and b params that hadn't been scaled yet but needed to be and regenerated their model files. This is fixing [issue 1424](https://github.com/NREL-Sienna/PowerSystems.jl/issues/1424)


src/models/generated/TModelHVDCLine.jl
jd-lara, line 1: Always in Natural Units

jd-lara (approved):

PR #1431, opened 2025-06-24
3WTransformer tap position correction to match PSSE

mcllerena:


src/parsers/pm_io/psse.jl
```diff
@@ -574,6 +574,33 @@ function _psse2pm_shunt!(pm_data::Dict, pti_data::Dict, import_all::Bool)
     end
 end
 
+function apply_tap_correction!(
```
m-bossart, line 577: Can we apply this function to the two winding transformer case also?

test/test_parse_psse.jl
```diff
@@ -364,6 +364,22 @@ end
     @test isapprox(tap, 0.979937; atol = 1e-6)  # Test corrected value matches PSSE
 end
 
+@testset "PSSE transformer 3w tap position correction testing" begin
+    sys = build_system(
+        PSSEParsingTestSystems,
+        "case14_3wt_tap_correction";
+        force_build = true,
+    )
+    trf_3w = get_component(TapTransformer, sys, "BUS 109-BUS 104-BUS 107-i_1")
```
m-bossart: Shouldn't this be getting a `Transformer3W`?

mcllerena: I'll do the fix.

src/parsers/pm_io/psse.jl
mcllerena: Sure, I'll apply it too.

PR #1432, opened 2025-06-24
Add get_num_components for System

daniel-thom:
This requires code currently in IS main branch + https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/474, and so CI will fail.


GabrielKS (commented):
LGTM pending https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/474

PR #1434, opened 2025-06-25
Fix logging for valid bustype and devices

m-bossart:
Fix the logic of proper combinations of bustypes and devices:

- Checks that PV and SLACK buses have either a non-electric load static injection or an hvdc connected.

- Does not check PQ buses. 

Fixes excessive logging in tests to to ACTIVG systems. If the underlying data for these systems is corrected, the tests can be updated.

### DIFF
--- a/test/test_parse_dynamics.jl
+++ b/test/test_parse_dynamics.jl
@@ -163,15 +163,20 @@ end
 
 @testset "2000-Bus Parsing" begin
     test_dir = mktempdir()
-    texas2000_raw_file = joinpath(TAMU_DIR, "ACTIVSg2000.RAW")
-    texas2000_dyr_file = joinpath(PSSE_DYR_DIR, "ACTIVSg2000_dynamics.dyr")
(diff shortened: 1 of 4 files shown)


src/utils/IO/system_checks.jl
```diff
@@ -5,31 +5,43 @@
 
 function buscheck(sys::System)
     buses = get_components(ACBus, sys)
-    bus_to_active_gens_map = Dict((i, Set{String}()) for i in get_number.(buses))
-    for gen in get_components(Generator, sys)
-        if !get_available(gen)
+    bus_to_pv_device_map = Dict((i, Set{String}()) for i in get_number.(buses))
+    for gen in get_components(StaticInjection, sys)
+        if !get_available(gen) || typeof(gen) == ElectricLoad
```
jd-lara, line 10: This makes me think we need a trait for static injectors as something like `supports_voltage_control` because we have a similar pattern in PowerFlows.jl

jd-lara (approved):

PR #1440, opened 2025-06-26
Add synchronous condenser logic to ps_table_data

mcllerena:


src/parsers/power_system_table_data.jl
```diff
@@ -1098,6 +1100,29 @@ function make_reactive_params(
...
+    gen,
+    bus,
+)
+    (reactive_power, reactive_power_limits) = make_reactive_params(gen)
+    active_power_limits =
+        (min = gen.active_power_limits_min, max = gen.active_power_limits_max)
+    rating = calculate_rating(active_power_limits, reactive_power_limits)
+    base_power = gen.base_mva
+
+    println(active_power_limits, "   ", reactive_power_limits, "   ", rating)
```
daniel-thom: Can you remove the debug print?

src/parsers/power_system_table_data.jl
```diff
@@ -1098,6 +1100,29 @@ function make_reactive_params(
     return reactive_power, reactive_power_limits
 end
 
+function make_synchronous_condenser(
```
daniel-thom: Is this the same function name as in the PSSE parser? Do we have some way of disambiguating the function calls? I'm wondering if we need a prefix on each of these "make" functions. I also see that this may be an existing problem.

mcllerena: I have changed the name for this one, didn't realize the others have different naming as well.

daniel-thom (approved):

jd-lara (approved):

PR #1443, opened 2025-06-29
fix psy5 tests1

jd-lara:
This PR fixes the tests in PSY 5 using IS#main, PSTD#master, PSB#PSY5

image

### DIFF
--- a/test/test_parse_dynamics.jl
+++ b/test/test_parse_dynamics.jl
@@ -169,6 +169,15 @@ end
             "psse_ACTIVSg2000_sys",
         )
     for g in get_components(ThermalStandard, sys)
+        if isnothing(get_dynamic_injector(g))
+            @error "ThermalStandard $(get_name(g)) should have a dynamic injector"
+        end
+        @test !isnothing(get_dynamic_injector(g))
+    end
+    for g in get_components(SynchronousCondenser, sys)
+        if isnothing(get_dynamic_injector(g))
+            @error "SynchronousCondenser $(get_name(g)) should have a dynamic injector"
+        end
         @test !isnothing(get_dynamic_injector(g))
     end
     path = joinpath(test_dir, "test_dyn_system_serialization_2000.json")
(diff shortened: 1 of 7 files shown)


test/test_parse_dynamics.jl
```diff
@@ -169,6 +169,15 @@ end
...
         )
     for g in get_components(ThermalStandard, sys)
+        if isnothing(get_dynamic_injector(g))
+            @error "ThermalStandard $(get_name(g)) should have a dynamic injector"
+        end
+        @test !isnothing(get_dynamic_injector(g))
+    end
+    for g in get_components(SynchronousCondenser, sys)
+        if isnothing(get_dynamic_injector(g))
+            @error "SynchronousCondenser $(get_name(g)) should have a dynamic injector"
```
daniel-thom, line 179: Should these be calls to `error()`? The `@error` should cause a test failure indirectly because we check for error log messages. Seems like it would be better to

test/test_parse_dynamics.jl
```diff
@@ -169,6 +169,15 @@ end
             "psse_ACTIVSg2000_sys",
         )
     for g in get_components(ThermalStandard, sys)
+        if isnothing(get_dynamic_injector(g))
+            @error "ThermalStandard $(get_name(g)) should have a dynamic injector"
```
daniel-thom, line 173: These error log messages are redundant, right?

daniel-thom (approved):
Looks good after the test log messages are fixed.

test/test_parse_dynamics.jl
jd-lara: Not really, it makes it easier to debug when the condition isn't met for the generator.

jd-lara:
> Looks good after the test log messages are fixed.

Keeping the logs makes it easier to debug when the tests fail

jd-lara:
This PR also fixed the tests in PSB 

image

PR #1444, opened 2025-06-29
interruptible standard load

jd-lara:
Use a new type of load for PSSe parsing.


src/descriptors/power_system_structs.json
```diff
@@ -2577,6 +2577,198 @@
                     },
                     "validation_action": "warn"
                 },
+                {
+                    "name": "conformity",
+                    "comment": "Indicator of scalability of the load. Indicates whether the specified load is conforming or non-conforming.",
+                    "data_type": "LoadConformity",
+                    "default": "LoadConformity.UNDEFINED"
+                },
```
mcllerena: We would need to add the "conformity" field after operation cost, since there's an error when loading PowerSystems: **optional positional arguments must occur at end**

jd-lara: ok, I need to relocate the field. My bad

jd-lara:
The constructors are fixes and the tests pass for those sets of tests 

image

src/parsers/power_models_data.jl
```diff
@@ -593,7 +624,7 @@ function read_loads!(sys::System, data, bus_number_to_bus::Dict{Int, ACBus}; kwa
                 ),
             )
         elseif data["source_type"] == "pti" && is_interruptible && d["interruptible"] == 1
-            load = make_interruptible_powerload(d, bus, sys_mbase; kwargs...)
+            load = make_interruptible_standardload(d, bus, sys_mbase; kwargs...)
             has_component(InterruptiblePowerLoad, sys, get_name(load)) && throw(
```
mcllerena: I think this should be InterruptibleStandardLoad, right?

src/parsers/power_models_data.jl
```diff
@@ -532,6 +532,37 @@ function make_interruptible_powerload(d::Dict, bus::ACBus, sys_mbase::Float64; k
...
+        impedance_active_power = d["py"],
+        impedance_reactive_power = d["qy"],
+        max_constant_active_power = d["pd"],
+        max_constant_reactive_power = d["qd"],
+        max_current_active_power = d["pi"],
+        max_current_reactive_power = d["qi"],
+        max_impedance_active_power = d["py"],
+        max_impedance_reactive_power = d["qy"],
+        base_power = sys_mbase,
+        conformity = d["conformity"],
```
mcllerena: Field "base_power" is repeated here.

jd-lara: I forgot to push

src/parsers/power_models_data.jl
mcllerena: @jd-lara this one is also missing

jd-lara: Fixed in [457dfb9](https://github.com/NREL-Sienna/PowerSystems.jl/pull/1444/commits/457dfb910512aaa670a046bf005b6de0300a8bfb)

mcllerena (approved):

PR #1445, opened 2025-06-29
add detailed log message

### LINKED ISSUE #1409: Error message when a time series is removed and the reference in the device was kept
An ambiguous error message is shown when a time series isn't handled correctly when loading a system (reading from a JSON here), a better message is needed to indicate that a timeseries was removed and the reference device is still there. 

Image

jd-lara:
closes #1409

### DIFF
--- a/src/models/serialization.jl
+++ b/src/models/serialization.jl
@@ -114,6 +114,15 @@ function deserialize_uuid_handling(field_type, val, component_cache)
             value = _vals
         else
             uuid = deserialize(Base.UUID, val)
+            if !haskey(component_cache, uuid)
+                error(
+                    "UUID $uuid not found in component cache. \
+                     This may indicate that a timeseries was removed improperly \
+                     and the reference device is still has a reference to the timeseries object. \
+                     Check the documentation for the `remove_timeseries!` function. \
+                     ",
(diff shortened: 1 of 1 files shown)


src/models/serialization.jl
```diff
@@ -114,6 +114,15 @@ function deserialize_uuid_handling(field_type, val, component_cache)
             value = _vals
         else
             uuid = deserialize(Base.UUID, val)
+            if !haskey(component_cache, uuid)
```
daniel-thom: This error relates to component composition, not time series. A user would get this error if the bus attached to a generator was removed apart from system methods (unless there was a code bug).

jd-lara: This is the location of the error in https://github.com/NREL-Sienna/PowerSystems.jl/issues/1409. Am I getting something wrong?

jd-lara: nvm I see now, this is not about time series. This was an Arc

jd-lara: corrected in [3063b51](https://github.com/NREL-Sienna/PowerSystems.jl/pull/1445/commits/3063b51f0721b08606c19dd2649b597eb8d590ef)

PR #1456, opened 2025-07-01
Convert zero impedance branches to a closed switch during parsing

m-bossart:
Requires for testing to pass:
https://github.com/NREL-Sienna/PowerSystemsTestData/pull/87 
https://github.com/NREL-Sienna/PowerSystemCaseBuilder.jl/pull/139

### DIFF
--- a/src/parsers/power_models_data.jl
+++ b/src/parsers/power_models_data.jl
@@ -1080,8 +1080,7 @@ function make_branch(
     alpha = d["shift"]
     branch_type = get_branch_type(d["tap"], alpha, d["transformer"])
     if d["br_r"] == 0.0 && d["br_x"] == 0.0
-        @warn "Branch $name has zero impedance; converting to a closed switch"
-        value = _make_closed_switch(name, d, bus_f, bus_t)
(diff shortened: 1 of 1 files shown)


src/parsers/power_models_data.jl
```diff
@@ -1079,8 +1079,10 @@ function make_branch(
     primary_shunt = d["b_fr"]
     alpha = d["shift"]
     branch_type = get_branch_type(d["tap"], alpha, d["transformer"])
-
-    if d["transformer"]
+    if d["br_r"] == 0.0 && d["br_x"] == 0.0
```
jd-lara, line 1082: Have you included the option of checking for status to set it correctly to open or closed?

m-bossart: No, right now I only match the `availability`. Are you thinking we should implement this ?
Line w/ zero impedance + `available = true` -> DiscreteControlledACBranch  w/ `available = true` and `status = closed`
Line w/ zero impedance + `available = false` -> DiscreteControlledACBranch  w/ `available = false` and `status = open`

jd-lara: This makes sense, we should record this in a log entry.

PR #1460, opened 2025-07-01
provide defaults so all PSB systems build

luke-kiernan:
Patched things so all PSB systems build properly. However, I'm leaving this as a draft because I'm not sure if the defaults I've provided are sensible or desired: maybe it's bad input. (I'm providing a default value for `ext` in `make_power_load` only, when other `make_{component}` functions don't have one...that's super arbitrary.)


jd-lara:
Several of the issues this PR tackles are fixed already, the ones about PowerLoad

PR #1462, opened 2025-07-02
Add PhaseShiftingTransformer3W struct

mcllerena:


src/models/supplemental_accessors.jl
```diff
@@ -32,7 +32,7 @@ Returns the series susceptance of a 3 winding transformer as three values
 (for each of the 3 branches) following the convention
 in power systems to define susceptance as the inverse of the imaginary part of the impedance.
 """
-function get_series_susceptance(b::Transformer3W)
+function get_series_susceptance(b::Union{Transformer3W, PhaseShiftingTransformer3W})
```
m-bossart: Shouldn't we have the tap settings in this calculation, similar to the `PhaseShiftingTransformer` and `TapTransformer`

mcllerena: I updated this so that the PST3W uses the taps for series susceptance calculation, just like the PST and added a test for the PST3W struct.
All test passing.

src/base.jl
```diff
@@ -2347,7 +2347,10 @@ function check_attached_buses(sys::System, component::Branch)
     return
 end
 
-function check_attached_buses(sys::System, component::Transformer3W)
+function check_attached_buses(
+    sys::System,
+    component::Union{Transformer3W, PhaseShiftingTransformer3W},
```
jd-lara: We can create an constant or abstract to dispatch better than using a Union type here

jd-lara:
The pr looks good, but lets merge once there is a test for pnm that validates the ybus

src/base.jl
m-bossart: I think it makes sense to add abstract types for both two winding and three winding transformers. `ThreeWindingTransformer <: ACTransmission`
`TwoWindingTransformer <: ACTransmission`
@jd-lara thoughts on this?

m-bossart:
> The pr looks good, but lets merge once there is a test for pnm that validates the ybus

I have a passing test in PNM which compares to PSSE. The only outstanding issue I think is whether to add abstract types

src/base.jl
jd-lara: yes this is good for me

PR #1463, opened 2025-07-02
add a trait to check if a device has attribute active_power

rbolgaryn:
Add a trait to be able to check if a device has active_power attribute


jd-lara (commented):
I am not crazy about this trait's name. Can you propose something more along the lines of `supports_active_power_injecton`?

src/models/devices.jl
```diff
@@ -191,3 +191,32 @@ function clear_reservoirs!(device::Device)
...
 end
+
+"""
+Check if a device has attribute 'active_power' for active power consumption or generation.
+"""
+has_active_power(d::Device) = throw(
+    IS.NotImplementedError(
+        "has_active_power not implemented for this device type: $(typeof(d))!",
+    ),
+)
```
jd-lara:
```suggestion
function has_active_power(::T) = where {T <: Device} 
throw(
    IS.NotImplementedError(
        "has_active_power not implemented for this device type $(T)",
    ),
)
end 
```

src/models/devices.jl
```diff
@@ -191,3 +191,32 @@ function clear_reservoirs!(device::Device)
...
+
+"""
+Most StaticInjection models have active power consumption or generation, so return true
+"""
+has_active_power(::StaticInjection) = true
```
jd-lara: here you are making them all default. I'd rather implement this for Generator Types and ElectricLoads separately

src/models/devices.jl
```diff
@@ -191,3 +191,32 @@ function clear_reservoirs!(device::Device)
...
+
+"""
+FixedAdmittance does not have active power consumption / generation, so return false
+"""
+has_active_power(::FixedAdmittance) = false
+
+"""
+SwitchedAdmittance does not have active power consumption / generation, so return false
+"""
+has_active_power(::SwitchedAdmittance) = false
```
jd-lara: you should also add the Facts Devices

rbolgaryn:
Also, should we export it in PowerSystems, or keep it internal?

jd-lara (commented):
I don't want to merge this until we are absolutely right that the design is correct in PFS.jl.

GabrielKS:
@rbolgaryn have you seen `PF.get_active_power_limits_for_power_flow`? Is this meant to replace that? Some more info on the intended design/why we want to go this route would be great

luke-kiernan:
> @rbolgaryn @luke-kiernan have you seen `PF.get_active_power_limits_for_power_flow`? Is this meant to replace that? Some more info on the intended design/why we want to go this route would be great

I believe the  `SwitchedAdmittance` component is what prompted this: see [PF issue #180](https://github.com/NREL-Sienna/PowerFlows.jl/issues/180). That's all I know--Roman or Jose might have more info.

luke-kiernan:
This came up again. My impression is that we want to be able to write the following in `PowerFlows.jl`:
```julia
function _get_{injections/withdrawals}!(
    sys::PSY.System, 
    # ... other args ...
)
    for device in PSY.get_components(PSY.Device, sys)
        # ... calculate `bus_ix` from network reduction ...
        if PSY.supports_active_power_{injection/withdrawal}(device)
            bus_activepower_{injections/withdrawals}[bus_ix] += PSY.get_active_power(device)
        end
        # same pattern, but with reactive power
    end
    # we need to handle constant current, constant impedance components separately, so do that here
end
```

What more needs to happen for that to work? I suppose we should add a test here in PSY: anything under the `Device` umbrella for which `supports_{re}active_power_{injection/withdrawal}` returns true must have a method called `get_{re}active_power`. Also, we currently only have `injection` variants, not `withdrawal`. Is there anything else missing?

jd-lara:
We decided to move this functionality to PowerFlows.jl

PR #1466, opened 2025-07-09
Fix issues for tests in new struct PST3W

mcllerena:


src/models/supplemental_accessors.jl
```diff
@@ -21,7 +21,7 @@ Returns the series susceptance of a controllable transformer following the conve
 in power systems to define susceptance as the inverse of the imaginary part of the impedance.
 In the case of phase shifter transformers the angle is ignored.
 """
-function get_series_susceptance(b::TwoWindingTransformer)
+function get_series_susceptance(b::Union{TapTransformer, PhaseShiftingTransformer})
```
jd-lara, line 24: This should apply to for the 2W Regular Phase Shifter no?

mcllerena: Only the ones that have the alpha tap field

mcllerena:
All tests passing in PSY and PFs.

src/models/supplemental_accessors.jl
jd-lara: PhaseShiftingTransformer has alpha 

image

mcllerena: Yeah, but Transformer2W doesn't

jd-lara: We discussed in the last meeting to include it no?

rbolgaryn: yes but it will be added in a separate PR.

src/models/branches.jl
jd-lara, line 1: I already made this fix in #1467 so that psy5 functionality is restored

mcllerena: I reverted the change of the type positioning.

PR #1470, opened 2025-07-10
Feat/component attribute pairs

### LINKED ISSUE #1464: Document attributes maps
It will be good to document all the ways that we can do mappings between attributes and devices

daniel-thom:
Companion PR to https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/471/files

Do not merge until the above IS PR is merged.

Fixes #1464

### DIFF
--- a/docs/src/explanation/supplemental_attributes.md
+++ b/docs/src/explanation/supplemental_attributes.md
@@ -15,11 +15,75 @@ flowchart LR
 
 Supplemental attributes can also contain timeseries in the same that a component can allowing the user to model time varying attributes like outage time series or weather dependent probabilities. See the section [`Working with Time Series Data`](@ref tutorial_time_series) for details on time series handling.
 
-## Getting the attributes in a component or system
+## Getting the attributes in a system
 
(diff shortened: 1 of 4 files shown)


docs/src/explanation/supplemental_attributes.md
```diff
@@ -15,11 +15,75 @@ flowchart LR
...
+for outage in get_supplemental_attributes(
+    x -> get_mean_time_to_recovery(x) >= 0.5,
+    FixedForcedOutage,
+    system,
+)
+    @show summary(outage)
+end
+```
+
+## Getting the attributes associated with a component
```
daniel-thom, line 43: We also have the inverse of this method: `get_associated_components` and `get_associated_available_components`. Should we list it here? I'm happy to add it. I'm concerned about having too many options.

jd-lara: I think we need to list both. But that isn't blocking

jd-lara:
@daniel-thom I will merge this to avoid some failing tests and make the changes to the API we discussed this afternoon in a subsequent PR

Issue #1469, opened 2025-07-10
Synchronous condenser parsing criteria


luke-kiernan:
When parsing `.raw` files in the `psy5` branch, we're turning available generators at PV buses with pg at 0.0 into synchronous condensers. Relevant code [here](https://github.com/NREL-Sienna/PowerSystems.jl/blob/psy5/src/parsers/pm_io/psse.jl#L329C1-L333C16). Is that the correct criteria? Does PSSe do the same thing?

It's causing issues downstream: e.g. run this code snippet on the `psy5` branch of `PowerFlows.jl`.
```julia
using PowerFlows, PowerSystems, PowerSystemCaseBuilder
using JSON3

function roundtrip(
    exporter::PSSEExporter,
    scenario_name::AbstractString,
    export_location::AbstractString
)
    raw_path, metadata_path = get_psse_export_paths(joinpath(export_location, scenario_name))
    write_export(exporter, scenario_name; overwrite = true)
    md = JSON3.read(metadata_path, Dict)
    return System(raw_path, md)
end


sys = build_system(PSISystems, "RTS_GMLC_DA_sys")
export_location = "foo"
exporter = PSSEExporter(sys, :v33, export_location; write_comments = true)
sys2 = roundtrip(exporter, "test_roundtrip", export_location)
```
Compare the generator counts: 30+ generators `sys` have been converted into synchronous condensers in `sys2`, just by doing an import-export round trip. This seems like a recipe for confusion.

luke-kiernan:
From discussion during a meeting, the earlier linked lines of code in `psse.jl` are fine, but `psse_metadata_reimport.jl` should take into account the extra metadata to differentiate between generators and synchronous condensers.

rodrigomha:
Shouldn't we also use that if `PT` and `PB` entries are zero to parse it as SynCon? Are we 100% sure that 0 `PG` is always a SynCon?

jd-lara:
check IREG too

PR #1471, opened 2025-07-15
Fix SC parsing logic

mcllerena:
Fixes issue https://github.com/NREL-Sienna/PowerSystems.jl/issues/1469#issuecomment-3070720793

### DIFF
--- a/src/parsers/pm_io/psse.jl
+++ b/src/parsers/pm_io/psse.jl
@@ -298,6 +298,22 @@ function transformer3W_isolated_bus_modifications!(pm_data::Dict, branch_data::D
     end
     return
 end
+
+"""
+    _is_synch_condenser(sub_data, pm_data)
+
+Returns `true` if the generator described by `sub_data` and `pm_data` meets the criteria for a synchronous condenser.
+"""
+function _is_synch_condenser(sub_data::Dict{String,Any}, pm_data::Dict{String,Any})
+    is_available = sub_data["gen_status"] == 1
+    is_zero_pg = sub_data["pg"] == 0.0
+    is_pv_bus = pm_data["bus"][sub_data["gen_bus"]]["bus_type"] == 2
+    has_q_limits = (sub_data["qmax"] != 0.0 || sub_data["qmin"] != 0.0)
+    has_zero_p_limits = (sub_data["pmax"] == 0.0 && sub_data["pmin"] == 0.0)
+
+    return is_available && is_zero_pg && is_pv_bus && has_q_limits && has_zero_p_limits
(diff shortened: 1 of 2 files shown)


src/parsers/pm_io/psse.jl
```diff
@@ -326,8 +326,11 @@ function _psse2pm_generator!(pm_data::Dict, pti_data::Dict, import_all::Bool)
             sub_data["r_source"] = pop!(gen, "ZR")
             sub_data["x_source"] = pop!(gen, "ZX")
 
-            if sub_data["gen_status"] == 1 && sub_data["pg"] == 0.0 &&
-               pm_data["bus"][sub_data["gen_bus"]]["bus_type"] == 2
+            if sub_data["gen_status"] == 1 &&
```
jd-lara: @mcllerena we should wrap this up in a function called `_is_synch_condenser` this is too long to understand what the conditional is.

src/parsers/power_models_data.jl
```diff
@@ -1002,7 +1002,7 @@ function make_synchronous_condenser(
         available = Bool(d["gen_status"]),
         bus = bus,
         reactive_power = d["qg"] * base_conversion,
-        rating = d["qmax"] * base_conversion,
+        rating = sqrt(d["qmax"]^2) * base_conversion,
```
jd-lara: why are you taking the square root of a pow 2 here?

mcllerena: For the large files, the qmax = -100 and the qmin = 100, making the rating negative. So is that or just taking the abs()

src/parsers/pm_io/psse.jl
```diff
@@ -298,6 +298,24 @@ function transformer3W_isolated_bus_modifications!(pm_data::Dict, branch_data::D
...
     return
 end
+
+"""
+    _is_synch_condenser(sub_data, pm_data)
+
+Returns `true` if the generator described by `sub_data` and `pm_data` meets the criteria for a synchronous condenser.
+"""
+function _is_synch_condenser(sub_data::Dict{String, Any}, pm_data::Dict{String, Any})
+    is_available = sub_data["gen_status"] == 1
```
jd-lara: Why are we checking for available here to determine if it's a condenser? we shouldn't included that in the criteria

mcllerena: That stayed from the first approach. It's removed now.

m-bossart (approved):
This looks good to me. Just flagging that we should add a note in the documentation / psy5 guide to describe this change

PR #1475, opened 2025-07-15
Approach to add threshold of 1e-4 for zero impedance values

mcllerena:
Addresses issue https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/issues/127


src/parsers/pm_io/psse.jl
```diff
@@ -1137,6 +1137,15 @@ function _psse2pm_transformer!(pm_data::Dict, pti_data::Dict, import_all::Bool)
                 Zx_s = 1 / 2 * (br_x23_sysbase - br_x31_sysbase + br_x12_sysbase)
                 Zx_t = 1 / 2 * (br_x31_sysbase - br_x12_sysbase + br_x23_sysbase)
 
+                # See PSSE Manual 1.15.1 on T3W Notes
```
m-bossart: Add more detail to make it easy to find: 
See PSSE Manual (Section 1.15.1 "Three-Winding Transformer Notes" of Data Formats file)

m-bossart (commented):
I want to test the impact of this change on the ybus validations for large systems. After that is done I will approve

src/parsers/pm_io/psse.jl
```diff
@@ -1137,6 +1137,15 @@ function _psse2pm_transformer!(pm_data::Dict, pti_data::Dict, import_all::Bool)
                 Zx_s = 1 / 2 * (br_x23_sysbase - br_x31_sysbase + br_x12_sysbase)
                 Zx_t = 1 / 2 * (br_x31_sysbase - br_x12_sysbase + br_x23_sysbase)
 
+                # See PSSE Manual (Section 1.15.1 "Three-Winding Transformer Notes" of Data Formats file)
+                Zr_p = Zr_p == 0.0 ? 1e-4 : Zr_p
```
jd-lara: This value should go in a constant at the top of the file

src/parsers/pm_io/psse.jl
```diff
@@ -1137,6 +1137,15 @@ function _psse2pm_transformer!(pm_data::Dict, pti_data::Dict, import_all::Bool)
                 Zx_s = 1 / 2 * (br_x23_sysbase - br_x31_sysbase + br_x12_sysbase)
                 Zx_t = 1 / 2 * (br_x31_sysbase - br_x12_sysbase + br_x23_sysbase)
 
+                # See PSSE Manual (Section 1.15.1 "Three-Winding Transformer Notes" of Data Formats file)
```
jd-lara, line 1139: We need to add a warning or info that this value was set to this value.

src/parsers/pm_io/psse.jl
```diff
@@ -1137,6 +1137,15 @@ function _psse2pm_transformer!(pm_data::Dict, pti_data::Dict, import_all::Bool)
...
+                Zr_p = Zr_p == 0.0 ? 1e-4 : Zr_p
+                Zr_s = Zr_s == 0.0 ? 1e-4 : Zr_s
+                Zr_t = Zr_t == 0.0 ? 1e-4 : Zr_t
+
```
m-bossart: I don't think the resistances should be changed according to this logic. Only the reactances.

m-bossart (changes requested):
This isn't matching the description from PSSE; only the reactances should be modified.

src/definitions.jl
```diff
@@ -183,6 +183,8 @@ const PSSE_PARSER_TAP_RATIO_UBOUND = 1.5
 const PSSE_PARSER_TAP_RATIO_LBOUND = 0.5
 const PARSER_TAP_RATIO_CORRECTION_TOL = 1e-5
 
+const ZERO_IMPEDANCE_THRESHOLD = 1e-4
```
m-bossart: Can we give this a more descriptive name. This name makes it seem like this is the treshold for all branches, however we are only using this for a very specific case of the three winding transformers

src/parsers/pm_io/psse.jl
```diff
@@ -1137,6 +1136,15 @@ function _psse2pm_transformer!(pm_data::Dict, pti_data::Dict, import_all::Bool)
                 Zx_s = 1 / 2 * (br_x23_sysbase - br_x31_sysbase + br_x12_sysbase)
                 Zx_t = 1 / 2 * (br_x31_sysbase - br_x12_sysbase + br_x23_sysbase)
 
+                # See PSSE Manual (Section 1.15.1 "Three-Winding Transformer Notes" of Data Formats file)
+                if Zx_p == 0.0 || Zx_s == 0.0 || Zx_t == 0.0
+                    @info "Zero impedance detected, setting value to zero impedance threshold $(ZERO_IMPEDANCE_THRESHOLD)"
```
m-bossart: Please also make this more descriptive and specific to 3WTransformer.

m-bossart (commented):
The Ybus test in PNM is passing

m-bossart (approved):

mcllerena:
Tests on PSY are passing as well.

jd-lara (approved):

PR #1481, opened 2025-07-18
Fix bug in InterruptibleStandardLoad parsing

mcllerena:
This could be fixed using a Union to handle SL and ISL or set it as a separate function for ISL.

### DIFF
--- a/src/models/supplemental_accessors.jl
+++ b/src/models/supplemental_accessors.jl
@@ -110,6 +110,14 @@ function get_max_active_power(d::StandardLoad)
     return total_load
 end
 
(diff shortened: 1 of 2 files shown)


src/models/supplemental_accessors.jl
```diff
@@ -110,6 +110,14 @@ function get_max_active_power(d::StandardLoad)
     return total_load
 end
 
+function get_max_active_power(d::InterruptibleStandardLoad)
+    total_load = get_max_constant_active_power(d)
+    # TODO: consider voltage
```
jd-lara, line 115: this is hard one. Lets open an issue to track it since it depends on the modeling assumptions.

PR #1482, opened 2025-07-21
Add tap correction case for CW = 2

mcllerena:
The message triggered by this is the same as what PSSE does when windings voltages (kV) are set for CW = 2.
I'll add more testing for the different CW settings.


m-bossart (commented):
It would be good to eventually test all three values of CW. This is part of a larger issue of more comprehensive parsing testing of possible data combinations

mcllerena:
> It would be good to eventually test all three values of CW. This is part of a larger issue of more comprehensive parsing testing of possible data combinations

Added to PSTD and PSCB, and the tests are updated for this two files.

jd-lara (approved):

PR #1485, opened 2025-07-22
services fixes

jd-lara:
This was kind off an annoying bug that allowed to call `add_service!` for devices that don't support services. As a result even though the `AreaInterchange` objects were missing the services field the code was adding the service to the system.

### DIFF
--- a/src/models/branches.jl
+++ b/src/models/branches.jl
@@ -19,7 +19,7 @@ abstract type TwoTerminalHVDC <: ACBranch end
 """ Supertype for all DC branches (branches that connect only DC nodes)"""
 abstract type DCBranch <: Branch end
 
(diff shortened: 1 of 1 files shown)


src/models/branches.jl
```diff
@@ -20,5 +19,9 @@ abstract type TwoTerminalHVDC <: ACBranch end
 """ Supertype for all DC branches (branches that connect only DC nodes)"""
 abstract type DCBranch <: Branch end
 
+function supports_services(::T) where {T <: ACBranch}
```
daniel-thom: Probably better style to do this:

```suggestion
function supports_services(::ACBranch)
```

daniel-thom (approved):

PR #1487, opened 2025-07-23
motor loads

jd-lara:
This PR adds the Motor loads which usually are represented in PSSe as generators using negative injections. In WECC there are motor loads with synchronous motors that provide reactive power support. While we can't really handle this automatically in the parsing, we can at least tell the user that we detected something that potentially is a motor load.


rodrigomha (approved):
This looks good to me for merge

PR #1497, opened 2025-08-05
Jd/plant attributes

jd-lara:
This is a first draft of the power plant attributes


src/plant_attribute.jl
```diff
@@ -0,0 +1,21 @@
...
+end
+
+struct HydroPowerPlant <: PowerPlant
+    name::String
+    penstock_map::Dict{Int, String}
+end
+
+struct PumpedHydroPowerPlant <: PowerPlant
+    name::String
+    penstock_map::Dict{Int, String}
```
pesap: Shouldn't we add a way to enforce that is a sequential map?

jd-lara: what do you mean a sequential map?

pesap: Are we going to allow user to pass this? penstock_map = Dict([(1, "2"), (-1, "test"), (0, "four"])

jd-lara: In theory the user wouldn't use negative numbers or zeros but yes. The idea is that the dict gets which penstock number is connected to which turbine name.

src/plant_attribute.jl
```diff
@@ -0,0 +1,157 @@
+abstract type PowerPlant <: SupplementalAttribute end
+
+"""Get `internal`."""
+get_internal(x::PowerPlant) = x.internal
+
+"""
+Attribute to represent ThermalGeneration power plants with synchronous generation.
```
GabrielKS:
```suggestion
Attribute to represent [`ThermalGen`](@ref) power plants with synchronous generation.
```
if that's what you mean. Or just "thermal generation" if you're not trying to refer to a PSY symbol

src/plant_attribute.jl
```diff
@@ -0,0 +1,157 @@
...
+For CombinedCycle plants consider using [`CombinedCycleBlock`](@ref).
```
GabrielKS: `CombinedCycle` is written like a symbol but it doesn't exist, same situation as above

src/plant_attribute.jl
```diff
@@ -0,0 +1,157 @@
...
+
+The shaft map field is used to represent shared shafts between units.
+
+# Arguments
+- `name::String`: Name of the power plant
+- `shaft_map::Dict{Int, String}`: Mapping of shaft numbers to unit names
```
GabrielKS: Nitpick: you could have two generators with the same name if they have different concrete types

jd-lara: I'll change this to have UUIDs instead and manage the addition via the API

src/definitions.jl
```diff
@@ -530,6 +530,14 @@ Enumeration representing the location of a hydro reservoir relative to its assoc
...
 " ReservoirLocation
 
+IS.@scoped_enum(
+    CombinedCycleConfiguration,
+    SingleShaftCombustionSteam = 1,
+    SeparateShaftCombustionSteam = 2,
+    DoubleCombustionOneSteam = 3,
+    TripleCombustionOneSteam = 4,
+)
+
```
jarry7, line 541: I assume this is the extensive listing of all possible configurations?

src/plant_attribute.jl
```diff
@@ -0,0 +1,157 @@
...
+end
+
+"""Get [`HydroPowerPlant`](@ref) `name`."""
+get_name(value::HydroPowerPlant) = value.name
+"""Get [`HydroPowerPlant`](@ref) `penstock_map`."""
+get_penstock_map(value::HydroPowerPlant) = value.penstock_map
+
+"""
+Attribute to represent renewable power plants.
+
```
jarry7, line 355: What is the raison d'etre for `RenewablePowerPlant`?

jarry7 (commented):
Left a few random comments to consider.

My working assumption is that this is just a mapping exercise to support data representation of each PowerPlant.  Is that accurate?

Or is this to then also eventually enable/improve simplified constraint formulation at a plant level not covered by existing components (like `HydroReservoir` and `HydroTurbine` components).

src/plant_attribute.jl
```diff
@@ -0,0 +1,693 @@
...
+Add a thermal generator to a [`CombinedCycleBlock`](@ref) by associating it with a prime mover unit type.
+Only combined-cycle prime movers (CA, CC, CT, CS) are valid.
+
+# Arguments
+- `sys::System`: The system containing the generator
+- `component::ThermalGen`: The thermal generator to add to the block
+- `attribute::CombinedCycleBlock`: The combined cycle block
+- `unit_type::PrimeMovers`: The prime mover type (must be CA, CC, CT, or CS)
+"""
+function add_supplemental_attribute!(
```
m-bossart, line 769: @jd-lara Why is `unit_type` a separate input to this method? Why not have the unit_type come from the ThermalGen so that only ThermalGens with the appropriate prime movers can have the attribute added?

m-bossart:
@jd-lara We should add an Other option for `CombinedCycleConfiguration` enum to account for non-standard options

src/plant_attribute.jl
```diff
@@ -115,47 +123,57 @@ function CombinedCycleBlock(
...
 function CombinedCycleBlock(;
     name,
     configuration,
     heat_recovery_to_steam_factor = 0.0,
-    unit_map::AbstractDict = Dict{PrimeMovers, Vector{Base.UUID}}(),
-    reverse_unit_map::AbstractDict = Dict{Base.UUID, PrimeMovers}(),
+    hrsg_ct_map::AbstractDict = Dict{Int, Vector{Base.UUID}}(),
+    hrsg_ca_map::AbstractDict = Dict{Int, Vector{Base.UUID}}(),
+    ct_hrsg_map::AbstractDict = Dict{Base.UUID, Int}(),
+    ca_hrsg_map::AbstractDict = Dict{Base.UUID, Int}(),
```
m-bossart: Shouldn't this be `Dict{Base.UUID, Vector{Int}}`. For the classic 2 x 1 setup, the CA unit needs to be mapped to both HRSGs.

jd-lara: Yes

.claude/claude.md
rodrigomha, line 1: are these claude files supposed to be getting pushed? should we add a gitignore for `.claude`

PR #1509, opened 2025-08-11
Update metadata reimport to match latest updates of PowerFlows.jl

mcllerena:
Besides updating the metadata reimport file to match recent PF updates namings and missing components inclusion, some fields are included into the ext dictionary of some of the components due to not being able to track back these missing fields from the created System.

### DIFF
--- a/src/parsers/pm_io/psse.jl
+++ b/src/parsers/pm_io/psse.jl
@@ -1414,20 +1414,7 @@ function _psse2pm_transformer!(pm_data::Dict, pti_data::Dict, import_all::Bool)
                     "ANSTAR" => transformer["ANSTAR"],
                 )
 
-                for prefix in [
-                    "COD",
-                    "CONT",
-                    "NOMV",
-                    "WINDV",
-                    "RMA",
-                    "RMI",
-                    "NTP",
-                    "VMA",
-                    "VMI",
-                    "RATA",
-                    "RATB",
-                    "RATC",
-                ]
+                for prefix in TRANSFORMER3W_PARAMETER_NAMES
                     for i in 1:length(WINDING_NAMES)
                         key = "$prefix$i"
                         sub_data["ext"][key] = transformer[key]
--- a/src/definitions.jl
+++ b/src/definitions.jl
@@ -288,3 +288,8 @@ const WINDING_NAMES = Dict(
     WindingCategory.SECONDARY_WINDING => "secondary",
     WindingCategory.TERTIARY_WINDING => "tertiary",
 )
+
+const TRANSFORMER3W_PARAMETER_NAMES = [
+    "COD", "CONT", "NOMV", "WINDV", "RMA", "RMI",
+    "NTP", "VMA", "VMI", "RATA", "RATB", "RATC",
+]


src/parsers/psse_metadata_reimport.jl
```diff
@@ -73,7 +78,9 @@ end
 function remap_bus_numbers!(sys::System, bus_number_mapping)
     for bus in collect(get_components(Bus, sys))
         old_number = get_number(bus)
-        new_number = parse(Int, bus_number_mapping[old_number])
+        # new_number = parse(Int, bus_number_mapping[old_number])
```
GabrielKS: remove?

src/parsers/psse_metadata_reimport.jl
```diff
@@ -92,12 +99,22 @@ retransformation.
...
+            md["transformer_ckt_mapping"],
+        ),
         md["bus_number_mapping"],
         Tuple{Int64, Int64},
     )
 
-    bus_name_formatter = device_dict -> bus_name_map[device_dict["name"]]
+    bus_name_formatter = device_dict -> begin
+        name = device_dict["name"]
+        if startswith(name, "starbus_")
```
GabrielKS: I'm not convinced we should hardcode special cases in the reimporter depending on the names of things. Is something like this absolutely necessary, or could we have a special case in the _exporter_ (if necessary) such that the bus name mapping preserves starbus names? (same issue as https://github.com/NREL-Sienna/PowerFlows.jl/pull/214#pullrequestreview-3139001710)

test/Project.toml
```diff
@@ -14,6 +14,7 @@ NLsolve = "2774e3e8-f4cf-5e23-947b-6d7e65073b56"
 PowerSystemCaseBuilder = "f00506e0-b84f-492a-93c2-c0a9afc4364e"
 PowerSystems = "bcd98974-b02a-5e2f-9ee0-a103f5c450dd"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
GabrielKS: Let's not add this here in this PR, the tests themselves don't depend on it

GabrielKS (changes requested):
A few small comments. I'll try running the round trip myself as soon as a test system is provided in https://github.com/NREL-Sienna/PowerFlows.jl/pull/214 unit tests

src/parsers/pm_io/psse.jl
```diff
@@ -1374,17 +1399,46 @@ function _psse2pm_transformer!(pm_data::Dict, pti_data::Dict, import_all::Bool)
...
                     "CW" => transformer["CW"],
                     "CZ" => transformer["CZ"],
                     "CM" => transformer["CM"],
+                    "MAG1" => transformer["MAG1"],
+                    "MAG2" => transformer["MAG2"],
+                    "VMSTAR" => transformer["VMSTAR"],
+                    "ANSTAR" => transformer["ANSTAR"],
                 )
 
+                for prefix in [
```
jd-lara: move this long list to a const

jd-lara (approved):
I have a few comments on making the code more readable for lgtm. I'd like to merge to test in PSI

Issue #1515, opened 2025-08-20
Implement `status` for HydroGen components


jarry7:
Similar to ThermalGen, implementing `status` for the purposes of UC.

<img width="800" height="632" alt="Image" src="https://github.com/user-attachments/assets/29a1853c-5ac7-421e-8763-92a5a493e08b" />

<img width="1206" height="1108" alt="Image" src="https://github.com/user-attachments/assets/d77718ae-0d67-4f2e-bb81-3f22b067f930" />

PR #1517, opened 2025-08-22
add turbine enum

jd-lara:
Some more changes for enums and status are pending but this the general direction for the change. 

This might break some tests... so before merge it needs to run the tests. 

fixes: #1515 #1513

### DIFF
--- a/src/models/generated/HydroPumpTurbine.jl
+++ b/src/models/generated/HydroPumpTurbine.jl
@@ -22,6 +22,8 @@ This file is auto-generated. Do not edit.
         ramp_limits::Union{Nothing, UpDown}
         time_limits::Union{Nothing, UpDown}
         base_power::Float64
+        status::Bool
+        time_at_status::Float64
         operation_cost::Union{HydroGenerationCost, MarketBidCost}
         active_power_pump::Float64
         efficiency::TurbinePump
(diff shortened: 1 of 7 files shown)


src/models/generated/HydroTurbine.jl
jarry7, line 1: Not a showstopper but what would be the envisioned enums for HydroTurbine?(`HydroPumpTurbineType.UNKNOWN` for now).

src/models/generated/HydroPumpTurbine.jl
```diff
@@ -55,14 +56,15 @@ A hydropower pumped turbine that needs to have two [`HydroReservoir`](@ref)s att
 - `ramp_limits::Union{Nothing, UpDown}`: ramp up and ramp down limits in MW/min, validation range: `(0, nothing)`
 - `time_limits::Union{Nothing, UpDown}`: Minimum up and Minimum down time limits in hours, validation range: `(0, nothing)`
 - `base_power::Float64`: Base power of the unit (MVA) for [per unitization](@ref per_unit), validation range: `(0, nothing)`
+- `status::Bool`: (default: `false`) Initial commitment condition at the start of a simulation (`true` = on or `false` = off)
```
rodrigomha: This status should be the enum

rodrigomha (changes requested):

rodrigomha (approved):

PR #1521, opened 2025-08-28
Adding a trait for supports active/reactive power

luke-kiernan:
Continuation of PR [#1463](https://github.com/NREL-Sienna/PowerSystems.jl/pull/1463), but I don't have sufficient permissions to push to the NREL remote.

Changes: instead of `supports_{re}active_power_{withdrawal/injection}`, I just went with `supports_{re}active_power`, plus another function that says if the results of `get_{re}active_power(component)` should be added (injections) or subtracted (withdrawals) from the net power.

I could use some guidance on how to handle FACTS devices: the physical devices provide reactive power regulation, but the struct doesn't simply have a `reactive_power` field.


src/models/devices.jl
```diff
@@ -195,3 +195,60 @@ function clear_reservoirs!(device::Device)
...
+# most static components inject power
+supports_active_power(::StaticInjection) = true
+active_power_contribution_sign(::StaticInjection) = 1
+# carve out the exceptions: we handle constant impedance/current loads separately
+supports_active_power(::SynchronousCondenser) = false
+supports_active_power(::FixedAdmittance) = false
+supports_active_power(::SwitchedAdmittance) = false
+supports_active_power(::FACTSControlDevice) = false
+# loads withdraw power.
+active_power_contribution_sign(::ElectricLoad) = -1
```
jd-lara: I am not sure about this. We do the multiplier in PSI for these cases and it could be dependent on the user. So I don't think we should have this in PSY

luke-kiernan: In PF, power contributions are split into withdrawals and injections: I need to know if I should include `get_active_power(device)` in each. My plan was to include it in injections if `active_power_contribution_sign(device) == 1`, withdrawals if `-1`. While it'd be natural to use these `contribution_sign` functions elsewhere, that's why I added them.

How about I switch to an enum (or scoped enum) instead? Something like `@enum PowerContributionType INJECTION = 1 WITHDRAWAL = 2`

luke-kiernan:
I believe Jose and I both hit the "close" button within seconds of each other, leading to one of them being parsed as "reopen." So I'm going to close this again. (If @jd-lara you actually meant to re-open it, re-open it again.)

PR #1524, opened 2025-09-02
get_power returns constant power

luke-kiernan:
Add a supplemental accessor, so that there's a uniform interface and we can handle all constant power contributions by calling `get_{re}active_power(device)`.

### DIFF
--- a/src/models/supplemental_accessors.jl
+++ b/src/models/supplemental_accessors.jl
@@ -166,3 +166,9 @@ function get_α_tertiary(t::Transformer3W)
         return get_tertiary_group_number(t).value * -(π / 6)
     end
 end
+
+# this way, we can handle all constant power contributions using the same function call.
+get_active_power(l::StandardLoad) = l.constant_active_power
+get_reactive_power(l::StandardLoad) = l.constant_reactive_power
+get_active_power(l::InterruptibleStandardLoad) = l.constant_active_power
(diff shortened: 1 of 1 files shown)


src/models/supplemental_accessors.jl
```diff
@@ -166,3 +166,9 @@ function get_α_tertiary(t::Transformer3W)
         return get_tertiary_group_number(t).value * -(π / 6)
     end
 end
+
+# this way, we can handle all constant power contributions using the same function call.
+get_active_power(l::StandardLoad) = l.constant_active_power
```
jd-lara, line 171: This is not correct for StandardLoads

luke-kiernan: I realize that there are other active power contributions, besides that from the `constant_active_power` field. If this is too confusing, then how about I do the reverse, and make `get_constant_active_power` alias to `get_active_power` for all other components?

The reason I have things this way: I want to be able to write things multiple dispatch style, without carving out exceptions for components. E.g.
```julia
for device in get_components(PSY.StaticInjection, sys)
    if contributes_active_power(device) && active_power_contribution_type(device) == INJECTION
        [calculate bus_ix from get_bus(device)]
        active_power_injections[bus_ix] += get_{constant}_active_power(device)
    end
end
[handle constant impedance and constant voltage here...
[...for the types of components that have those things]
```

PR #1529, opened 2025-09-05
[BREAKING] Remove HydroEnergyReservoir struct

rodrigomha:
@jarry7 @jd-lara @m-bossart @kdayday 

This is one of the major changes that we need to do to complete the Hydro updates. This will break many things and we need to debug what is breaking everywhere.

As of right now, the generator mapping will require update for the RTS to handle the EnergyReservoir that is there.


jarry7:
@rodrigomha  - ok, it seems to be what it is.

For clarity:
1. We will not be supporting `HydroEnergyReservoir` anymore (and related PSI formulations). 
2. We are replacing this with `HydroTurbine` (which then needs a `HydroReservoir` attached to it).

If so, which formulations are we confident will work at runtime right now and which in the near future (with PSI and HPS)? 

We likely do not have extensive amounts of time to debug and address issues in our large-scale models considering project-specific timelines (this is why I ask).

jd-lara (approved):

PR #1534, opened 2025-09-20
add contributing reserves for agc

m-bossart:
Changes to a three way mapping: AGC -> Reserves -> Devices.
Note: the test that the devices that ultimately participate in the AGC will be handled during model building.


test/test_services.jl
```diff
@@ -453,3 +453,21 @@ end
         [area_interchange12, area_interchange13, line],
     )
 end
+
+@testset "Test AGC" begin
```
jd-lara, line 457: @m-bossart we need a test that serializes and de-serializes a system with the AGC and the reserves. I have a sense that it might cause an issue somewhere else

PR #1537, opened 2025-09-23
Fix bugs found in V35 PJM file

### LINKED ISSUE #1536: Issue with building sys from PJM PSSE file
Using ] PowerSystems v4.6.2 `https://github.com/NREL-Sienna/PowerSystems.jl.git#psy5`
to build system from raw file with the following error. 

[ Info: Parsing PSS(R)E Two-Terminal and VSC DC line data into a PowerModels Dict...
[ Info: VSC-HVDC lines are supported via a dc line model approximated by two generators and an associated loss.
ERROR: KeyError: key "CONVERTER BUSES" not found
Stacktrace:
  [1] getindex(h::Dict{String, Any}, key::String)
    @ Base ./dict.jl:477

mcllerena:
- The V35 PJM files were missing the OWNER data, which were manually fixed.
- The 3w parameters were missing a conditional to be added to the ext.
- The IC data in PJM file is a bit different from the ones we've been working with, requiring an extra step to handle it with the parser.

closes #1536

### DIFF
--- a/src/parsers/pm_io/psse.jl
+++ b/src/parsers/pm_io/psse.jl
@@ -2007,13 +2007,18 @@ function sort_values_by_key_prefix_v35(imp_correction::Dict{String, <:Any}, pref
         )
     ]
 
-    first_non_zero_index = findfirst(x -> x != 0.0, reverse(sorted_values))
-    if first_non_zero_index !== nothing
-        sorted_values = sorted_values[1:(length(sorted_values) - first_non_zero_index + 1)]
-    end
+    # Remove trailing zeros in IC data that indicate end of entry (0.00000,0.00000,0.00000)
+    # Acoording to PSSE DataFormat Section 1.18. TIC Tables
+    # Check if the last entry is all zeros across (T, Re(F), Im(F))
+    while !isempty(sorted_values)
+        last_index = length(sorted_values)
+        keys_to_check = ["T$last_index", "Re(F$last_index)", "Im(F$last_index)"]
 
-    if length(sorted_values) < 2
-        push!(sorted_values, 0.0)
+        if all(k -> haskey(imp_correction, k) && imp_correction[k] == 0.0, keys_to_check)
+            pop!(sorted_values)
+        else
+            break
+        end
     end
 
     return sorted_values
--- a/src/parsers/pm_io/psse.jl
+++ b/src/parsers/pm_io/psse.jl
@@ -2012,6 +2012,10 @@ function sort_values_by_key_prefix_v35(imp_correction::Dict{String, <:Any}, pref
         sorted_values = sorted_values[1:(length(sorted_values) - first_non_zero_index + 1)]
     end
 
+    if length(sorted_values) < 2
+        push!(sorted_values, 0.0)
+    end
+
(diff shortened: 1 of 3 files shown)


src/parsers/pm_io/psse.jl
```diff
@@ -2008,6 +2012,10 @@ function sort_values_by_key_prefix_v35(imp_correction::Dict{String, <:Any}, pref
         sorted_values = sorted_values[1:(length(sorted_values) - first_non_zero_index + 1)]
     end
 
+    if length(sorted_values) < 2
```
jd-lara: where does the logic for this check come from?

mcllerena: I updated this with more comments to explain the issue. Basically for each entry of the TIC data, a block of (0.00, 0.00, 0.00) needs to be specified to indicate the end of that entry. And for single TIC entries, the PiecewiseLinearData function_data throws an error because it needs at least two entries to define the curve, hence the conditional to skip these.

src/parsers/pm_io/pti.jl
```diff
@@ -2096,16 +2093,12 @@ function _parse_pti_data(data_io::IO)
...
-                                re_val = parse(Float64, re_str)
-                                im_val = parse(Float64, im_str)
-
-                                if !(t_val == 0.0 && re_val == 0.0 && im_val == 0.0)
-                                    section_data_prev["T$point_index"] = t_val
-                                    section_data_prev["Re(F$point_index)"] = re_val
-                                    section_data_prev["Im(F$point_index)"] = im_val
-                                    point_index += 1
-                                end
+                                section_data_prev["T$point_index"] = parse(Float64, t_str)
```
jd-lara, line 2096: What's the data you are interpolating here?

src/parsers/power_models_data.jl
```diff
@@ -1571,18 +1575,33 @@ function make_tap_transformer(
     ext = haskey(d, "ext") ? d["ext"] : Dict{String, Any}()
     control_objective_formatter =
         get(kwargs, :transformer_control_objective_formatter, nothing)
-    control_objective =
-        if control_objective_formatter !== nothing
-            control_objective_formatter(name)
-        else
-            get(d, "COD1", -99)
-        end
+    control_objective = if control_objective_formatter !== nothing
```
jd-lara, line 1578: is this logic correct? looks like you have the case covered with the else

src/parsers/pm_io/pti.jl
mcllerena: That's specific for the "Ti" parameter of the IC data:
"_Either off-nominal turns ratio in pu of the controlling windings bus voltage base
or phase shift angle in degrees_."
I'm just looping through all entries in 3 by 3 blocks.

src/parsers/power_models_data.jl
mcllerena: The "else" is for cases when no control objective formatter is passed, I'm just passing the formatter in case there's an undefined -99 that PSSE files does not support. 
Got this bug when testing for large cases (ERCOT/WECC/EI).

PR #1539, opened 2025-10-02
Update hydro models to enable cascading models

rodrigomha:
This is the draft PR for updating models. The fundamental difference is that now the HydroReservoir is handling which turbines are upstream and downstream. I have removed the reservoirs in the Turbine and PumpTurbine, so now is possible to auto-generate the HydroReservoir struct.

I have added back the `prime_mover_type` with defaults for `HydroTurbine` and `HydroPumpTurbine`.

@pesap Can you help with the connected devices thingy so now we are able to get the connected upstream reservoirs and connected downstream reservoirs for a turbine.

TODO:
- [x] TableDataParser: @jd-lara can you help here?
- [x] Additional methods for finding reservoirs for a turbine, removing stuff, etc. Basically flip around the old methods to find the turbines in a reservoir, to now find the reservoirs in a turbine.
- [x] Tests

I will be opening PRs on TestData, CaseBuilder soon. HydroPowerSimulations will come a bit later since I need to discuss first the right approach for formulations if we want to ignore cascading or not (basic or complex modeling).

@jarry7 I have added as a reviewer so you can check what you think of the design of this.

### DIFF
--- a/test/test_power_system_table_data.jl
+++ b/test/test_power_system_table_data.jl
@@ -271,3 +271,26 @@ end
     @assert isapprox(get_proportional_term(cost_curve), 0.0, atol = 0.01)
     @assert isapprox(get_constant_term(cost_curve), 0.0, atol = 0.01)
 end
+
+@testset "Test Reservoirs and Turbines" begin
+    cdmsys = PSB.build_system(
+        PSB.PSITestSystems,
+        "test_RTS_GMLC_sys";
+        force_build = true,
+    )
+    @test !isempty(get_components(HydroTurbine, cdmsys))
+    for turbine in get_components(HydroTurbine, cdmsys)
+        reservoir = get_connected_head_reservoirs(cdmsys, turbine)
+        @test !isempty(reservoir)
+        reservoir = get_connected_tail_reservoirs(cdmsys, turbine)
+        @test isempty(reservoir)
+    end
+
+    @test !isempty(get_components(HydroReservoir, cdmsys))
+
+    for reservoir in get_components(HydroReservoir, cdmsys)
+        turbines = get_downstream_turbines(reservoir)
+        @test !isempty(turbines)
+        @test isempty(get_upstream_turbines(reservoir))
+    end
+end


pesap:
@rodrigomha  sure will do. Also, aren't we ditching the tabledata parser?

src/models/generated/HydroReservoir.jl
```diff
@@ -53,71 +52,68 @@ mutable struct HydroReservoir <: Device
...
-        initial_level=0.0,
         storage_level_limits=(min=0.0, max=0.0),
+        initial_level=0.0,
         spillage_limits=nothing,
         inflow=0.0,
         outflow=0.0,
-        level_targets=0.0,
-        travel_time=0.0,
+        level_targets=nothing,
+        travel_time=nothing,
```
pesap: `travel_time` should be a property of the turbine not the reservoir.

src/base.jl
```diff
@@ -1436,31 +1436,64 @@ function get_contributing_device_mapping(sys::System)
...
-Returns a ReservoirConnectedDevices object.
+Returns a TurbineConnectedDevices object.
 """
-function _get_connected_devices(sys::System, reservoir::T) where {T <: HydroReservoir}
-    uuid = IS.get_uuid(reservoir)
-    devices = ReservoirConnectedDevices(reservoir, Vector{Device}())
-    for device in get_components(HydroTurbine, sys)
-        for _reservoir in get_reservoirs(device)
-            if IS.get_uuid(_reservoir) == uuid
+function _get_connected_tail_devices(sys::System, turbine::T) where {T <: HydroUnit}
```
pesap, line 1540: I think we are using upstream/downstream instead of head of tail.

src/models/generated/HydroReservoir.jl
```diff
@@ -53,71 +52,68 @@ mutable struct HydroReservoir <: Device
...
         intake_elevation=0.0,
```
pesap, line 106: This is only relevant for the downstream turbine?

jd-lara: yeah, travel time needs to be moved to the turbines now

src/models/devices.jl
```diff
@@ -125,27 +125,31 @@ function clear_services!(device::Device)
     return
 end
 
+""" Ensures that reservoirs cannot provide services """
+supports_services(::HydroReservoir) = false
+
+# TODO this the other way around
```
jd-lara: @rodrigomha what's this?

src/models/devices.jl
```diff
@@ -154,47 +158,86 @@ function has_reservoir(device::HydroTurbine, reservoir::HydroReservoir)
...
     end
 
     return false
 end
 
-has_reservoir(T::Type{<:HydroReservoir}, device::Device) = has_reservoir(device, T)
+"""
+Return true if any upstream hydro unit is attached to the reservoir.
+"""
+function has_upstream_turbine(reservoir::HydroReservoir)
```
jd-lara, line 164: Does this check make sense? Is it just checking for the existence of a turbine?

src/models/devices.jl
```diff
@@ -154,47 +158,86 @@ function has_reservoir(device::HydroTurbine, reservoir::HydroReservoir)
...
+
+    return false
+end
 
 """
-Remove service from device if it is attached.
+Return true if any downstream hydro unit is attached to the reservoir.
+"""
+function has_downstream_turbine(reservoir::HydroReservoir)
+    for _turbine in get_downstream_turbines(reservoir)
```
jd-lara: can't we just return `!isempty(get_downstream_turbines(reservoir)`

src/models/supplemental_setters.jl
```diff
@@ -1,4 +1,13 @@
-function set_reservoirs!(turbine::HydroTurbine, reservoir::HydroReservoir)
-    set_reservoirs!(turbine, [reservoir])
+function set_upstream_turbines!(reservoir::HydroReservoir, turbine::HydroUnit)
+    set_upstream_turbines!(reservoir, [turbine])
     return
 end
+
+function set_downstream_turbines!(reservoir::HydroReservoir, turbine::HydroUnit)
```
jd-lara:
```suggestion
function set_downstream_turbine!(reservoir::HydroReservoir, turbine::HydroUnit)
```

src/parsers/power_system_table_data.jl
```diff
@@ -1539,6 +1539,7 @@ function make_hydro_turbine(
         outflow_limits = nothing,
         reservoirs = reservoirs,
     )
+    # TODO
```
jd-lara: What's this TODO?

rodrigomha: This TODO is part of the work that we need to fix for the TableData later. Basically change the structs to work

jd-lara (approved):

PR #1548, opened 2025-10-15
psse parse fixes

jd-lara:
Fixes the issues when parsing the breakers and adds more checks

### DIFF
--- a/test/test_subsystems.jl
+++ b/test/test_subsystems.jl
@@ -215,6 +215,9 @@ end
     device = nothing
     service = nothing
     for dev in get_components(Device, sys)
+        if !PSY.supports_services(dev)
+            continue
+        end
         services = get_services(dev)
         if !isempty(services)
             device = dev


src/parsers/pm_io/psse.jl
```diff
@@ -2159,8 +2176,8 @@ function _pti_to_powermodels!(
     _psse2pm_shunt!(pm_data, pti_data, import_all)
     _psse2pm_generator!(pm_data, pti_data, import_all)
     _psse2pm_facts!(pm_data, pti_data, import_all)
-    _psse2pm_switch_breaker!(pm_data, pti_data, import_all)
     _psse2pm_branch!(pm_data, pti_data, import_all)
```
m-bossart, line 2216: Should we add a comment here about maintaining order

src/parsers/pm_io/pti.jl
```diff
@@ -2601,6 +2602,26 @@ function parse_pti(io::IO)::Dict
...
+    breakers = sizehint!(eltype(data["BRANCH"])[], length(data["BRANCH"]))
+    delete_ixs = Int[]
+    for (ix, item) in enumerate(data["BRANCH"])
+        if first(item["CKT"]) == '@' || first(item["CKT"]) == '*'
+            push!(breakers, item)
+            push!(delete_ixs, ix)
+        end
+    end
+    if isempty(delete_ixs)
+        @info "No breakers found in the system."
```
m-bossart: There can still be breakers in the system? Just not in the branch section. If so I think this comment should change

src/parsers/pm_io/psse.jl
m-bossart: Order no longer matters

src/parsers/pm_io/pti.jl
jd-lara: resolved in [172687d](https://github.com/NREL-Sienna/PowerSystems.jl/pull/1548/commits/172687d601b0c0ddcb49055eaa1867a3d2b19284)

PR #1560, opened 2025-11-02
Add full time-variable `ImportExportCost` interface

GabrielKS:
There's quite a lot of stuff in `cost_function_timeseries.jl` to support interaction with `MarketBidCost`, and I think we should implement the corresponding methods for `ImportExportCost`. So far I'm focusing on the methods that are necessary to support PowerSimulations, but we should also do the user-facing ones.

### DIFF
--- a/src/models/cost_function_timeseries.jl
+++ b/src/models/cost_function_timeseries.jl
@@ -1,17 +1,13 @@
-const OfferCurveCost = Union{MarketBidCost, ImportExportCost}
-
 # VALIDATORS
 function _validate_market_bid_cost(cost, context)
     (cost isa MarketBidCost) || throw(TypeError(
         StackTraces.stacktrace()[2].func, context, MarketBidCost, cost))
 end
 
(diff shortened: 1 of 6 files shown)


src/models/cost_function_timeseries.jl
```diff
@@ -1,3 +1,5 @@
+const OfferCurveCost = Union{MarketBidCost, ImportExportCost}
```
jd-lara: Do we want to make this into an abstract type?

GabrielKS: I'd support that

luke-kiernan: I'd vote for `OfferCostCurve` over `OfferCurveCost`, for consistency with the existing `CostCurve` type. edit: well, unless you named it specifically to make it clear it isn't a `CostCurve` subtype. But in that case, I'd choose different words, not flip the order.

jd-lara: I support Luke's semantic suggestion

PR #1568, opened 2025-11-04
changes to use some additional abstractions

jd-lara:


src/models/cost_functions/operational_cost.jl
```diff
@@ -7,8 +7,6 @@ Current concrete types include:
 - [`RenewableGenerationCost`](@ref)
 - [`StorageCost`](@ref)
 - [`LoadCost`](@ref)
-- [`MarketBidCost`](@ref)
-- [`ImportExportCost`](@ref)
```
GabrielKS, line 11: Why are we deleting these from this list? They're still current concrete types

PR #1581, opened 2025-11-12
Fix CSV parser to handle numeric and thermal multistart startup values

taranraj123:
Fixes several issues in the CSV-based system parser that prevented proper parsing of generator data with quadratic heat rate coefficients.

### Changes

- Fixed `parse_maybe_nothing` helper function to handle both string and numeric types from CSV parsers (appears in 5 locations across cost parsing functions)
- Multi-start cost fixes: simplified `make_thermal_generator_multistart` to derive hot/warm/cold startup costs from base thermal generator costs when not explicitly specified
- Added test cases for numeric CSV value handling

### Testing

All 10,062 tests pass. Added new test cases in test_power_system_table_data.jl to verify correct handling of numeric CSV values (Float64 and Int64) in addition to string values.

### DIFF
--- a/src/parsers/power_system_table_data.jl
+++ b/src/parsers/power_system_table_data.jl
@@ -897,7 +897,7 @@ function make_generator(
     gen,
     cost_colnames,
     bus,
-    gen_storage::Dict{String, Any},
+    gen_storage,
 )
     generator = nothing
     gen_type =


taranraj123:
Hi reviewers, I'm unable to sign the CLA in the Contributions guide. The link points to a Page not found error.

hannahchubin (commented):
This looks good to me. I got most of what's going on, but just clarifying this is what you're trying to do in `power_system_table_data.jl`: it looks like you're making it so that generator storage (line 900) is no longer a dictionary, meaning it can also be a single value; and in `make_cost` (starting line 940), implementing `parse_maybe_nothing()` in a better way, meaning it now checks if x is a number before parsing it as a float, instead of going from x isn't zero or nothing, then parsing x as a float.

taranraj123:
@jd-lara @hannahchubin I notice that the CrossPackageTests that are failing are unrelated to this patch. I am planning a path to PowerSimulations.jl which should address the missing package error. However I see that the failure in PowerFlows.jl is happening on main as well.

jd-lara (approved):
LGMT thanks for the contribution

Issue #1575, opened 2025-11-05
Disallow zero base power


luke-kiernan:
We should disallow 0.0 as a base power for components. Otherwise, you can get thoroughly confusing things like PSB issue [#174](https://github.com/NREL-Sienna/PowerSystemCaseBuilder.jl/issues/174): `get_reactive_power(comp)` yields `0.0`, but calling `set_reactive_power!(comp, 0.0)` then changes the reactive power to `NaN`.

luke-kiernan:
I didn't write any tests for my fix...I should've. The following still fails.
```julia
@testset "Test nonzero base power" begin
    sys = build_system(PSISystems, "RTS_GMLC_DA_sys")
    for comp in get_components(PSY.SynchronousCondenser, sys)
        @test get_base_power(comp) != 0.0
    end
end
```
Will work on an actual fix.

PR #1587, opened 2025-11-15
fix zero base power and add tests

luke-kiernan:
Fix issue #1575. 2 things I changed:

1. I was using`1e-6` as "slightly bigger than 0" for base power validation ranges in `power_system_structs.json`. However, the bounds tolerance is also `1e-6`: I switched to using `1e-4`.
2. Add a line to `check(sys::System)` that goes through all static injection components and changes any base powers of 0.0 to match that of the system.

One small caveat: we correct to nonzero base power when de-serializing (assuming `runchecks = true`), but we only warn when adding a new component with zero base power. So in this specific edge case, a serialize-and-deserialize roundtrip changes the system. One test is failing (line 16 of `test_dynamics_source.jl`) due to this inconsistency.

I'd prefer to correct the base power upon adding a new component...but I don't see precedent for that in the code. [It's very much "warn/error on bad input; don't sanitize."] Alternatively, we could throw a hard error.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -2180,21 +2180,7 @@ function check(sys::System)
     critical_components_check(sys)
     adequacy_check(sys)
     check_subsystems(sys)
-    check_base_powers(sys)
     return
-end
-
-"""
-Check that base power on static injection components are nonzero.
-"""
-function check_base_powers(sys::System)
-    for component in get_components(StaticInjection, sys)
-        if abs(get_base_power(component)) < eps()
-            @warn "Base power for $(summary(component)) is zero. Changing device base " *
-                  "power to system base power." maxlog = PS_MAX_LOG
-            set_base_power!(component, get_base_power(sys))
-        end
-    end
 end
 
 """
--- a/src/parsers/power_system_table_data.jl
+++ b/src/parsers/power_system_table_data.jl
@@ -934,6 +934,12 @@ function make_generator(
         @error "Skipping unsupported generator" gen.name gen_type
     end
 
+    if abs(get_base_power(generator)) <= 1e-6
+        @warn "Generator $(summary(generator)) has base power of zero: changing device " *
+              "base power to match system base power, $(data.base_power)"
+        set_base_power!(generator, data.base_power)
+    end
+
     return generator, reservoirs
 end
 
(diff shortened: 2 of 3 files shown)


src/base.jl
```diff
@@ -2180,9 +2180,23 @@ function check(sys::System)
...
     adequacy_check(sys)
     check_subsystems(sys)
+    check_base_powers(sys)
     return
 end
 
+"""
+Check that base power on static injection components are nonzero.
+"""
+function check_base_powers(sys::System)
```
jd-lara: We don't want to modify the systems in runchecks. As the name implies this is meant to do the checks. You can keep the warning if the base power it below 1.0 that is looks like a bad data input. But don't modify the system

jd-lara (changes requested):
Make fixes for 0 base power in the parsers

src/parsers/power_system_table_data.jl
```diff
@@ -934,6 +934,12 @@ function make_generator(
         @error "Skipping unsupported generator" gen.name gen_type
     end
 
+    if abs(get_base_power(generator)) <= 1e-6
```
luke-kiernan, line 937: For reviewers: here's where the correction happens.

(`1e-6` here is kind of arbitrary. I use `1e-4` as "almost zero" elsewhere...)

test/test_base_power.jl
```diff
@@ -0,0 +1,50 @@
...
+end
+
+@testset "Test adding component with zero base power" begin
+    sys = build_system(PSISystems, "RTS_GMLC_DA_sys")
+    bus = first(get_components(PSY.Bus, sys))
+    gen = thermal_with_base_power(bus, "Test Gen with Zero Base Power", 0.0)
+    @test_logs (:warn, "Invalid range") match_mode = :any add_component!(sys, gen)
+    gen2 = thermal_with_base_power(bus, "Test Gen with Non-Zero Base Power", 100.0)
+    @test_nowarn add_component!(sys, gen2)
+    # uncomment if we correct to non-zero base power.
```
jd-lara, line 42: @luke-kiernan what is this about? correct to non-zero base power where?

luke-kiernan: The commented-out code block is the original bug, the behavior we're trying to avoid. In this PR, I've changed things such that:
- if the user creates a system from table data, then we correct zero base power to the base power of the system
- if the user adds a component to an existing system, then we print a warning for zero base power, but still add the component as normal

So the behavior isn't fully avoided: the user can still add a component with zero base power and have weird things happen.

jd-lara (approved):

PR #1592, opened 2025-11-20
Speed up JSON deserialization with parallelization

jd-lara:
This commit implements parallelization for JSON deserialization to significantly improve performance when loading large system files.

Key changes:
- Parallelize component deserialization within each component type using Julia's Threads.@threads macro
- Use a two-phase approach:
  1. Deserialize components in parallel (CPU-intensive)
  2. Add components to system sequentially (maintains thread-safety)
- Only parallelize when there are >10 components and multiple threads available
- Preserve sequential processing for small batches to avoid overhead
- Maintain thread-safe access to component_cache (reads are lock-free, writes are sequential)

Performance benefits:
- Deserialization is CPU-bound and benefits from parallelization
- No changes to serialization format or API
- Backward compatible - falls back to sequential processing when appropriate
- Maintains referential integrity through ordered type-by-type processing

To use parallelization, run Julia with multiple threads:
  julia --threads=auto
  julia --threads=8

Thread-safety considerations:
- Component deserialization reads from component_cache (thread-safe in Julia)
- Component addition and cache updates are sequential (maintains consistency)
- Preserves the original type-based ordering required for referential integrity

Check the [contributor guidelines](https://nrel-sienna.github.io/PowerSystems.jl/stable/api/developer_guidelines/)

3. If you are contributing a new struct: please refer to the [testing requirements for new structs](https://nrel-sienna.github.io/PowerSystems.jl/stable/how_to/add_new_types/#Testing-the-addition-of-new-struct-to-the-code-base)

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -2427,6 +2427,22 @@ function deserialize_components!(sys::System, raw)
         return any(x -> type <: x, types)
     end
 
+    # Helper function to deserialize a single component from its dictionary representation
+    function deserialize_component(component_dict, component_type, cache)
+        handle_deserialization_special_cases!(component_dict, component_type)
+        return deserialize(component_type, component_dict, cache)
+    end
+
+    # Helper function to add a deserialized component to the system and update cache
+    function add_and_cache_component!(comp, post_add_func)
+        add_component!(sys, comp)
+        component_cache[IS.get_uuid(comp)] = comp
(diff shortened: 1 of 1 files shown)


jd-lara:
after testing this code we have realized that the main cost is precompilation as stated in #1593 so I'll close because that's the actual improvement we can make.

Issue #1583, opened 2025-11-13
Update PrettyTables dependencies


kennedy-mindermann:
Update PrettyTables dependencies across packages: 
- [ ] PowerSimulations
- [ ] PowerNetworkMatrices
- [ ] SiennaPRASInterface
- [ ] HydroPowerSimulations
- [ ] PowerFlows
- [ ] InfrastructureSystems
- [ ] PowerSystemCaseBuilder
- [ ] PowerSystems

PR #1594, opened 2025-11-20
updated PrettyTables and ran all tests

hannahchubin:
I'm getting no errors or breaks related to pretty tables, but I am getting two errors that seem related to PR 1581 that was already merged last week. This PR addresses [this issue](https://github.com/NREL-Sienna/PowerSystems.jl/issues/1583).

I think when all the repos have PrettyTables.jl updated is when we should merge all its PRs. Because some of the packages depend on one another I'm worried something might break if one repo has the newest PrettyTables.jl and another doesn't.


jd-lara:
closed in favor of #1595

PR #1596, opened 2025-11-24
add supplemental accesor for time series multiplier for HydroReservoir

rodrigomha:
This method is convenient for bunch of systems that have the scaling_multiplier using this function, such as RTS and some test systems

### DIFF
--- a/src/models/supplemental_accessors.jl
+++ b/src/models/supplemental_accessors.jl
@@ -226,9 +226,9 @@ function get_max_active_power(d::Union{InterruptibleStandardLoad, StandardLoad})
 end
 
 """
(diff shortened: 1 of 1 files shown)


src/models/supplemental_accessors.jl
```diff
@@ -225,6 +225,13 @@ function get_max_active_power(d::Union{InterruptibleStandardLoad, StandardLoad})
     return total_load
 end
 
+"""
+Get the maximum storage capacity for HydroReservoir. Convenient getter for `get_storage_level_limits(reservoir).max` that matches [`EnergyReservoirStorage`](@ref)
+"""
+function get_storage_capacity(reservoir::HydroReservoir)
```
jd-lara: This should have the appropriate name for the field is accessing.

jd-lara (approved):

Issue #1590, opened 2025-11-19
`with_units_base` bug


luke-kiernan:
There's `with_units_base` for system, and for components. I've discovered that one followed by the other yields strange results.

```julia
using PowerSystems, PowerSystemCaseBuilder
const PSY = PowerSystems
line = first(get_components(PSY.Line, sys))
set_active_power_flow!(line, 0.3)

# this is fine
P_natural_manual = 0.3 * get_base_power(sys)
with_units_base(sys, PSY.UnitSystem.NATURAL_UNITS) do
    @assert isapprox(P_natural_manual, get_active_power_flow(line))
end

# this is also fine
set_units_base_system!(sys, "NATURAL_UNITS")
P_natural_manual2 = get_active_power_flow(line)
set_units_base_system!(sys, "SYSTEM_BASE")
with_units_base(sys, PSY.UnitSystem.NATURAL_UNITS) do
    @assert isapprox(P_natural_manual2, get_active_power_flow(line))
end

# this isn't
P_natural = with_units_base(line, "NATURAL_UNITS") do
    PSY.get_active_power_flow(line)
end
with_units_base(sys, PSY.UnitSystem.NATURAL_UNITS) do
    @assert isapprox(P_natural, get_active_power_flow(line))
end

# if you instead specify the units on the component, then it's fine again
with_units_base(line, "NATURAL_UNITS") do
   @assert isapprox(P_natural, get_active_power_flow(line))
end
```

PR #1604, opened 2025-12-31
`with_units_base` on comp vs. on sys bugfix

luke-kiernan:
Fix issue #1590. The bug comes from the fact that `_set_units_base!(c, units)` creates a new object, breaking the link between the component's units and the system's units (shared by reference). Instead, save and restore the reference.

While this fixes the bug, this setup probably merits more careful thought and investigation. e.g. if I change the system's units in the middle of a `with_units_base(component, "NATURAL_UNITS")`, will things still restore correctly?

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,4 +65,4 @@ code contribution guidelines.
 
 PowerSystems is released under a BSD [license](https://github.com/NREL/PowerSystems.jl/blob/main/LICENSE).
 PowerSystems has been developed as part of the Scalable Integrated Infrastructure Planning (SIIP)
-initiative at the U.S. Department of Energy's National Renewable Energy Laboratory ([NREL](https://www.nrel.gov/)) Software Record SWR-23-105.
+initiative at the U.S. Department of Energy's National Laboratory of the Rockies ([NLR](https://www.nlr.gov/), formerly NREL) Software Record SWR-23-105.
--- a/docs/src/api/citation.md
+++ b/docs/src/api/citation.md
@@ -17,6 +17,6 @@ author = {José Daniel Lara and Clayton Barrows and Daniel Thom and Dheepak Kris
 keywords = {Power Systems, Julia, Energy},
 ```
 
-PowerSystems has been developed as part of the [Sienna platform](https://www.nrel.gov/analysis/sienna.html)
-by the U.S. Department of Energy's National Renewable Energy Laboratory
-([NREL](https://www.nrel.gov/)).
+PowerSystems has been developed as part of the [Sienna platform](https://www.nlr.gov/analysis/sienna.html)
+by the U.S. Department of Energy's National Laboratory of the Rockies
+([NLR](https://www.nlr.gov/), formerly NREL).
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -12,8 +12,8 @@ CurrentModule = PowerSystems
 
 ## About
 
-`PowerSystems.jl` is part of the National Renewable Energy Laboratory's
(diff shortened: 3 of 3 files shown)


luke-kiernan:
More edge cases:

1. Remove `comp` from the system mid-`with_units_base(comp, ...)`. Unattached components get their units set to `nothing`, but then when the `with` ends, it gets restored to `sys.units_settings`.
2. Attach `comp` to a system mid-`with_units_base(comp, ...)`. Errors: `units_info` is `nothing` for an unattached component, and `set_units_base!` tries to access `units_info.base_value`.

I've patched 1: don't restore the units if `comp.internal.units_info` is `nothing`.

For 2: what should [`_set_units_base!`](https://github.com/NREL-Sienna/PowerSystems.jl/blob/main/src/base.jl#L626) do if `units_info` is `nothing`? Simply return? Error/warn?

edit: I suppose there's a 3rd case, where we move from one system to another.

GabrielKS (approved):
I support merging the PR as written because it does fix a bug. Then I support creating a new issue with the further complexities @luke-kiernan has described (in particular, the surprising behavior of system `with_units_base` nested within component `with_units_base` is pretty concerning to me and I don't immediately see an elegant way to fix it). I agree with @jd-lara that it might not be worth spending that much more effort on the current version of `with_units_base` until we have more clarity on how my units redesign fits in.

daniel-thom:
> I support merging the PR as written because it does fix a bug. Then I support creating a new issue with the further complexities @luke-kiernan has described (in particular, the surprising behavior of system `with_units_base` nested within component `with_units_base` is pretty concerning to me and I don't immediately see an elegant way to fix it). I agree with @jd-lara that it might not be worth spending that much more effort on the current version of `with_units_base` until we have more clarity on how my units redesign fits in.

In an offline discussion I gave Luke the opinion that triggering these edge cases should throw an error saying that it is an unsupported operation. Unless there is a clear use case I'm not thinking of, I don't think we need to support those edge cases - especially removing a component inside this code block.

luke-kiernan:
> I support merging the PR as written because it does fix a bug. Then I support creating a new issue with the further complexities

> In an offline discussion I gave Luke the opinion that triggering these edge cases should throw an error saying that it is an unsupported operation. Unless there is a clear use case I'm not thinking of, I don't think we need to support those edge cases - especially removing a component inside this code block.

Agreed that the nested context managers is concerning and should be supported; the edge case where the component's units got changed mid-block is less concerning, and probably shouldn't be supported.

edit: there must be pretty few places where `with_units_base` is used on components, else this bug probably would've been caught earlier.

luke-kiernan:
> I agree we should throw errors for the edge cases, we should implement that in this branch.

Done: key line here is
```julia
internal.units_info === temp_units_info || throw(ErrorException(
            "Units info was modified during with_units_base."))
```

src/base.jl
```diff
@@ -644,12 +644,19 @@ end
...
+    temp_units_info = internal.units_info  # The temporary object we just created
     try
         f()
     finally
-        _set_units_base!(c, old_unit_system)
+        # Only restore if units_info is still temp_units_info.
+        # The user may have changed it in the function body, by e.g. removing the component
+        # and then attaching it to a different system.
+        internal.units_info === temp_units_info || throw(ErrorException(
+            "Units info was modified during with_units_base."))
```
daniel-thom: nit
```suggestion
        internal.units_info === temp_units_info || error(
            "Units info was modified during with_units_base.")
```

daniel-thom (approved):

PR #1629, opened 2026-02-05
fix: remove r,x and tap formatters from exporter metadata

mcllerena:
We are reverting the storing of r, x and tap parameters to leave only control_objective information in the metadata. The impedance values of transformers are being extracted in the exporter logic.


jd-lara (approved):
@hannahchubin we need to make this change un PowerFlowDataParsers.jl too

PR #1634, opened 2026-02-18
clarify wind profile in time series tutorial

yasirroni:
Fix https://github.com/NREL-Sienna/PowerSystems.jl/issues/1633

But, I'm not actually sure about the default value as I can't find related docs for that, thus only relying on trial and error. See the issue.


kdayday:
@yasirroni This tutorial is about PowerSystems input data only, not PowerSimulations data usage and outputs. It is correct as is (and copilot confirms that). Separately, I will open an issue in PowerSimulations about what appears to be a bug.

yasirroni:
Issue tracked in https://github.com/NREL-Sienna/PowerSimulations.jl/issues/1557

PR #1646, opened 2026-03-27
Fix type inference in unit conversion hot path

luke-kiernan:
## Summary
- Add type assertion `units_info::IS.SystemUnitsSettings` in `get_system_base_power` to work around abstract `UnitsData` field type
- Add `@inline` on `get_system_base_power` and `_get_multiplier`
- Add `isnothing` guard with clear error message

The `units_info` field on `InfrastructureSystemsInternal` is `Union{Nothing, UnitsData}` where `UnitsData` is abstract. After an `isnothing` check, the compiler still can't infer the concrete type, causing `jl_f_getfield` dynamic dispatch on every getter/setter call.

See also NREL-Sienna/InfrastructureSystems.jl#560 which fixes the root cause.

## Test plan
- [ ] Existing tests pass (no behavioral change)


jd-lara:
Holy shit this is massive

luke-kiernan:
shouldn't be needed any more now that the above linked IS PR is merged.

Issue #1128, opened 2024-06-20
Discrepancy between pretty print and getter functions of unattached components


kdayday:
- It is unclear if the expected behavior of the `get_` functions is to return in natural units or device or system base. We've just added natural units (MW) to many of the docstrings
- There is a discrepancy between the component print out and the getter functions of unattached components
- In the example below, we are trying to define a 30 MW generator:
    -  `rating` is defined in Device Base as 1.0 with a `base_power` of 30
    -  The print statement reports the `rating` in Natural Units as 30.0
    - The getter function returns 1.0 (on device base)
    - After attaching the generator to the system, the print statement reports `rating` on the system base as 0.3
    - The getter function returns 0.3 (on system base)

A different example is attached -- for `Line`, which doesn't have a `device_base`, it's unclear how to define the `rating`. Inputting it as natural units (200 MW) appears to get multiplied by System Base. 

```
julia> gas2 = ThermalStandard(
                      "gas_in_db",
                      true,
                      true,
                      bus2,
                      0.0,
                      0.0,
                      1.0,
                      (min=0.30, max=1.0),
                      nothing,
                      (up=.10, down=.10),
                      ThermalGenerationCost(nothing),
                      30.0,
                      (up=0.3, down=0.3),
                      false,
                      PrimeMovers.CC,
                      ThermalFuels.NATURAL_GAS)

┌ Warning: SystemUnitSetting not defined, using NATURAL_UNITS for displaying device specification.
└ @ PowerSystems ~/Documents/repos/Public_Github/PowerSystems.jl/src/utils/print.jl:192
ThermalStandard: gas_in_db:
   name: gas_in_db
   available: true
   status: true
   bus: ACBus: bus2
   active_power: 0.0
   reactive_power: 0.0
   rating: 30.0
   active_power_limits: (min = 9.0, max = 30.0)
   reactive_power_limits: nothing
   ramp_limits: (up = 3.0, down = 3.0)
   operation_cost: ThermalGenerationCost composed of variable: CostCurve{LinearCurve}
   base_power: 30.0
   time_limits: (up = 0.3, down = 0.3)
   must_run: false
   prime_mover_type: PrimeMovers.CC = 4
   fuel: ThermalFuels.NATURAL_GAS = 7
   services: 0-element Vector{Service}
   time_at_status: 10000.0
   dynamic_injector: nothing
   ext: Dict{String, Any}()
[...]

Issue #308, opened 2019-08-24
Add .editorconfig file for consistent styles


kdheepak:
https://editorconfig.org/

Adding a .editorconfig file to the root of the repository and installing an extension will ensure we use the same whitespace style across a project. What do others think about this?

PR #1659, opened 2026-04-17
Unitful getters and setters

luke-kiernan:
Add unitful getters and setters. Interface:
```julia
get_active_power(gen) # errors: must specify units
get_active_power(gen, {NU/SU/DU}) # natural units/system base/device base, with units attached 
get_active_power(gen, Float64) # system base, no units attached: internal usage.

5 + get_active_power(gen, NU) # errors: unitless combined with unitful
5 MW + get_active_power(gen, NU) # ok
0.5 SU + get_active_power(gen, NU) # errors: relative combined with natural

set_active_power!(gen, 0.5) # errors: must specify units
set_active_power!(gen, 0.5 {SU/DU}) # system base/device base
set_active_power!(gen, 50 MW) # natural units
```

Relies on IS [#574](https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/574). This will break stuff downstream left and right, yes, but here's why I think it's worth it:

1. Usability. It is very very easy to unknowingly mix units. The classic one: read in a time series from a CSV in natural units, then attach it to a system in `SYSTEM_BASE`.
2. Type stability. Our stateful unit system is inherently type unstable.
3. Maintenance. After the catch-all "not having adequate test cases," I've found more bugs due to units issues than anything else.

edit: at the moment, turns out 1-argument getters/setters assume system base. But I'd prefer to have them error: better that than get silently incorrect results. Also, I'd consider adding a 3-arg setter, with the units separate: `set_active_power!(gen, 1, {DU/SU/NU}`

edit: You can still mess up by adding or comparing device-base quantities from components with different base powers. Same goes for comparing `SU` quantities between different systems.


luke-kiernan:
Here is perhaps the biggest argument for these changes:
units_performance_comparison
This is profiling "add up active powers on all thermal gens in the 10k bus system:"
```julia
function sum_su_strip(gens)
    s = 0.0
    for g in gens
        s += IS.ustrip(get_active_power(g, IS.SystemBaseUnit())) # change this line
    end
    return s
end
```

luke-kiernan:
Change `get_active_power` etc. to be non-unitful. Move unitful ones to a different name. Setters are fine as-is. 

Does mean for get-then-set you need:
`set_active_power!(gen, get_active_power_unitful(gen, SU))`

luke-kiernan:
I plan to address #1128 (and am looking for some input on that issue), but otherwise this PR is pretty much finished.

The one other detail that comes to mind: once I go make a compatible branch of PSB, I could also reset those 3 unit tests to using PSB systems. (I currently have them using tiny systems we build here so that the tests pass.)

luke-kiernan:
As I'm working my way through all the downstream packages, I'm finding a few more things here and there. I'll post the linked PRs here:

- **PSB**: [#194](https://github.com/Sienna-Platform/PowerSystemCaseBuilder.jl/pull/194) Basically done. Tests pass locally, but there's a few errors logged. Looks like they're pre-existing. Depends on [#17](https://github.com/Sienna-Platform/PowerFlowFileParser.jl/pull/17) in PowerFlowFileParsers.jl
- **PNM**: [#308](https://github.com/Sienna-Platform/PowerNetworkMatrices.jl/pull/308). Done. All tests pass locally. edit: oh, there's merge conflicts.

jd-lara:
### Code review

Found 2 issues:

1. Four flow-limit helper accessors still call the single-argument getter forms that this PR removed, so they will throw `MethodError` at runtime. `get_flow_limits` and `get_active_power_flow_limits` now only exist in their two-argument `(component, units)` form (see [`AreaInterchange.jl#L91`](https://github.com/Sienna-Platform/PowerSystems.jl/blob/93c99a5c856c8b2cb23e8d3f31d192e15c0a675d/src/models/generated/AreaInterchange.jl#L90-L92) and [`TransmissionInterface.jl#L68`](https://github.com/Sienna-Platform/PowerSystems.jl/blob/93c99a5c856c8b2cb23e8d3f31d192e15c0a675d/src/models/generated/TransmissionInterface.jl#L67-L69)), but these callers were not migrated:

https://github.com/Sienna-Platform/PowerSystems.jl/blob/93c99a5c856c8b2cb23e8d3f31d192e15c0a675d/src/models/supplemental_accessors.jl#L311-L334

2. The `with_units_base` docstring examples call the removed single-argument `get_active_power(gen)` form, so the documented usage would now error. Additionally, the new getters take `units` explicitly and no longer consult the unit state that `with_units_base` sets, so the docstring's behavior is stale.

https://github.com/Sienna-Platform/PowerSystems.jl/blob/93c99a5c856c8b2cb23e8d3f31d192e15c0a675d/src/base.jl#L543-L548

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jd-lara (approved):

PR #1664, opened 2026-04-20
fix: update parsing logics for v35 wecc case

mcllerena:
- Fix issue with switching STYPE field parsing to use existing device mapping.
- In this WECC psse case, there are branch type elements with similar IDs:

> 1234, 4321, '1 '
> 1232, 4321, ' 1'

messing with the parsing logic. The fix adds a function counter to address this particular cases with _1, _2, etc.. suffixes.


jd-lara (commented):
@mcllerena can you open a PR with this fix in PowerFlowDataParser.jl?

PR #1666, opened 2026-04-22
Jd/drop json3

jd-lara:
Companion to https://github.com/Sienna-Platform/InfrastructureSystems.jl/pull/571


jd-lara:
the tests will still fail until TableDataParser is merged onto the registry

PR #1676, opened 2026-05-06
parse tap transformers from psse

m-bossart:
Parse all transformers from PSSE as tap transformer by default, to avoid parsing mixed transformer types in parallel.


m-bossart:
The downstream cross package test failures in PSI are due to network formulations that require all device types to be formulated.

jd-lara:
Closing this to implement a different logic as discussed in our meeting

PR #1693, opened 2026-06-09
Fix docs build: add check_parallel_branch_type_consistency to public API reference

mcllerena:

Check the [contributor guidelines](https://sienna-platform.github.io/PowerSystems.jl/stable/api/developer_guidelines/)

3. If you are contributing a new struct: please refer to the [testing requirements for new structs](https://sienna-platform.github.io/PowerSystems.jl/stable/how_to/add_new_types/#Testing-the-addition-of-new-struct-to-the-code-base)

## Description

The GitHub Actions `build` (docs) job was failing with the following error:

```
Error: 1 docstring not included in the manual:
    PowerSystems.check_parallel_branch_type_consistency :: Tuple{System}
```

`check_parallel_branch_type_consistency` is a public exported function with a docstring in `src/utils/IO/system_checks.jl`, but it was not referenced in any `@docs` or `@autodocs` block in the documentation. Documenter.jl's `:missing_docs` check treats this as a build error.

**Fix**: Added a `@docs check_parallel_branch_type_consistency` block to `docs/src/api/public.md` in the System section so the docstring is included in the public API reference.


src/parsers/power_models_data.jl
```diff
@@ -1205,9 +1205,23 @@ function get_branch_type_matpower(
 
     _add_vector_control_group(d, "shift", "group_number")
 
+    is_identity_tap = abs(tap - 1.0) <= IDENTITY_TAP_TOL || tap == 0.0
```
jd-lara, line 1208:
```suggestion
    is_identity_tap = abs(tap - 1.0) <= IDENTITY_TAP_TOL || iszero(tap)
```

src/definitions.jl
```diff
@@ -578,6 +578,15 @@ const PSSE_PARSER_TAP_RATIO_LBOUND = 0.5
 const PARSER_TAP_RATIO_CORRECTION_TOL = 1e-5
 
 const ZERO_IMPEDANCE_REACTANCE_THRESHOLD = 1e-4
+const ZERO_IMPEDANCE_RESISTANCE_THRESHOLD = 1e-6
```
m-bossart: How do we set this value? My impression from PSSE is that a line is only considered zero impedance if the resistance is zero and the reactance is below the user defined threshold (with default 1e-4). Just wondering what the logic is for having it be non-zero. Does this change significantly the number of parsed zero-impedance branches in the EI?

src/parsers/power_models_data.jl
```diff
@@ -1236,10 +1251,24 @@ function get_branch_type_psse(
...
     _add_vector_control_group(d, "shift", "group_number")
     is_tap_controllable, is_alpha_controllable = _determine_control_modes(d, "COD1", "tap")
+
+    is_identity_tap = abs(tap - 1.0) <= IDENTITY_TAP_TOL || tap == 0.0
+    is_zero_shift = abs(shift) <= ZERO_ANGLE_SHIFT_TOL
+
     if d["group_number"] == WindingGroupNumber.UNDEFINED || is_alpha_controllable
+        # Degenerate PST: unrecognized vector group with near-zero shift → demote
+        if is_zero_shift && !is_alpha_controllable
+            return is_identity_tap ? Transformer2W : TapTransformer
```
m-bossart: Shouldn't we also be checking for `is_tap_controllable` as well on this path when deciding between `Transformer2W` and `TapTransformer` here? Similar to line 1267?

src/parsers/power_models_data.jl
```diff
@@ -1248,15 +1277,133 @@ function get_branch_type_psse(
...
+    return abs(d["br_r"]) < ZERO_IMPEDANCE_RESISTANCE_THRESHOLD &&
+           abs(d["br_x"]) < ZERO_IMPEDANCE_REACTANCE_THRESHOLD
+end
+
+function _is_active_pti_branch(d::Dict)
+    return get(d, "br_status", 0) == 1
+end
+
+function _expected_discrete_state(d::Dict, bus_f::ACBus, bus_t::ACBus)
+    available = d["br_status"] == 1
```
m-bossart: use `_is_active_pti_branch`

src/utils/IO/system_checks.jl
```diff
@@ -127,3 +127,52 @@ function total_capacity_rating(sys::System)
...
+PSY types (e.g. a `Transformer2W` in parallel with a `TapTransformer`). Such mixed-type parallel
+groups can introduce issues in the network reduction.
+
+# Example
+```julia
+n = check_parallel_branch_type_consistency(sys)
+n == 0 || @warn "System has \$n arcs with mixed-type parallel branches"
+```
+"""
+function check_parallel_branch_type_consistency(sys::System)
```
m-bossart, line 144: I wonder if this function should distinguish between cases where both branches are available (will actually cause an issue in the network reduction/modeling) and cases where you have parallel mixed types but one branch is unavailable and therefore has no downstream impact assuming it stays unavailable.

src/parsers/power_models_data.jl
```diff
@@ -1248,15 +1277,133 @@ function get_branch_type_psse(
...
+               all(t -> t in (Line, DiscreteControlledACBranch), branch_types)
+            all_lines_near_zero =
+                get(has_line_by_arc, arc_key, false) &&
+                get(line_near_zero_by_arc, arc_key, false)
+            if all_lines_near_zero
+                overrides[arc_key] = DiscreteControlledACBranch
+                @warn "Normalizing mixed parallel Line/DiscreteControlledACBranch on near-zero-impedance arc $(arc_key[1])-$(arc_key[2]) to DiscreteControlledACBranch." _group =
+                    IS.LOG_GROUP_PARSING maxlog = PS_MAX_LOG
+            else
+                @warn "Keeping mixed parallel Line/DiscreteControlledACBranch on arc $(arc_key[1])-$(arc_key[2]) because at least one Line has non-negligible impedance." _group =
```
m-bossart: Couldn't we theoretically get to this last else with some other combination of branch types in the case of bad data? Might make sense to at lease include the branch types in the log message instead of just "Line/DiscreteControlledACBranch".

src/parsers/power_models_data.jl
m-bossart: This path when the winding group number is undefined I don't think is explicitly tested.

mcllerena:
@copilot resolve the merge conflicts in this pull request

src/definitions.jl
mcllerena: Agreed, I'm reverting this since just a few of them are getting converted, and following PSSE directives, only when r=0 and x<thresh, the lines is ZIL.

jd-lara (approved):

PR #1695, opened 2026-06-10
Claude/inspiring johnson pqzkyz

jd-lara:
Relies on the companion IS branch https://github.com/Sienna-Platform/InfrastructureSystems.jl/pull/585


luke-kiernan (approved):
I like the simplifications. I did some pretty thorough performance optimization stuff on the old version, though. So requests:
1. Verify that the internal conversions get compiled away: `code_lowered` on `get_active_power(g, {SU/DU/NU})` should show `getfield` plus straight float arithmetic.
2. Verify that `ustrip` is a no-op when types are inferred: `ustrip(get_active_power_unitful(g, {SU/DU/NU}))` compiles to the same thing as the non-unitful getter.
3. Performance check: should get similar results to [here](https://github.com/Sienna-Platform/PowerSystems.jl/pull/1659#issuecomment-4291284878)
4. Type stability audit.

PR #1702, opened 2026-06-22
fix: use actual R value when switch/breaker symbol is present in v35 BRANCH entry

### LINKED ISSUE #1701: PSS(e) parser: breakers in `BRANCH` section with nonzero r
When parsing a v35 raw BRANCH entry whose CKT starts with `@` (breaker), PSS(e) retains the r value. Meanwhile, we're unconditionally setting `r = 0.0` [here](https://github.com/Sienna-Platform/PowerSystems.jl/blob/main/src/parsers/pm_io/psse.jl#L1980) for v35, and using the parsed r for v33. Should we use the parsed r for v35 too?

The downstream effect is that PNM's ybus entries differ significantly from PSS(e)'s (abs error `~5e2`) for certain `DiscreteControlledACBranch` elements in the EI. Matching PSS(e)'s behavior here would knock the biggest discrepancies between the y bus matrices all the way down to `~3`, two orders of magnitude better.

If we switch to using the input file's `r` value, we may want to revise the ZIL handling in PNM to check if `r != 0.0` before merging two buses connected by a closed `DiscreteControlledACBranch`.

Probably also applies to BRANCH entries whose CKT starts with `*` (switch), but would need to verify by asking Marck to export PSS(e)'s y bus for a RAW input with such an entry.

mcllerena:


luke-kiernan:
Hmm the cross-package test that's failing is PNM's `test_ybus_reductions.jl:966`: `Expression: all(isfinite, (ABA_Matrix(yb; factorize = false)).data.nzval)`. I will look into it.

luke-kiernan:
Claude diagnosed the issue. The failure is a version control artifact. I'll re-run and it should pass.

> The failing test ("ZIBR fast merge: a cancelled...") was added in de266644 (2026-06-20), but the fix that makes it pass (35edc0d9, the BA_Matrix guards) wasn't committed until 2026-06-22 20:13 UTC. The PR #1702 cross-package CI ran at 2026-06-22 17:03–17:21 UTC — after the test landed but ~3 hours before the fix. So CI cloned a PNM main that had the failing test but not its fix.

luke-kiernan (approved):

PR #1707, opened 2026-06-26
PSS/E parser: apply `base_conversion` to `ramp_limits`

### LINKED ISSUE #1706: PSS/E parser: base_conversion not applied to (specifically) ramp_limits
Discovered while testing WECC .jl 

In src/parsers/power_models_data.jl, make_hydro_dispatch, make_hydro_reservoir, and make_thermal_gen scale active_power_limits from the system per-unit base into the device base via base_conversion = sys_mbase / mbase, but do not apply base_conversion to ramp_limits. Leaving them un-converted makes the stored ramp too large by 1 / base_conversion = MBASE / 100.

Ramp limits should be in the device base. The no-data fallback should give ramp == Pmax (ratio 1.0), matching active_power_limits.

PabloBotin:
`make_hydro_dispatch`, `make_hydro_reservoir`, and `make_thermal_gen` in
`src/parsers/power_models_data.jl` scale `pmin` / `pmax` / `pg` / `qg` /
`rating` from the PowerModels system per-unit base into the device per-unit
base via `base_conversion = sys_mbase / mbase`, but the `ramp_limits` returned
by `calculate_ramp_limit` (`src/parsers/common.jl`) were left in the system
base. For any generator with `MBASE ≠ 100` this stored a ramp too large by
`MBASE / 100` — e.g. a unit with `MBASE = 448.5` hitting the no-ramp-data
`pmax` fallback ended up with `ramp / Pmax ≈ 4.485`.

### Fix

Added an optional positional argument `base_conversion::Float64 = 1.0` to
`calculate_ramp_limit` and multiplied each returned `up` / `down`, preserving
the `nothing` branch. The three call sites in `power_models_data.jl` now pass
the already-computed `base_conversion`. This matches the existing pattern in
the sibling helper `calculate_gen_rating` in the same file.

The default value keeps `calculate_ramp_limit` backwards-compatible for any
hypothetical external caller and for systems with `MBASE = 100`. The helper is
unexported and has no callers outside the three sites in
`power_models_data.jl`.

### Tests

`test/test_parse_ramp_limits.jl` adds two testsets:

- **Helper-level unit tests** for `calculate_ramp_limit` covering every input
  branch and the new argument: `ramp_agc` precedence, `ramp_10`, `ramp_30`,
  `pmax` fallback (including negative `pmax`), `nothing` branch, and the
  default `base_conversion = 1.0`.
- **End-to-end parser tests** that prove each call site forwards
  `base_conversion` at `MBASE = 250`:
  - `make_thermal_gen` `pmax` fallback — exact symptom from the issue
    (`ramp_limits.up == max_active_power`).
  - `make_hydro_dispatch` `pmax` fallback via `(fuel=HYDRO, type=ROR)`
    routing. `make_hydro_reservoir` uses identical scaffolding around the
    same `calculate_ramp_limit` invocation.

Full test suite (11753 / 11753) passes.

Closes #1706.


jd-lara (changes requested):
I don't think that passing a default of 1.0 for the base conversion here is the cleanest solution. This PR is essentially a no-op. Can you clarify the design and intent?

PabloBotin:
@jd-lara 

I agree that passing a default of 1.0 for the base conversion is not the cleanest solution and I removed that default. 

On the other hand, why do you say this PR is a no-op? 

I came across this issue when I was parsing the ramp limits for WECC: any generator with `MBASE ≠ 100` was getting an unrealistically large ramp.

Looking at the PSY parsed I realized `base_conversion` was applied to `pg`, `qg`, `rating` but was missing for `ramp_limits`. 

I applied the same `base_conversion` to the ramps on my PSY fork and it worked fine so I decided to send it as a PR. 

The tests that I added showcase the functionality of this fix. 

Please let me know why you think it was not a good PR overall and how would you deal with this issue.. I really need to solve it for WECC!

jd-lara (approved):

Issue #1684, opened 2026-05-19
Missing fields tap transformer


jd-lara:
The tap transformers have missing fields to model their controllability 

`tap min/max, step count, regulated bus, Vset`

jd-lara:
Note to myself, make the new transformer winding object have a mandatory rating value

PR #1714, opened 2026-07-08
transformer refactor

### LINKED ISSUE #1271: Add `WindingData` struct that contains `active_power`, `reactive_power`, `r`, `x`, and all the other fields that exist for each of the three windings so that we can just have three `WindingData` instances rather than duplicating all these fields?
Thoughts on having like a `WindingData` struct that contains `active_power`, `reactive_power`, `r`, `x`, and all the other fields that exist for each of the three windings so that we can just have three `WindingData` instances rather than duplicating all these fields?

Accepted but deferred to PSY 6.0 

_Originally posted by @GabrielKS in https://github.com/NREL-Sienna/PowerSystems.jl/pull/1252#discussion_r1927714767_

m-bossart:
Accompanying PRs:
https://github.com/Sienna-Platform/PowerSystemsTestData/pull/121
https://github.com/Sienna-Platform/PowerFlowFileParser.jl/pull/19
https://github.com/Sienna-Platform/PowerSystemCaseBuilder.jl/pull/200

closes #1271 #1684


src/descriptors/power_system_structs.json
```diff
@@ -915,40 +1030,19 @@
...
                 },
                 {
                     "null_value": "0.0",
                     "name": "r",
                     "data_type": "Float64",
-                    "comment": "Resistance in pu ([`SYSTEM_BASE`](@ref per_unit))",
+                    "comment": "Resistance in pu (device base on `base_power`), referenced to the primary (winding) base voltage",
                     "valid_range": {
-                        "min": 0,
+                        "min": -2,
```
m-bossart: Shouldn't be changed. Should remain at 0

docs/src/api/public.md
```diff
@@ -48,7 +51,7 @@ Modules = [IS]
 Pages   = ["geographic_supplemental_attribute.jl"
         ]
 Order = [:type, :function]
-Filter = t -> nameof(t) in names(PowerSystems)
+Filter = t -> !(t isa Union) && nameof(t) in names(PowerSystems)
```
m-bossart, line 62: Why is this addition necessary?

src/models/generated/TwoWindingTransformer.jl
```diff
@@ -0,0 +1,127 @@
+#=
```
m-bossart, line 1: We should remove base_power from `TwoWindingTransformer`. The base power of the winding is the base power of the transformer. This is duplicative.

src/models/generated/TransformerWinding.jl
```diff
@@ -0,0 +1,177 @@
...
+#! format: off
+
+"""
+    mutable struct TransformerWinding <: DeviceParameter
+        arc::Arc
+        tap::Float64
+        α::Float64
+        winding_group_number::WindingGroupNumber
+        control::Union{Nothing, TransformerControl}
+        available::Bool
```
m-bossart: Make available the first entry in the struct

src/models/transformer_control.jl
```diff
@@ -0,0 +1,99 @@
...
+    )
+end
+
+get_objective(c::TransformerControl) = c.objective
+get_regulated_bus_number(c::TransformerControl) = c.regulated_bus_number
+get_limits(c::TransformerControl) = c.limits
+get_controlled_quantity_limits(c::TransformerControl) = c.controlled_quantity_limits
+get_number_of_tap_positions(c::TransformerControl) = c.number_of_tap_positions
+
+# `TransformerControl` is a plain struct (not an `IS.InfrastructureSystemsType`) and has no
```
m-bossart: This comment is too verbose

src/models/generated/ThreeWindingTransformer.jl
jd-lara, line 1: Should we add validations to these constructors? Like no repeated group numbers etc? I am not sure if these should be auto generated so we can implement some additional validation logic.

src/models/generated/ThreeWindingTransformer.jl
```diff
@@ -0,0 +1,207 @@
...
+#! format: off
+
+"""
+    mutable struct ThreeWindingTransformer <: ACTransmission
+        name::String
+        primary_winding::TransformerWinding
+        secondary_winding::TransformerWinding
+        tertiary_winding::TransformerWinding
+        star_bus::ACBus
+        r_12::Float64
```
jd-lara: Aren't these parameters stored in the windings?

m-bossart: The design here is that the the TransformerWinding is for the winding to star bus data. Because PSSE gives the impedance data between windings, it doesn't fit in TransformerWinding. We handle it in PNM with a wrapper over the winding with the actual impedances that we use for modeling.

rodrigomha:
Discussion:
- Rename x_13 and r_13 and whatever_13 to whatever_31.
- Flatten TransformerControl into the Winding struct
- Add r and x to the Winding struct. Remove r and x to the TwoWindingTransformer. Keep r_ab, x_ab and base_power_ab in the ThreeWindingTransformer struct. This is repeated data but its okay.
- Need a new enum for magnetizing shunt location

src/models/generated/ThreeWindingTransformer.jl
luke-kiernan: >  Mutating either representation after construction does not update the other.

What downstream consumers need `x_12`, `r_12`, and `base_power_12`? The only one I can think of is the PSS(e) exporter. Everywhere else uses the Y model. If that's it, I'd vote for eliminating the `r_12` fields. Keeping different data fields in sync with each other is a frequent source of bugs.

src/models/supplemental_accessors.jl
```diff
@@ -321,84 +173,30 @@ end
...
+function get_max_active_power_flow_limit(tx::TransmissionInterface, units::UnitArg)
     return get_active_power_flow_limits(tx, units).max
 end
 
-"""
-Calculate the phase shift angle α for a [`TapTransformer`](@ref) or [`Transformer2W`](@ref) based on its winding group number.
-Returns the angle in radians, calculated as -(π/6) * `winding_group_number`.
-If the `winding_group_number` is `WindingGroupNumber.UNDEFINED`, returns 0.0 and issues a warning.
-"""
-function get_α(t::Union{TapTransformer, Transformer2W})
```
luke-kiernan, line 353: Add forwarding for alpha, `get_α(t::TwoWindingTransformer) = get_α(get_winding(t))`

src/models/generated/TransformerCircuit.jl
```diff
@@ -0,0 +1,248 @@
...
+        number_of_tap_positions::Int
+        rating::Union{Nothing, Float64}
+        rating_b::Union{Nothing, Float64}
+        rating_c::Union{Nothing, Float64}
+        active_power_flow::Float64
+        reactive_power_flow::Float64
+        base_power::Float64
+        base_voltage_primary::Union{Nothing, Float64}
+        base_voltage_secondary::Union{Nothing, Float64}
+        units_info::Union{Nothing, SystemUnitsSettings}
```
luke-kiernan: Ick. We're trying to move away from stateful units stuff...

What's going on here: `TransformerCircuit` needs to know the system base power for unit conversion purposes. However, `TransformerCircuit` subtypes `DeviceParameter`, _not_`Component`. So the existing plumbing in IS that goes thru `internal::InfrastructureSystemsInternal` doesn't apply. Thus we need to strap on this extra `units_info` field and manage it here in PSY (see `set_units_setting!`, `UnitsBearer`).

Could we promote `TransformerCircuit` to a `Component`? We already have nested components-- the `arc::Arc` field of lines--so it could be managed in the same way. That way it'd carry around an `internal` field and the units plumbing just works. Added benefit: `IS.serialize` and `IS.deserialize` come for free, instead of needing separate implementations.

[Another potential solution, which I like less: add a back-pointer and forward the `_get_system_base_power` call to the parent transformer.`PairBase` uses a similar strategy.]

docs/src/how_to/handle_3W_transformers.md
```diff
@@ -1,10 +1,10 @@
 # [Handle 3-winding transformer data](@id 3wtdata)
 
-PowerSystems.jl stores the topological data for the [`Transformer3W`](@ref) as the common equivalent circuit in the star (or wye) configuration. In this representation, the series impedances of each winding are transformed into an equivalent star network with a common star bus.
+PowerSystems.jl models the [`ThreeWindingTransformer`](@ref) with a star (or wye) topology: three [`TransformerCircuit`](@ref)s ([`get_primary_circuit`](@ref), [`get_secondary_circuit`](@ref), [`get_tertiary_circuit`](@ref)), each connecting a terminal bus to a common star (hidden) bus ([`get_star_bus`](@ref)). PowerSystems.jl does **not** store or compute the equivalent star-leg impedances $Z_1, Z_2, Z_3$ themselves. It stores only the measured (PSS/E-shaped) pairwise impedances `r_12`/`x_12`, `r_23`/`x_23`, `r_31`/`x_31` between winding pairs, on their respective per-pair base powers. Deriving the star-leg impedances from these pairwise values is left to downstream packages — PowerNetworkMatrices owns the cached star derivation used for network-matrix assembly.
```
luke-kiernan, line 3: > PowerSystems.jl does **not** store or compute the equivalent star-leg impedances $Z_1, Z_2, Z_3$ themselves. 

What?? Seems contradictory or stale. Maybe AI wrote this comment based on PNM's _current code_, but I don't think this accurately reflects where we're heading.

I asked Claude to trace how parsing PSS(e) .raw works. It says that `PowerFlowFileParsers.jl` does the delta-Y conversion and hands those values off to PSY. Meaning: PSY _does_ store the equivalent star-leg impedances.

src/models/components.jl
```diff
@@ -167,210 +196,126 @@ _convert_from_device_base(base, v::StartUpShutDown, cu, u) = (
...
 _unit_category(::Val{:ohm}) = IMPEDANCE
 _unit_category(::Val{:siemens}) = ADMITTANCE
+
+# Base provider for the pairwise impedance fields of a ThreeWindingTransformer.
+# Convention: Z_ij is pu on base_power_ij referenced to the first-index circuit's
+# base voltage (PSS/E CZ = 1): r_12/x_12 -> primary, r_23/x_23 -> secondary,
+# r_31/x_31 -> tertiary (R3-1 references NOMV3, the tertiary winding). The
+# transformer-level `magnetizing_shunt` is pu on base_power_12 referenced to the
+# primary circuit's base voltage.
+struct PairBase{T <: ThreeWindingTransformer}
```
luke-kiernan, line 308: This gets inserted into the middle of the units conversion machinery, which makes me wonder about performance. edit: seems fine. 0 allocations.

Also, see my other comment: do we actually need `x_12`, `r_12`, and `base_power_12` downstream? If not and we remove those fields, then this likely becomes redundant.

luke-kiernan (commented):
2 big picture things that merit discussion:

1. `TransformerCircuit`'s `units_info` field. That "internal units state" stuff is slated to get removed before IS4 and PSY6 gets released. Now, here it's just used to keep the component's knowledge of the system's base power up-to-date, not for "which units are we in." But it's still clunky and borderline type piracy: I'd prefer to use the existing units plumbing in IS, instead of strapping stuff on top.
2. Delta-Y conversions and our data model. Do we really need the delta model parameters downstream?

edit: just saw the following.
> This is repeated data but its okay.

I'm ok with repeated data, but for how little they get used, I'd throw them in the `ext` dictionary instead of making them fields.

src/models/generated/ThreeWindingTransformer.jl
m-bossart: Definitely agree that having repeated data fields that need to be in sync is not ideal. We settled on this decision because of the desire to have the data schema be acceptable to a wide range of users, some of whom will want to store data in the format from PSSE directly. That being said, we should make those fields optional.

m-bossart:
> 2 big picture things that merit discussion:
> 
> 1. `TransformerCircuit`'s `units_info` field. That "internal units state" stuff is slated to get removed before IS4 and PSY6 gets released. Now, here it's just used to keep the component's knowledge of the system's base power up-to-date, not for "which units are we in." But it's still clunky and borderline type piracy: I'd prefer to use the existing units plumbing in IS, instead of strapping stuff on top.
> 2. Delta-Y conversions and our data model. Do we really need the delta model parameters downstream?
> 
> edit: just saw the following.
> 
> > This is repeated data but its okay.
> 
> I'm ok with repeated data, but for how little they get used, I'd throw them in the `ext` dictionary instead of making them fields.

On the units: I don't think `TransformerCircuit` should be a `Component`. We should try to find an acceptable solution for the units handling that does not require this. Maybe the back-pointer is the best option. let's discuss this afternoon

docs/src/how_to/handle_3W_transformers.md
m-bossart: Yes this is stale from the first refactoring and was not properly updated.

m-bossart:
Notes from meeting:

- @luke-kiernan will look into a generic units solution for non-components. From there I will update the PRs in all downstream packages. 
- I will make the psse data fields (`r_12`, `x_12`) optional for final design.
- Two distinct enums for shunt location in two winding vs. three winding is OK.

PR #1718, opened 2026-07-09
make susceptance calculation consistent for three winding transformers.

m-bossart:
Consolidates `get_series_susceptances` for all `ThreeWindingTransformer`s. 
@luke-kiernan Do we have tests in PowerFlows.jl comparing DC power flow results with PSSE for systems with `Transformer3W`?

### DIFF
--- a/src/models/supplemental_accessors.jl
+++ b/src/models/supplemental_accessors.jl
@@ -55,14 +55,13 @@ function get_series_susceptance(::Union{PhaseShiftingTransformer3W, Transformer3
 end
 
 """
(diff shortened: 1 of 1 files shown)


luke-kiernan:
> Do we have tests in PowerFlows.jl comparing DC power flow results with PSSE for systems with Transformer3W?

No: the one test where we compare DC vs PSS(E) uses `"2area-5-bus-system"`, which doesn't have any 3WTs (or even 2WTs). Should I add one?

jd-lara (approved):

PR #1720, opened 2026-07-13
parse multi swing cases

jd-lara:
this is a first pass at parsing systems with multiple ref/swing buses. I added the raw file here but can change it to PSB later


m-bossart (commented):
These code changes look ok to me, but the tests are referencing raw files that were not added. We should either add those raw files here so the tests pass and open a new issue to transfer some of these fixtures into case builder.

m-bossart:
This should now be passing tests once the new PSCB release lands

m-bossart (approved):

PR #1733, opened 2026-07-28
Mb/dgen bug

### LINKED ISSUE #1731: Parsing distributed generation doesn't account for DGENM (distributed generation status)
See:
https://github.com/Sienna-Platform/PowerSystems.jl/blob/a2cd37f0bfdb46adb6d2431f09386feb930a57eb/src/parsers/pm_io/psse.jl#L586-L598

m-bossart:
Fixes: https://github.com/Sienna-Platform/PowerSystems.jl/issues/1731

Requires (in order) for tests to pass:
https://github.com/Sienna-Platform/PowerSystems.jl/pull/1732
https://github.com/Sienna-Platform/PowerSystemsTestData/pull/122
https://github.com/Sienna-Platform/PowerSystemCaseBuilder.jl/pull/209

### DIFF
--- a/src/parsers/pm_io/psse.jl
+++ b/src/parsers/pm_io/psse.jl
@@ -592,13 +592,13 @@ function _psse2pm_load!(pm_data::Dict, pti_data::Dict, import_all::Bool)
                 dgenm = pop!(load, "DGENM", 0.0)
             end
 
-            # PSS(R)E nets a load's distributed generation out of the demand only while it
-            # is in service. 
-            pl = pop!(load, "PL")
-            ql = pop!(load, "QL")
-            dgen_in_service = dgenm != 0
-            sub_data["pd"] = dgen_in_service ? pl - dgenp : pl
-            sub_data["qd"] = dgen_in_service ? ql - dgenq : ql
+            # The load keeps the gross demand; distributed generation (DGENP/DGENQ)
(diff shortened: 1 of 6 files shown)


src/parsers/pm_io/psse.jl
```diff
@@ -592,10 +592,13 @@ function _psse2pm_load!(pm_data::Dict, pti_data::Dict, import_all::Bool)
                 dgenm = pop!(load, "DGENM", 0.0)
             end
 
-            # PSS(R)E models distributed generation as negative demand on the load record.
-            # Net active/reactive demand seen by PF should be gross load minus DGEN.
-            sub_data["pd"] = pop!(load, "PL") - dgenp
-            sub_data["qd"] = pop!(load, "QL") - dgenq
+            # PSS(R)E nets a load's distributed generation out of the demand only while it
+            # is in service. 
+            pl = pop!(load, "PL")
```
jd-lara: Are we recovering the dgen quantity as a RenewableFix?

m-bossart: Currently no, we are only netting the load. This fix just correctly reproduces PSSE behavior. We can change the behavior to have a separate RenewableFix and use the DGENM to set the status.

jd-lara: we should do that in here so that the net injections always match and reproduce PSSe fully

jd-lara (approved):

PR #1745, opened 2026-08-06
Declare loss curve fields as LossCurve so their units are explicit

### LINKED ISSUE #1728: Loss curves and units
Currently, fields that are loss curves are declared as `ValueCurve` types. Unlike `CostCurve` types, those objects don't have units data attached to them. Hence the units are currently ambiguous. 

This is causing actual issues. `psi_library.jl:1829-1957` in PSB builds a loss curve on an `InterconnectingConverter` with the y axis in system base. Meanwhile, code in an open POM PR assumes that same curve has its y axis in device base...

Potential solutions:
1.  Eliminate the ambiguity by adding units somewhere.
    a. Add units to `ValueCurve`.
    b. Introduce some other curve type that carries units, but isn't a `CostCurve`.
    c. Just make them `CostCurve`s, even though losses aren't costs.
2. Decide on some fixed units and document that choice prominently. 

edit: here I'm talking about the `IS4`/`psy6` branches.

luke-kiernan:
Fixes #1728. Pairs with InfrastructureSystems.jl#603 — `Project.toml`/`test/Project.toml` are pinned to that branch so CI resolves; revert to `rev = "IS4"` once it merges.

`TwoTerminalGenericHVDCLine.loss`, `TwoTerminalLCCLine.loss`, `TwoTerminalVSCLine.converter_loss_{from,to}` and `InterconnectingConverter.loss_function` were bare `ValueCurve`s with no units attached. They now take `LossCurve`s, which carry an `AbstractUnitSystem` marker as a type parameter, so the units travel with the value. Permitted curve shapes are unchanged.

`LossCurve` shares the units machinery with the cost curves via the new `ValueCurveWithUnits` supertype, but is not a `ProductionVariableCostCurve`, so cost fields still reject it.

## Note on field typing

The fields go from a union of two concrete types to a union of two `UnionAll`s (open in the units parameter), so `isconcretetype` is now false. In practice this costs nothing:

- **Storage is unchanged** — neither `LinearCurve` nor `PiecewiseIncrementalCurve` is `isbitstype`, so the field was already a boxed pointer.
- **Curve-shape inference is unchanged** — old and new both infer `get_function_data` as `Union{LinearFunctionData, PiecewiseStepData}`. Each arm still pins the curve parameter, and the open units parameter carries no data.
- The one difference is `get_power_units`, which infers `Any` — un-inferable by design, since fixing it would mean fixing the units per field.

For the record: deferring the shape check to a runtime setter validation (bare `LossCurve` field) would be worse, dropping `get_function_data` inference to `Any` for no gain.

## Downstream

Callers must now say which unit system they mean. Not fixed here:

- **PSCB** — `dc_branch_csv_parser!` and `psi_library.jl` pass bare curves; this is what currently fails PSY CI (58 errors, single root cause).
- **PSI / POM** — `isa(loss, PSY.LinearCurve)` guards need to look through the wrapper. POM's `mt_hvdc_models/HVDCsystems.jl` also has the device-base assumption from #1728 written down in a comment; that can now be checked rather than assumed.

Tests: 5685 pass, 0 failed. The 58 errors are the PSCB issue above.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -17,7 +17,7 @@ Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [sources]
-InfrastructureSystems = {url = "https://github.com/Sienna-Platform/InfrastructureSystems.jl.git", rev = "IS4"}
+InfrastructureSystems = {url = "https://github.com/Sienna-Platform/InfrastructureSystems.jl.git", rev = "lk/issue-1728-loss-curve-units"}
 
 [compat]
 DataStructures = "^0.19"
--- a/docs/src/api/public.md
+++ b/docs/src/api/public.md
@@ -74,6 +74,17 @@ Order = [:type, :function]
 Filter = t -> !(t isa Union) && nameof(t) in names(PowerSystems)
 ```
 
+## Curves with Units
+
+```@autodocs
+Modules = [IS]
+Pages   = ["value_curve_with_units.jl",
+            "loss_curve.jl",
+           ]
+Order = [:type, :function]
+Filter = t -> !(t isa Union) && nameof(t) in names(PowerSystems)
+```
+
 ## Time Series
 
 ```@autodocs
--- a/src/PowerSystems.jl
+++ b/src/PowerSystems.jl
@@ -95,12 +95,13 @@ export get_points
 export get_x_coords
 export get_y_coords
 
-# from IS value_curve.jl, cost_aliases.jl, and production_variable_cost_curve.jl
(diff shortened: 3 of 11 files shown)


luke-kiernan:
> Caveat: these fields are no longer concretely typed

Cost curves are already not concrete types, so imo trying to keep the loss curves concrete isn't worth the effort. They're used much much more rarely than cost curves.

luke-kiernan:
Reverted to draft--revisit after the schema stuff lands, so that the folks working on integrating that stuff don't have to deal with a moving target.

luke-kiernan:
Closing. Needs to be redesigned: type parameter ought to apply to both the x axis and the y axis. See issue for details

PR #1746, opened 2026-08-07
schema matching and serde update

jd-lara:
This PR updates several models to match the schema and also updates the code auto generation to produce the needed code to come back and forth from PowerOpenAPIModels


src/openapi/export_document.jl
```diff
@@ -0,0 +1,649 @@
...
+# string -> function table — export does not need one, since the *type* is already known;
+# only the *emitted* `attribute_type` string (read by `add_supplemental_attribute!` on the way
+# back in) is derived from it, via `string(nameof(typeof(attr)))`, matching
+# PowerTableDataParser.jl/src/openapi/container.jl's own convention exactly.
+
+const POLLUTANTTYPE_TO_STRING = Dict(v => k for (k, v) in POLLUTANTTYPE_FROM_STRING)
+const EMISSIONBASIS_TO_STRING = Dict(v => k for (k, v) in EMISSIONBASIS_FROM_STRING)
+const MASSUNIT_TO_STRING = Dict(v => k for (k, v) in MASSUNIT_FROM_STRING)
+const ENERGYUNIT_TO_STRING = Dict(v => k for (k, v) in ENERGYUNIT_FROM_STRING)
+const COMBINEDCYCLECONFIGURATION_TO_STRING =
```
daniel-thom: Why are we using squashed case?

src/openapi/import_document.jl
```diff
@@ -0,0 +1,705 @@
...
+# ── Dependency-ordered component pass ───────────────────────────────────────────
+# Topology first, then arcs and branches, then injectors, then reserves. `TransformerCircuit` is the one
+# `addable = false` entry — a `DeviceParameter` embedded in `TwoWindingTransformer.circuit`
+# or one of `ThreeWindingTransformer.{primary,secondary,tertiary}_circuit`, never a
+# standalone System component, but still registered in `OpenAPIRefs` so the transformer
+# that references it can resolve it. `TransmissionInterface` (a `Service`, membership
+# carried by a `supplemental_attribute_associations` row like the reserves) is placed after
+# them for that reason, though it has no forward references of its own.
+const DOCUMENT_PLAN = [
+    (PO.Area, Area, "Area", true),
```
daniel-thom: What is the 4th parameter?

src/openapi/import_document.jl
```diff
@@ -0,0 +1,705 @@
...
+        PO.TwoTerminalGenericHVDCLine, TwoTerminalGenericHVDCLine,
+        "TwoTerminalGenericHVDCLine", true,
+    ),
+    (PO.OnlineReserve, OnlineReserve, "OnlineReserve", true),
+    (PO.OfflineReserve, OfflineReserve, "OfflineReserve", true),
+    (PO.GroupReserve, GroupReserve, "GroupReserve", true),
+    (PO.TransmissionInterface, TransmissionInterface, "TransmissionInterface", true),
+]
+
+const DOCUMENT_PLAN_KEYS = Set(p[3] for p in DOCUMENT_PLAN)
```
daniel-thom: Should probably assert that the length of DOCUMENT_PLAN_KEYS is the same as DOCUMENT_PLAN.

src/openapi/import_document.jl
```diff
@@ -0,0 +1,705 @@
...
+end
+
+"""
+Document `time_series_type` string → the PSY/IS type to reconstruct, dispatched on below
+rather than branched on.
+"""
+const TIME_SERIES_TYPE_FROM_STRING = Dict{String, DataType}(
+    "SingleTimeSeries" => SingleTimeSeries,
+    "Deterministic" => Deterministic,
+    "DeterministicSingleTimeSeries" => DeterministicSingleTimeSeries,
```
daniel-thom, line 358: Why are Probabilistic and Scenarios not included?

src/openapi/import_document.jl
```diff
@@ -0,0 +1,705 @@
...
+        SingleTimeSeries, external_storage, single_metadata, 1:Int(assoc.length), 1:1,
+    )
+    IS.serialize_time_series!(IS.get_time_series_storage(entity), single_ts)
+    IS.add_metadata!(
+        IS.get_metadata_store(IS.get_time_series_manager(entity)), entity, metadata,
+    )
+    return nothing
+end
+
+"""Attach one `time_series_associations` row to `entity`. Dispatched on the resolved type —
```
daniel-thom, line 565: I'm not following this section. Why do we need it? Aren't we already loading the SQLite catalog from the .h5 file? Additionally, this is triggering a partial read of every time series, which is slow.

jd-lara: This is just temporary code to make the code work, I couldn't test without some way to reattach the table

src/openapi/import_document.jl
jd-lara: There are missing entries in the table for the scenario count and probabilistic. Also, we didn't have too many cases with this. I'd rather pick it up when infrastore is complete

src/openapi/export_document.jl
jd-lara: AI mistake

src/models/generated/ACBus.jl
```diff
@@ -149,3 +149,69 @@ set_area!(value::ACBus, val) = value.area = val
 set_load_zone!(value::ACBus, val) = value.load_zone = val
 """Set [`ACBus`](@ref) `ext`."""
 set_ext!(value::ACBus, val) = value.ext = val
+
+
```
luke-kiernan, line 153: Wait, is this a hand-edit to a generated file? Sure looks like it...

edit: never mind, it isn't.

src/models/generated/ACBus.jl
```diff
@@ -149,3 +149,69 @@ set_area!(value::ACBus, val) = value.area = val
...
+        angle = po.angle,
+        magnitude = po.magnitude,
+        voltage_limits = (if isnothing(po.voltage_limits); nothing; else; (min = po.voltage_limits.min, max = po.voltage_limits.max); end),
+        base_voltage = po.base_voltage,
+        area = resolve_ref(refs, po.area),
+        load_zone = resolve_ref(refs, po.load_zone),
+    )
+end
+
+function to_openapi(value::ACBus, refs::OpenAPIRefs, ::Val{:DEVICE_BASE})
```
luke-kiernan: > ::Val{:DEVICE_BASE}

Is there a reason why openAPI is using the old stateful enums instead of the new type hierarchy?

src/models/components.jl
```diff
@@ -10,10 +10,56 @@ field: the device base equals the system base.
...
+# independently-set device base (generators, loads, storage, ...) from the dozen
+# arc/area-ish types below whose `base_power` field only exists because the
+# schema records the system base per-component "in lieu of a system-level table"
+# (see SiennaSchemas). `add_component!` uses this trait to keep that field in
+# sync with the system's base power; it must never be set independently.
+abstract type BasePowerKind end
+struct DeviceBasePower <: BasePowerKind end
+struct SystemBasePower <: BasePowerKind end
+
+base_power_kind(::Component) = DeviceBasePower()
```
luke-kiernan, line 30: nitpick: dispatch flag so could return types not instances

src/models/serialization.jl
```diff
@@ -59,7 +59,7 @@ function IS.serialize(component::T) where {T <: _CONTAINS_SHOULD_ENCODE}
     IS.add_serialization_metadata!(data, T)
 
     # This is a temporary workaround until these types are not parameterized.
```
luke-kiernan: I don't follow...

src/openapi/cost_conversion.jl
```diff
@@ -0,0 +1,200 @@
...
+        "NATURAL_UNITS, DEVICE_BASE",
+    )
+end
+
+# ── FunctionData ────────────────────────────────────────────────────────────────
+
+convert_cost(fd::PC.LinearFunctionData) =
+    LinearFunctionData(fd.proportional_term, fd.constant_term)
+convert_cost(fd::PC.QuadraticFunctionData) =
+    QuadraticFunctionData(fd.quadratic_term, fd.proportional_term, fd.constant_term)
```
luke-kiernan, line 32: All of these `convert_cost` implementations have basically the same form: return a new object with only the curve part (e.g. no VOM cost). Can they be consolidated into 1 or 2 functions on supertypes?

luke-kiernan (commented):
Halfway through. Have other things to work on but will come back to it Monday

src/openapi/import_document.jl
jd-lara: fixed in 15ec5c47e DOCUMENT_PLAN rows are named tuples now, so the field is addable instead of a bare positional value.

src/openapi/import_document.jl
jd-lara: Fixed in 15ec5c47e

src/models/components.jl
jd-lara: This way the outcome matches the `UnitCategory` trait pattern already used in src/units/conversions.jl (singleton instances, dispatch on the value). Switching to type-dispatch here would mean rewriting both `_set_base_power!` and `_sync_base_power!` signatures for no benefit

src/openapi/cost_conversion.jl
jd-lara: resolved in 15ec5c47e

jd-lara:
For the tests to pass we need to merge the Schemas and OpenAPIModels changes

src/openapi/cost_conversion.jl
luke-kiernan, line 1: Flagging this file and the one that follows (`export_cost_conversion.jl`) as rather boilerplate. It's only ~200 lines each, though...so if there's no easy simplification, I'm okay with leaving things as-is.

src/units/conversions.jl
```diff
@@ -135,19 +129,16 @@ convert_units(gen, 0.6, POWER, DU, Float64)  # → 0.3 (raw SU value)
 """
 function convert_units end
 
-# --- From DU ---
-# `Real`/`ComplexF64` are split so a Unitful/RelativeQuantity value (both <: Number)
-# falls through to the marker/value guards instead. ComplexF64 mirrors serve the
-# complex admittance getters (`magnetizing_shunt`).
+# `_BareNumber` is `Real` plus `ComplexF64` (the complex admittance getters, e.g.
```
luke-kiernan: Rather strange and ad-hoc type definition. What if someone adds a field that's `ComplexF32`?

src/units/conversions.jl
```diff
@@ -193,34 +171,19 @@ convert_units(
 
 function convert_units(
     c,
-    value::Float64,
+    value::Union{Float64, ComplexF64},
```
luke-kiernan: Another weird type choice: is this precisely what you want, and if so, why?

Also I thought I wrote `convert_units` such that the final arg is the desired units. So this reads like: "convert a `ComplexF64` quantity in device base to a `Float64`", a request that doesn't make sense.

src/utils/IO/branchdata_checks.jl
```diff
@@ -85,31 +85,27 @@ const MVA_LIMITS_TRANSFORMERS = Dict(
...
+    for field in (:rating, :rating_b, :rating_c)
+        rating_value = getfield(line, field)
         if isnothing(rating_value)
-            @assert field ∈ [:rating_b, :rating_c]
+            @assert field ∈ (:rating_b, :rating_c)
             continue
         end
-        if (rating_value >= 2.0 * closest_rate_range.max / basemva)
-            @warn "$(field) $(round(rating_value*basemva; digits=2)) MW for $(get_name(line)) is 2x larger than the max expected rating $(closest_rate_range.max) MW for Line at a $(closest_v_level) kV Voltage level." _group =
+        if (rating_value * device_base_power >= 2.0 * closest_rate_range.max)
```
luke-kiernan: just make `rating_value*device_base_power` a variable please

src/openapi/import_document.jl
```diff
@@ -0,0 +1,971 @@
...
+end
+
+# ── document `ext` ──────────────────────────────────────────────────────────────
+
+"""
+Ledger of `doc.ext` keys already known to have no PowerSystems field, each with a one-line
+reason for its class.
+
+`ext` is a producer-side escape hatch for source columns no schema field claims. A key on this
+list is skipped on import without comment — it is never stored on the resulting `System`. A key
```
luke-kiernan: > A key on this list is skipped on import without comment — it is never stored on the resulting `System`.

wait, so an export-import round-trip will empty all `ext` dictionaries?

src/openapi/import_document.jl
```diff
@@ -0,0 +1,971 @@
...
+        end
+    end
+    return "unknown component id"
+end
+
+"""
+Error naming every `ext` key a document carries that is not on [`IGNORED_EXT_KEYS`](@ref),
+together with the component id and component type it was found on.
+
+`ext` is a producer-side escape hatch for source columns no schema field claims. A key not on
```
luke-kiernan: identical comment to earlier (control + F "producer-side escape hatch")

src/openapi/import_document.jl
```diff
@@ -0,0 +1,971 @@
...
+    assoc::PC.TimeSeriesAssociation,
+)
+    metadata = _scenarios_metadata(assoc)
+    rows = 1:length(metadata)
+    columns = 1:IS.get_count(metadata)
+    return IS.deserialize_time_series(Scenarios, storage, metadata, rows, columns)
+end
+
+"""
+Materialize (read off the HDF5 sidecar) the time series identified by `assoc`'s own
```
luke-kiernan: yeah this docstring basically unreadable. waaaay too much going on in that first sentence.

src/openapi/import_handwritten.jl
```diff
@@ -0,0 +1,793 @@
+# Hand-written (not generated): `from_openapi` methods for PO/PSY type pairs the IS generator
+# cannot emit. Each was attempted through the generator first (annotating the descriptor with
+# `openapi_type` and regenerating); every one below failed or would have silently miscompiled,
+# for one of these reasons:
+#
+#   - Arc:                       PSY's own field type is the ABSTRACT `Bus`, not a struct_name
```
luke-kiernan: Each of these explanations about "here's why this isn't auto-generated" is basically repeated in front of the function definition too...

src/openapi/import_handwritten.jl
```diff
@@ -0,0 +1,793 @@
...
+#   - TwoTerminalGenericHVDCLine:
+#                                `loss::Union{LinearCurve, PiecewiseIncrementalCurve}` is a Union
+#                                of two concrete curve types, not the generator's `Union{Nothing,
+#                                X}` nullable pattern — unclassifiable. Gained its own `base_power`;
+#                                every other field is now generator-shaped.
+#   - TransformerCircuit:        PSY's field is named `α`; the PO field is named `alpha`. The
+#                                generator assumes identical names and would silently emit
+#                                `po.α`, which does not exist on the PO struct — a silent
+#                                runtime defect, not a generation-time error.
+#   - TwoWindingTransformer:     `magnetizing_shunt::Complex{Float64}` is not a scalar, compound
```
luke-kiernan: sanity check: is it really the case that OpenAPI can't handle complex quantities? I'm surprised it's that dumb.

[Separately: could we extend things so that they are handled?]

src/openapi/import_document.jl
```diff
@@ -0,0 +1,971 @@
...
+
+# ── Time series ingestion ───────────────────────────────────────────────────────
+
+"""Scaling-factor multiplier names a document may carry, resolved to the PSY getter each
+one names. The document field is a bare, unprefixed function name (confirmed against
+real pointer files — the schema docstring's "PowerSystems.get_max_active_power"
+module-qualified form does not occur in practice), so this is a literal table rather than
+a runtime `getproperty(PowerSystems, Symbol(name))` dispatch — consistent with every
+other document-string lookup in this package. An unmapped string errors loudly."""
+const SCALING_FACTOR_MULTIPLIERS = Dict{String, Function}(
```
luke-kiernan, line 318: I'm skeptical that you really need a dict here given that every pair is `"get_foo" => get_foo`.

src/PowerSystems.jl
```diff
@@ -957,7 +1006,41 @@ include("emissions_data.jl")
...
 # Definitions of PowerSystem
 include("base.jl")
+
+# OpenAPI round-trip ledger (needs `System`, just defined in base.jl — see the note in
+# refs.jl for why this is a separate file from the rest of the OpenAPI serde code).
+include("openapi/ledger.jl")
+
 include("plant_attribute.jl")
+
+# OpenAPI document-level serde: dependency-ordered component pass + ServiceAssociation
```
luke-kiernan: What is "serde?" edit: oh, serializer/deserializer.

Also I'd shorten or eliminate the explanation of the include order here.

src/openapi/ledger.jl
luke-kiernan, line 1: Does this exist purely for testing purposes?

src/openapi/import_document.jl
```diff
@@ -0,0 +1,971 @@
...
+# RenewablePowerPlant, CombinedCycleBlock, CombinedCycleFractional, and the Core
+# GeographicInfo. `Substation` (`src/substation.jl`) has neither a schema nor a generated
+# PO model anywhere in SiennaSchemas/PowerOpenAPIModels — a real, reportable gap, not
+# something to invent a converter for.
+#
+# None of these embed unit-converted fields, so unlike the per-component converters they
+# take no `Val{unit_system}`; only the three `Outage` types need `refs`, to resolve
+# document-id `monitored_components` into the UUIDs PSY stores.
+
+const POLLUTANT_TYPE_FROM_STRING =
```
luke-kiernan: There's enough of these `const FOO_{FROM/TO}_STRING` dictionaries that I'd consider writing a macro.

src/openapi/import_document.jl
```diff
@@ -0,0 +1,971 @@
...
+
+    _check_no_unconverted_component_types(doc.components)
+    _check_ext_keys_are_known(doc)
+
+    sys = System(base_power; system_kwargs...)
+    _apply_document_metadata!(sys, doc)
+
+    refs = OpenAPIRefs(unit_system, base_power)
+
+    for (po_type, psy_type, key, addable) in DOCUMENT_PLAN
```
luke-kiernan: Have you checked how this is getting compiled? If this isn't getting unrolled, then might be losing performance here...

luke-kiernan (commented):
At some point we should go through and ruthlessly cut down on the comments: so much of it is duplicative, overly verbose, long-winded AI garbage.

Half of my comments are syntactical nitpicks. But now to locate the comment and delete it with this many files is a pain...so I'm leaving them.

I should add: in the interest of time, I didn't look at the tests, the docs, or the Claude files.

src/openapi/import_document.jl
jd-lara: This is going away, temp code for now to handle time series

src/openapi/ledger.jl
jd-lara: this is also code for the UUID -> ID mappings

luke-kiernan (approved):
Looks ok. I have some concerns about the sheer volume of code we're accumulating, but that's not fixable by small incremental changes. (Also: get it working first then make it nice.)

src/openapi/import_document.jl
josephmckinsey: In PowerSystemSchemas, we used some IS functions to directly manipulate the SQL as well as list metadata to export. It was reasonably performant from what I recall.

PR #1747, opened 2026-08-07
fix: resolve branch type override property for TapTransformer

mcllerena:
This PR fixes a parser edge case where parallel transformer branches are converted into TapTransformer through branch_type_override to better handle different parallel elements, but the input dictionary does not include group_number. That caused branch construction to fail in override scenarios.

When branch_type_override is TapTransformer, the branch can skip the normal type-inference path that populates vector group metadata. If group_number is absent, TapTransformer creation expects it and errors.

This is found through testing on new EI psse .raw data.

### DIFF
--- a/src/parsers/power_models_data.jl
+++ b/src/parsers/power_models_data.jl
@@ -1435,6 +1435,12 @@ function make_branch(
     branch_type_override::Union{DataType, Nothing} = nothing,
     kwargs...,
 )
+    if !isnothing(branch_type_override) &&
+       branch_type_override == TapTransformer &&
+       !haskey(d, "group_number")
+        _add_vector_control_group(d, "shift", "group_number")
+    end
+
     if !isnothing(branch_type_override)
         branch_type = branch_type_override
     elseif source_type == "matpower"
--- a/test/test_parse_psse.jl
+++ b/test/test_parse_psse.jl
@@ -752,6 +752,59 @@ end
     @test PowerSystems.get_branch_type_psse(d) == TapTransformer
 end
 
+@testset "make_branch TapTransformer override fills missing group_number" begin
+    bus_f = ACBus(;
+        number = 1,
+        name = "bus_f",
+        available = true,
+        bustype = ACBusTypes.PQ,
+        angle = 0.0,
+        magnitude = 1.0,
+        voltage_limits = (min = 0.9, max = 1.1),
+        base_voltage = 230.0,
+    )
+    bus_t = ACBus(;
+        number = 2,
+        name = "bus_t",
+        available = true,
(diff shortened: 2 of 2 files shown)


luke-kiernan:
> When branch_type_override is TapTransformer, the branch can skip the normal type-inference path that populates vector group metadata. If group_number is absent, TapTransformer creation expects it and errors.

So then what are we using as the group number (and other vector group metadata) in such instances? Do we have sensible defaults?

mcllerena:
> > When branch_type_override is TapTransformer, the branch can skip the normal type-inference path that populates vector group metadata. If group_number is absent, TapTransformer creation expects it and errors.
> 
> So then what are we using as the group number (and other vector group metadata) in such instances? Do we have sensible defaults?

The parser already handles those defaults; this change just reuses that existing behavior in the override path. Specifically, when a branch is converted into TapTransformer and group_number is missing, add_vector_control_group() is called so the same shift-to-group mapping used during normal parser inference is applied before construction.

mcllerena:
@luke-kiernan a detail explanation on the parsing error I found with the new EI files
1. The transformer block has three transformers in parallel between the same two buses, and one of them is out of service. Because of this, that record is not classified during the first "transformer" assignment in pti.jl and psse.jl.
2. Then in power_models_data.jl they are assigned the transformer type, in this case the TapTransformer. One of them did not have the "group_number" attached previously because it was unavailable, making the parser fail with this error:

> ERROR: KeyError: key "group_number" not found
Stacktrace:
  [1] getindex(h::Dict{String, Any}, key::String)
    @ Base ./dict.jl:477
  [2] make_tap_transformer(name::String, d::Dict{String, Any}, bus_f::PowerSystems.ACBus, bus_t::PowerSystems.ACBus; kwargs::@Kwargs{})
    @ PowerSystems ~/.julia/packages/PowerSystems/zOISI/src/parsers/power_models_data.jl:1804
  [3] make_tap_transformer
    @ ~/.julia/packages/PowerSystems/zOISI/src/parsers/power_models_data.jl:1767 [inlined]
  [4] make_branch(name::String, d::Dict{…}, bus_f::PowerSystems.ACBus, bus_t::PowerSystems.ACBus, source_type::String; branch_type_override::DataType, kwargs::@Kwargs{})
    @ PowerSystems ~/.j

3. These few lines of code in the PR aim to apply the _add_vector_control_group to the elements missing a group number.

jd-lara (commented):
I think this is a sensible change, we don't need to port it because in psy6 we parse all the cases. @m-bossart we probably only want to move the test to make sure that we catch this case.

jd-lara (approved):

PR #1756, opened 2026-08-14
Accept unit-tagged values in component constructors

### LINKED ISSUE #1115: When creating non-attached device, give option to use system_base or natural_units
Currently, by default is system_units (or I think device base)

luke-kiernan:
Closes #1115.

A component can now be constructed with natural-unit or device-base values while it is detached from a `System`:

```julia
ThermalStandard(; ..., active_power = 50.0u"MW", rating = 100.0u"MW", base_power = 100.0)
TransformerCircuit(; ..., r = 5.29u"Ω", base_power = 100.0, base_voltage_primary = 230.0)
ThermalStandard(; ..., active_power = 0.5 * DU)   # device base, tagged
ThermalStandard(; ..., active_power = 0.5)        # untagged: still device base
```

**Requires the companion struct-generator change, Sienna-Platform/InfrastructureSystems.jl#606.** `[sources]` in `Project.toml`, `test/Project.toml` and `docs/Project.toml` points at that IS branch so CI resolves; each carries a TODO to restore `rev = "IS4"` once it merges.

## Design

Getters and setters resolve their per-unit bases from the component. A constructor has no component yet — and no `System` to fall back on — so the bases come from the constructor's own arguments. The generated constructors collect them into a `_construction_fields` NamedTuple and call `construct_value` once per unit-bearing field.

Key pieces, all in `src/models/components.jl`:

- `UnderConstruction{T, F, P, V}` — isbits carrier for one field's bases; `T`/`F` are type parameters used only for error messages. It joins `UnitsBearer`, so it is a first-class base provider for the existing engine rather than a parallel mechanism.
- `_construction_base` mirrors `_conversion_base`, reading constructor arguments instead of component fields. The generic resolver walks `base_power`, then `base_voltage_primary` → `base_voltage` → `arc.from` → `bus`. Overrides cover the transformers' `magnetizing_shunt` (delegates to the already-formed `TransformerCircuit`, which is its own base provider) and the six 3W pairwise impedances (`base_power_ij` referenced to the first-index circuit's voltage, matching the existing `PairBase` convention).
- `_construct_value` delegates every tagged value to `set_value`, so **the conversion math is not duplicated** — there is still exactly one conversion engine. It adds only what the setters cannot have: untagged `Real`/`Complex` pass through as device base, and compound fields (`MinMax`, `UpDown`, `FromTo`, `FromTo_ToFrom`, `StartUpShutDown`) recurse so untagged entries inside them keep working.

## Behavior notes

Untagged numbers keep meaning device base, so every existing call site is unaffected. (The setters reject bare floats; the constructors cannot, since that would break essentially all existing construction. Making units mandatory here is a separate, later step.)

Two cases throw, with the field named and an actionable message:

1. **`SU` values, always** — the system base is unknown before `add_component!`.
2. **Natural units on structs with no `base_power` field** (transmission, areas, reserves). Their device base *is* the system base, so `Line(; r = 5.29u"Ω")` is not answerable while detached — it errors rather than guessing. Device-base values work for all of them, and the transformers' own data lives on `TransformerCircuit`, which carries its bases and converts fine.

## Verification

- **Full suite: 8165 pass, 0 fail**, including `test_generate_structs` after regenerating all 40 affected files.
- New `Units-aware construction` testset in `test/test_units.jl` (101 → 129 tests): `MW`/`DU`/untagged agreement across both the positional and keyword forms; construction ≡ construct-then-set; `TransformerCircuit` from Ω while detached; 3W pairwise against `base_power_12`; and both error paths.
- Conversion costs ~5 ns and 0 bytes per field. Constructors stay fully type-inferred, and both forms are at or below their previous cost, which is dominated by uuid4 in `InfrastructureSystemsInternal()`:

  | | ns | allocs | before |
  |---|---|---|---|
  | `ThermalStandard` positional, untagged | 1154 | 6 | 1170–1295 / 8 |
  | `ThermalStandard` positional, `MW` | 1150 | 6 | — |
  | `ThermalStandard` kwarg, untagged | 1221 | 8 | 1512 / 10 |
  | `Line` positional, untagged | 1146 | 5 | 1133 / 5 |

Docs are not addressed here: `docs/make.jl` has failing `@example` blocks and unresolved `@ref`s on `psy6` that I did not attempt to attribute or fix.


jd-lara:
@luke-kiernan you should stack this PR on top of https://github.com/Sienna-Platform/PowerSystems.jl/pull/1746 since we moved the code generation here

luke-kiernan:
> you should stack this PR on top of https://github.com/Sienna-Platform/PowerSystems.jl/pull/1746 since we moved the code generation here

Yeah I realized that this would overlap with the OpenAPI schema stuff. I could merge these changes into that one, but there's already a lof of stuff going on over there...I'd be more inclined to have this PR wait until the schemas get merged. Will be a bit more tedium on my side (rebase after massive changes) but that's fine

luke-kiernan:
Jose has asked me to switch this to a stacked PR on top of #1746 . TBD whether I rebase and use this same PR or just close this one and reopen. Reverting to draft for now

luke-kiernan:
Closing and using #1759 instead

PR #1758, opened 2026-08-17
Integrate with IS that supports time series in InfraStore

daniel-thom:


docs/src/how_to/use_context_managers.md
```diff
@@ -112,16 +118,16 @@ data = Dict(
...
-# Use context manager for efficient bulk addition
-begin_time_series_update(sys) do
+# Use a transaction for efficient bulk addition
+time_series_transaction(sys) do txn
     for (i, gen) in enumerate(generators)
         forecast = Deterministic(
             "max_active_power",
             data,
             resolution;
             scaling_factor_multiplier = get_max_active_power,
```
rodrigomha: Is the scaling factor multiplier still being used?

jd-lara: not anymore

PR #1760, opened 2026-08-18
Display units: spell out DU/SU, split power natural units, dynamics footer

### LINKED ISSUE #1753: `display(gen)`: units on fields
@kdayday and @daniel-thom have voiced some concerns about the current `display(gen)` output. The current behavior:

> 1. Unattached components: fields with "needs_conversion": true are in natural units explicitly, e.g. active_power: 0.0 MWA. edit: ok I'd put rating in device base.
> 2. Attached components: base power in natural units. Rating in device base. Other fields with "needs_conversion": true display system base. 
> 3. Again, explicitly print SU/DU/[appropriate natural unit] for maximum clarity.
> Add an option to pass a different unit system as a keyword argument, in which case all are displayed in that unit system.

These things are pretty easy to change, but the question is: change to what?

One thing that came up (Kate I believe): `SU` and `DU` are useful shorthands, but aren't standard terminology. Maybe we should write out `system base` or `p.u. (system)` instead.

luke-kiernan:
Closes #1753 (all four action items).

## What changed

`display(gen)` says what its numbers mean. Per-unit markers are spelled out, active/reactive/apparent power read back in their own natural units, and dynamic models — which aren't wired into the units engine — say so instead of printing bare numbers.

**Before**

```
ThermalStandard: g1:
   active_power: 1.25 SU
   reactive_power: 0.25 SU
   rating: 1.0 DU
   active_power_limits: (min = 0.0 SU, max = 2.5 SU)
   base_power: 250.0 MVA

BaseMachine(0.0, 0.2, 1.0, Dict{String, Any}(), Symbol[], 0)
```

**After**

```
ThermalStandard: g1:
   active_power: 1.25 p.u. in system base
   reactive_power: 0.25 p.u. in system base
   rating: 1.0 p.u. in device base
   active_power_limits: (min = 0.0 p.u., max = 2.5 p.u.) in system base
   base_power: 250.0 MVA

BaseMachine:
   R: 0.0
   Xd_p: 0.2
   eq_p: 1.0
   ext: Dict{String, Any}()
   (parameters in p.u. on the parent device's base power)
```

Detached, the same generator now reads `reactive_power: 25.0 Mvar` and `rating: 250.0 MVA` where both used to say MW. `show_components` also takes per-column units: `units = Dict(:active_power => MW, :rating => DU)`.

## How

The natural-unit change is the substantive one: the single `POWER` category becomes three dispatch tokens — `ACTIVE_POWER` (MW), `REACTIVE_POWER` (Mvar), `APPARENT_POWER` (MVA) — under a shared `AbstractPowerCategory`. They share a dimension and a per-unit base, so every base, ratio, and conversion rule stays written once against the supertype; only `natural_unit` differs per token. Descriptor fields carry `:mw`, `:mvar`, or `:mva` accordingly (169 reclassified: 81 / 60 / 28), and the generated diff is token-only. This is at the getter level, not just display — `get_reactive_power_unitful(gen, NU)` returns `25.0 Mvar`.

Two fields kept their current behaviour rather than getting invented categories, since both are on the follow-up list in #1753: `storage_capacity` (really MWh) and `ramp_limits` (really MW/min) stay `:mw`.

## Depends on

InfrastructureSystems.jl#609, for `display_string` and the per-column `units` resolution. The `[sources]` pins in `Project.toml`, `test/Project.toml`, and `docs/Project.toml` point at that branch and go back to `rev = "IS4"` once it merges.


src/models/components.jl
```diff
@@ -346,8 +346,13 @@ get_x(t::TwoWindingTransformer, units) = get_x(get_circuit(t), units)
 get_x_unitful(t::TwoWindingTransformer, units) = get_x_unitful(get_circuit(t), units)
 set_x!(t::TwoWindingTransformer, val) = set_x!(get_circuit(t), val)
 
-# Physical category implied by a field's conversion unit.
-_unit_category(::Val{:mva}) = POWER
+# Physical category implied by a field's conversion unit. The three power tokens
```
luke-kiernan, line 349: We should still get the same performance with 5 unit categories here as we get with 3--union-splitting doesn't apply because these are hard-coded compile time values largely--but might be worth checking.

src/generate_structs.jl
```diff
@@ -185,8 +185,15 @@ const OPENAPI_COMPOUND_MEMBERS = Dict(
     "StartUpStages" => ("hot", "warm", "cold"),
     "TurbinePump" => ("turbine", "pump"),
 )
-const OPENAPI_CONVERSION_KINDS =
-    Dict(":mva" => :power, ":ohm" => :impedance, ":siemens" => :admittance)
+# The three power tokens share one per-unit base (S_base) and so one conversion
+# rule; they differ only in the natural unit the units engine prints them as.
+const OPENAPI_CONVERSION_KINDS = Dict(
```
luke-kiernan, line 190: Pre-existing and separate from this PR: should Open API distinguish between the 3 varieties of power?

luke-kiernan (commented):
Commenting on my own PR because Claude did 80% of this.

src/units/conversions.jl
```diff
@@ -62,7 +74,9 @@ const CURRENT = CurrentCategory()
 
 The natural (physical) unit for this category.
 """
-natural_unit(::PowerCategory) = u"MW"
+natural_unit(::ActivePowerCategory) = MW
+natural_unit(::ReactivePowerCategory) = Mvar
```
kdayday: Capitalization preference for "MVAr"

kdayday (commented):
This looks aligned to the direction I think we should go, but I have these concerns:
1. I think we should get internal agreement on the "acceptance criteria" doc about how units look in multiple places before finalizing an approach
2. As part of that, I'm going to make that argument that pretty prints by default report values in natural units, not system base (dynamics and branch resistance/reactance being exceptions)
3. Reiterating my concern that there are now two units definitions between fields and time series (https://github.com/Sienna-Platform/SiennaSchemas/blob/main/Core/units.json), with slightly divergent syntax (i.e., `ACTIVE_POWER` type vs. `ActivePower` enum member) and two places to maintain. This comes back to point (1), and my thought that if a more comprehensive system is being developed for time-series, working on this in parallel risks redundant work while we're still trying to understand if the time series stuff is reusable or not

luke-kiernan:
> 1. I think we should get internal agreement on the "acceptance criteria" doc about how units look in multiple places before finalizing an approach.

I'm on board with that. It's pretty easy to change display level stuff, names, how things look.

> 2. As part of that, I'm going to make that argument that pretty prints by default report values in natural units, not system base

I went with system base because that's the scale we do math in most often. E.g. ACPF bus voltages are around 1.0 p.u. There's some precedent: in PSY5, if you call `PSB.build_system`, the system you'll get is almost always in SU. Natural units has a nice simplicity to it, though: it always makes sense, whether the component is attached to a system or not.

Either way, defaults can be overridden, so I'm not losing sleep over which we choose. @jd-lara  do you have opinions here?

> 3. Reiterating my concern that there are now two units definitions between fields and time series (https://github.com/Sienna-Platform/SiennaSchemas/blob/main/Core/units.json), with slightly divergent syntax (i.e., ACTIVE_POWER type vs. ActivePower enum member) and two places to maintain. This comes back to point (1), and my thought that if a more comprehensive system is being developed for time-series, working on this in parallel risks redundant work while we're still trying to understand if the time series stuff is reusable or not

I acknowledge your concerns here...but I'm not sure what that means in terms of a practical course of action.

Worth mentioning: Gabriel secured that VAM funding for units work, and I had to use that up by June. Frankly, waiting wasn't an option. So I overhauled the stateful units system on that timeline. As an unfortunate side effect, my changes got incorporated into psy6 without a whole lot of buy in from folks (beyond Jose). And yeah, that's a problem. But again, not sure what to do with that. Well, besides PRs like this: tweak display options to bring it in line to what users expect.

src/units/conversions.jl
luke-kiernan: sure will change

PR #1763, opened 2026-08-19
fix issue 1762 and add regression test

### LINKED ISSUE #1762: Switched shunt parsing: v33 initial blocks status should be zero
In PSS(E) raw v33, switched shunts don't carry per-block status, so we currently fabricate data and use [all ones](https://github.com/Sienna-Platform/PowerSystems.jl/blob/main/src/parsers/pm_io/psse.jl#L737). That's causing us to double-count. Looks like we should switch to all zeros.

I've come across this issue before and patched it [here](https://github.com/Sienna-Platform/PowerSystems.jl/blob/main/src/parsers/pm_io/psse.jl#L742) for `MODSW = 0/1/2`. Now I'm running into the same problem but for `MODSW = 3`. [We don't support active adjustment of `MODSW=3` in PF's controlled reactive power loop stuff yet, but we still include its `Y_base` value in the network [here](https://github.com/Sienna-Platform/PowerFlows.jl/blob/main/src/common.jl#L137). And right now that number is coming out twice as big as it should be.]

luke-kiernan:
See issue #1762

### DIFF
--- a/src/parsers/pm_io/psse.jl
+++ b/src/parsers/pm_io/psse.jl
@@ -734,16 +734,9 @@ function _psse2pm_shunt!(pm_data::Dict, pti_data::Dict, import_all::Bool)
                 sub_data["ext"]["NREG"] = pop!(switched_shunt, "NREG")
             elseif pm_data["source_version"] ∈ ("32", "33")
                 sub_data["ext"]["SWREM"] = switched_shunt["SWREM"]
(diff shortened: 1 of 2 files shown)


mcllerena (approved):

luke-kiernan:
Closing in favor of a comprehensive fix: see #1774

PR #1767, opened 2026-08-20
Fix round trip of supplemental attribute time series; add attribute time series tests

daniel-thom:
## Summary

Two commits:

### 1. Tests for time series on supplemental attributes (`test/test_outages.jl`)

`create_system_with_outages()` already attached a `SingleTimeSeries` to each outage, but the only existing assertion was a removal loop — nothing read attribute-owned time series back. New testsets:

- **Time series on supplemental attributes** — reads every fixture series back (`has_time_series` in all arities, `get_time_series`, `_values`/`_timestamps`/`_array`, windowed reads, key lookup); ownership isolation between a generator and its attributes; `GeographicInfo` rejects series (`supports_time_series` default); counts and both summary tables report `SupplementalAttribute` rows; feature-tagged static series outside a transaction, a `Deterministic` forecast, one series on several attributes at once; `copy_time_series!`; detached attribute rejected; removal by name/feature and per attribute.
- **Colliding ids** — components and supplemental attributes draw ids from independent streams. Four generators and four outages share ids 2–5, each outage attached to a *different* generator than the one sharing its id. Both owner kinds get same-named static series and forecasts with different values; every owner reads back exactly its own, before and after a `to_file`/`from_file` round trip; removing one owner's series or an attribute leaves the id-sharing owner's series intact.
- **Attribute removal cascade** and **serialization round trip** (two round trips, values compared per outage).

### 2. Fix: `to_file`/`from_file` cross-wired attribute-owned time series

The sidecar store was serialized wholesale with attribute-owned series keyed by the exporter's IS attribute ids, while the importer minted fresh attribute ids in association order — each series landed on whichever attribute received the same number. The guard meant to reject this (`_check_time_series_ids_match`) never fired: `to_openapi` snapshotted `sorted_refs` before `_export_supplemental_attributes` registered the attributes, so the time-series walk never saw one either.

- **Export** (`src/openapi/export_document.jl`): re-snapshot the registry after attributes are registered; emit `TimeSeriesAssociation` rows for them under their document ids; after writing the sidecar, re-key their rows in that copy from IS id → document id with `InfraStore.replace_owner!` (two passes through a staging range above every attribute id, since one attribute's document id can be another's IS id). The live store is never touched. Dead guard removed.
- **Import** (`src/openapi/sqlite_load.jl`): `IS.set_id!(attribute, document_id)` before attaching, mirroring what components already do, so the adopted store resolves. `IS.assign_id!` keeps the preset id and advances the attribute counter.

Bundles written before this fix keep attribute series under the old IS ids; per the no-shim policy they should be regenerated rather than bridged.

Also noted: IS bulk `remove_time_series!(sys, T)` is deliberately scoped to component owners; the test encodes that, but PSY's docstring does not mention it.

## Test plan

- [x] `test_outages` → 356 passed, 0 failed, 0 broken
- [x] `test_openapi_document test_openapi_export test_openapi_file_io test_openapi_sqlite_load` → 838 passed
- [x] formatter run
- `test_serialization` errors locally on `Pkg.dependencies()` (`PowerTableDataParser` has no source path in `test/Manifest.toml`) — environment, independent of this change; `Pkg.dependencies()` fails in the bare test env too.

### DIFF
--- a/src/openapi/export_document.jl
+++ b/src/openapi/export_document.jl
@@ -535,30 +535,52 @@ end
 # same tuple `TimeSeriesAssociation` carries — so no `time_series_associations` rows are
 # emitted. PowerTableDataParser writes its documents the same way.
 #
-# The catalog's owner ids are IS component ids, while the document's ids come from
-# `_export_id!`. `_check_time_series_ids_match` below refuses to write a pair whose ids
-# disagree, since the result would be a document whose sidecar no importer could resolve.
+# The catalog's owner ids are IS ids. A component's document id is its IS id, so its rows
+# resolve on import by construction. A supplemental attribute's document id is minted fresh
+# from the document counter (its IS id comes from a stream independent of the components'
+# and would collide in the document), so after the sidecar is written
+# `_remap_sidecar_attribute_owners!` re-keys the attribute-owned rows in that copy to the
+# document ids; import presets each attribute's IS id to its document id, and the rows
+# resolve the same way a component's do.
 
 """
-Error when a supplemental attribute owns time series, which export cannot represent.
+Re-key every attribute-owned series in the sidecar just written at `path` from the
+attribute's IS id to the document id it was exported under, so the rows resolve on import
+the way component rows do.
 
-Component-owned series need no check: a component's document id is its IS component id, so
-the sidecar catalog's owner ids and the document's ids agree by construction. Attribute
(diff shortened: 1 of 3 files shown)


daniel-thom:
This is no longer relevant because we will have one stream of ids for components and supplemental attributes.

PR #1769, opened 2026-08-20
replace local dev'ed path with remote branch

luke-kiernan:
What the title says. [I decided to set up a dedicated psy6 folder and this is what I had to do to get the tests to compile and pass. The two dependencies I added here are only needed transitively, so that PSB doesn't try to grab the `main` branch of them.]


luke-kiernan:
Errors in CI. I got the same errors locally at first, but running `Pkg.update` on the dev'ed dependencies fixed it. Very strange. It's like it's using an older, cached `psy6` snapshot or something.

jd-lara:
Fixed on #1758

PR #1781, opened 2026-09-01
Type the HVDC loss fields as LossCurve

luke-kiernan:
Types the five curve-valued loss fields as `LossCurve` (InfrastructureSystems.jl#625), so a loss curve carries the power basis it is expressed in rather than leaving it implicit.

| Struct | Field |
|---|---|
| `TwoTerminalGenericHVDCLine` | `loss` |
| `TwoTerminalLCCLine` | `loss` |
| `TwoTerminalVSCLine` | `converter_loss_from`, `converter_loss_to` |
| `InterconnectingConverter` | `loss_function` |

Each keeps the curve shapes its `Union` already admitted, now wrapped — `Union{AnyLossCurve{LinearCurve}, AnyLossCurve{QuadraticCurve}}` and so on — defaulting to `LossCurve(LinearCurve(0.0), NaturalUnit())`. `power_units` has no default in IS by design, so the basis is named explicitly.

These are the **only** curve-typed loss fields in the descriptor. Pumped hydro expresses its losses as a scalar `efficiency::TurbinePump`, and `evaporative_loss` / `standing_loss` / `active_power_losses` are `Float64` — none are `ValueCurve`s, so none are candidates.

## The OpenAPI asymmetry

The OpenAPI schemas have no `LossCurve`: a document's loss field is a bare `ValueCurve` (`$ref: InputOutputCurve` / `TwoTerminalLoss`) with nowhere to record a basis. Until the schemas gain the type the two directions cannot be symmetric, so one helper pair in `cost_conversion.jl` carries the contract:

- **Import** synthesizes `NaturalUnit()` — the basis every current producer writes, and the one the schema's own `UnitSystem` default names.
- **Export refuses** any other basis rather than silently dropping it.

The guard matters: without it, a `LossCurve` on a relative basis would round-trip back as `NaturalUnit` with no diagnostic. Erroring instead of guessing mirrors the posture `admittance_units` / `voltage_units` already take on these same HVDC types, where only the implemented basis passes. Nothing currently produces a non-natural loss curve, so the error path should stay unreached; when the schemas gain `LossCurve`, the guard is deleted and export carries `power_units` for real.

## Notes for review

- Import funnels through `_hvdc_loss` and `_vsc_converter_loss`, so wrapping those two covers all eight call sites. Export funnels through `_hvdc_loss_to_openapi` plus four direct calls.
- `src/deprecated.jl` keeps working: `_deprecated_hvdc_loss` passes an `AnyLossCurve` through and wraps a bare `ValueCurve`. These shims exist to accept old spellings, so breaking the bare-curve form would defeat their purpose.
- The `_io_curve` sites in `test_openapi_converters.jl` are deliberately left bare — they build `PSY.PC.InputOutputCurve`, the document side, which has no `power_units`. That is the asymmetry working as intended.
- No other code in `src/` reads these fields, so the blast radius is the converters plus the deprecation shims.
- Three unrelated regenerated files (`TradingHub`, `PointToPointBid`, `VirtualParticipant`) were reverted to keep this focused. Regenerating them adds a two-arg `from_openapi(po, refs)` method missing from the versions committed with the market-component family — pre-existing drift, and `test/test_generate_structs.jl` documents that no freshness check catches it.

## Testing

Full suite, against a same-day baseline of clean `psy6`:

| | Pass | Fail | Error |
|---|---|---|---|
| baseline `psy6` | 12687 | 9 | 26 |
| this branch | 12688 | 8 | 26 |

**No failure is unique to this branch.** Every remaining one reproduces on clean `psy6` — chiefly `get_variable(::ThermalGenerationCost)` (the cost-field rename) and `MarketBidTimeSeriesCost` JSON round-trip (the IS4 thin-key rework). `test_printing.jl:78` fails on the baseline but not here, so that suite has some nondeterminism either way.

## Pins — revert all three once upstream merges

- `InfrastructureSystems` → `lk/loss-curve-units-v2` (IS#625, rebased onto current `IS4`)
- `PowerSystemCaseBuilder` → `lk/loss-curve-type` (PSCB#217) — **merge that first**

The PSCB pin was already stale at `jd/power-units-rename`. PSCB's `psy6` contains that commit fully and is five ahead of it, and this branch is based on `psy6`, so the move is forward, not a downgrade.

Unrelated but adjacent: psy6's `[sources]` comment claims IS4 "still requires the old (now-unregistered) `PowerTimeSeriesOpenAPIModels` and fails to resolve" and pins IS to `jd/openapi-package-split`. That merged into `IS4` earlier today, so the comment and pin are both stale — worth a separate cleanup.

https://claude.ai/code/session_014eHyC9XpEJSiH27VsFcQr4

### DIFF
--- a/src/models/generated/InterconnectingConverter.jl
+++ b/src/models/generated/InterconnectingConverter.jl
@@ -17,7 +17,7 @@ This file is auto-generated. Do not edit.
         reactive_power_limits::Union{Nothing, MinMax}
         dc_current::Float64
         max_dc_current::Float64
-        loss_function::Union{LinearCurve, QuadraticCurve}
+        loss_function::Union{AnyLossCurve{LinearCurve}, AnyLossCurve{QuadraticCurve}}
         dc_control::VSCDCControlModes
         ac_control::VSCACControlModes
         dc_setpoint::Float64
@@ -47,7 +47,7 @@ Interconnecting Power Converter (IPC) for transforming power from an ACBus to a 
 - `reactive_power_limits::Union{Nothing, MinMax}`: (default: `nothing`) Minimum and maximum reactive power limits. Set to `Nothing` if not applicable
 - `dc_current::Float64`: (default: `0.0`) DC current on the converter, in per unit power-equivalent on the converter `base_power` (I is approximately P at 1.0 pu DC voltage)
 - `max_dc_current::Float64`: (default: `1e8`) Maximum stable DC current limit, in per unit power-equivalent on the converter `base_power` (I is approximately P at 1.0 pu DC voltage)
-- `loss_function::Union{LinearCurve, QuadraticCurve}`: (default: `LinearCurve(0.0)`) Linear or quadratic loss function with respect to the converter current
+- `loss_function::Union{AnyLossCurve{LinearCurve}, AnyLossCurve{QuadraticCurve}}`: (default: `LossCurve(LinearCurve(0.0), NaturalUnit())`) Linear or quadratic loss function with respect to the converter current
 - `dc_control::VSCDCControlModes`: (default: `VSCDCControlModes.DC_VOLTAGE`) DC-side control mode of the converter; see [`VSCDCControlModes`](@ref).
 - `ac_control::VSCACControlModes`: (default: `VSCACControlModes.AC_REACTIVE_POWER`) AC-side control mode of the converter; see [`VSCACControlModes`](@ref).
 - `dc_setpoint::Float64`: (default: `0.0`) DC-voltage target (when dc_voltage_control is true) or active-power order (when false), in per unit.
@@ -86,7 +86,7 @@ mutable struct InterconnectingConverter <: StaticInjection
     "Maximum stable DC current limit, in per unit power-equivalent on the converter `base_power` (I is approximately P at 1.0 pu DC voltage)"
     max_dc_current::Float64
     "Linear or quadratic loss function with respect to the converter current"
-    loss_function::Union{LinearCurve, QuadraticCurve}
+    loss_function::Union{AnyLossCurve{LinearCurve}, AnyLossCurve{QuadraticCurve}}
     "DC-side control mode of the converter; see [`VSCDCControlModes`](@ref)."
     dc_control::VSCDCControlModes
     "AC-side control mode of the converter; see [`VSCACControlModes`](@ref)."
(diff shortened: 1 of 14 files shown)


luke-kiernan:
Rebased onto today's `psy6` (`08c017c25`) — merge conflicts resolved.

The conflicts were in `Project.toml` and `test/Project.toml`, and were exactly the pin lines: `psy6` had just reverted `InfrastructureSystems` → `IS4` and `PowerSystemCaseBuilder` → `psy6` (the stale-pin cleanup I noted in the description). I kept the co-dev pins, since they have to stay until the upstreams land.

Also note PowerSystemCaseBuilder#217 now pins **this** branch, so the two are mutually pinned for CI. That is fine — `[sources]` apply only to the top-level project and are not inherited, so each side resolves the other's branch without the cycle becoming a resolution problem.

## Unwind order

1. Merge PowerSystemCaseBuilder#217.
2. Here: revert `PowerSystemCaseBuilder` → `psy6`, then merge.
3. On PSCB: revert `PowerSystems` → `psy6` and `InfrastructureSystems` → `IS4`.

The `InfrastructureSystems` pin here comes out once InfrastructureSystems#625 merges into `IS4`.

luke-kiernan:
**Correction to the test numbers in the description.**

It claims 12688 pass / 8 fail / 26 error, "no failure unique to this branch", with the remainder pre-existing on `psy6`. That was measured against a stale environment and is wrong.

`Project.toml` pins the seven OpenAPI model packages at `rev = "main"`, but my `Manifest.toml` had them resolved at `jd/schemas-0.1.0` — the rev the *older* `psy6` pinned before this branch was rebased. `Pkg.instantiate()` honours an existing manifest and will not re-resolve a changed `[sources]` rev; only `update`/`resolve` will. The baseline I compared against was resolved at the same time, so both sides saw the same stale packages and were broken identically — which is why the failures looked pre-existing.

After `Pkg.update()` in both environments:

```
Begin PowerSystems tests | 14833  14833  3m50.6s
     Testing PowerSystems tests passed
```

**14833 / 14833, zero failures, zero errors.**

So the failures I attributed to `psy6` — `get_variable(::ThermalGenerationCost)`, the `MarketBidTimeSeriesCost` JSON round trip, the time-series errors — were artifacts of the stale OpenAPI packages, not real breakage. Apologies for the noise in the description; treat this comment as the result.

Nothing about the change itself moved. The PowerSystemCaseBuilder failures were real and are still fixed by PowerSystemCaseBuilder#217 — that was a genuine `MethodError` from constructing a loss field with a bare curve, independent of which OpenAPI packages were resolved.

src/models/generated/InterconnectingConverter.jl
```diff
@@ -17,7 +17,7 @@ This file is auto-generated. Do not edit.
         reactive_power_limits::Union{Nothing, MinMax}
         dc_current::Float64
         max_dc_current::Float64
-        loss_function::Union{LinearCurve, QuadraticCurve}
+        loss_function::Union{AnyLossCurve{LinearCurve}, AnyLossCurve{QuadraticCurve}}
```
jd-lara, line 20: Don't we have a better abstract instead of a massive union?

luke-kiernan: If we want to limit the curve to linear or quadratic at compile time without constraining the units, then this is what it has to be. The alternatives: widen the type to `AnyLossCurve` rejecting non-(linear/quadratic) at runtime, or switch to allowing for non-(linear/quadratic) here.

The thing preventing one nice abstract supertype: HVDC loss curves are linear or piecewise, while inter connecting converter loss curves are linear or quadratic.

PR #1782, opened 2026-09-01
Explain the missing units argument in generated accessors

luke-kiernan:
## Problem

With explicit units, `get_active_power(gen)` is an error — but the error Julia reports is a bare `MethodError` listing the two-argument signatures, which never says what a valid unit argument *is*. Setters were only slightly better: a bare float hit a generic `"setter requires explicit units"` that named neither the setter nor the field, and a bare **integer** fell through to a `MethodError` entirely.

## What this does

The accessor templates in `src/generate_structs.jl` now emit fallback methods for every convertible field, alongside the working accessors:

- a one-argument getter, and its `_unitful` companion;
- a setter taking an untagged `Real`;
- a setter taking a compound value whose elements are all untagged.

They route to `_units_arg_required` / `_units_tag_required` in `src/models/components.jl`, which name the accessor, the owning type, the field, and the units that would have worked. Both share one set of message builders (`_natural_unit_example`, `_field_description`, `_units_menu`, `_tag_menu`), so the getter and setter messages stay in step and each field advertises its own natural unit (`MW`, `MVAr`, `MVA`, `OHMS`, `SIEMENS`).

```
ArgumentError: `get_active_power` requires an explicit units argument:
`get_active_power(component, units)`. `ThermalStandard`'s `active_power` is a per-unit
quantity with no default unit system, so the units must be named at the call site: `DU`
(per unit on the device base), `SU` (per unit on the system base), `NU` or the natural
unit `MW`. For the unit-bearing value instead of a bare number, use
`get_active_power_unitful(component, units)`.

ArgumentError: `set_active_power_limits!` requires a units-tagged value:
`set_active_power_limits!(component, val * units)`. … A compound field takes one tagged
value per element, e.g. `(min = 0.0 * DU, max = 1.0 * DU)`.
```

The setter menu deliberately omits `NU`: it is a getter target only, and `val * NU` is not even constructible.

Dispatch is safe because `Unitful.Quantity` and `IS.RelativeQuantity` are `Number` but not `Real`, so tagged values still reach the working accessor. `TwoWindingTransformer`'s hand-written `get_r`/`get_x`/`set_r!`/`set_x!` forwarding gained matching arities so they reach the circuit's generated fallback rather than a `MethodError`.

## Two fixes that fell out of it

- `set_value`'s bare-number rejection matched `::Float64`, so `set_active_power!(gen, 1)` produced a raw `MethodError`. It now matches `::Real`, and its message names the field.
- `_to_device_base` discarded the field (`set_value(c, nothing, val, cu)`), so a bare element inside a compound could not report which field it belonged to; the field is now threaded through. No behavior change today — every compound field uses the default base provider — but the SU branch's `_conversion_base(c, field)` now gets the real field, which is the correct value if a compound field ever needs a field-specific one.

## Testing

- Full suite: 14852/14852 pass on this branch.
- New testsets in `test/test_base_power.jl` cover, for getters and setters: the error type, the message naming the accessor/field/type/unit menu/natural unit, the `_unitful` companion pointing back at the bare getter (not `..._unitful_unitful`), bare float, bare integer, all-untagged compound, partially tagged compound (caught one level down, still names the field), and round-trips through every accepted tagged form.
- Docs were **not** built: `julia --project=docs -e 'Pkg.instantiate()'` fails in this checkout with `Did not find subdirectory PowerTimeSeriesOpenAPIModels.jl` while resolving PowerOpenAPIModels, unrelated to this change. No docstrings were added or altered.

## One thing to know before the next regeneration

The generator re-emits three `from_openapi(po, refs)` two-arg methods (TradingHub, VirtualParticipant, PointToPointBid) that are hand-written in `src/openapi/import_handwritten.jl`; leaving them in breaks precompilation with *"Method overwriting is not permitted"*. They had already been deleted from the committed generated files, so I stripped them out again to keep this diff to the units change — but the generator/committed-file divergence is pre-existing and will bite the next person who regenerates.

https://claude.ai/code/session_013JSuicq3FsLGV29kFyoHpr

### DIFF
--- a/src/generate_structs.jl
+++ b/src/generate_structs.jl
@@ -15,6 +15,13 @@ const IS = InfrastructureSystems
 const MU = IS.Mustache
 import InfrastructureSystems: DataFormatError
 
+# Convertible fields also get fallback accessor methods for the calls that omit the
+# units — a one-argument getter, and setters taking an untagged number or an
+# all-untagged compound. They route to `_units_arg_required`/`_units_tag_required`
+# (hand-written, `src/models/components.jl`), which name the accessor, the field and
+# the units that would have worked. Without them, `get_active_power(gen)` and
+# `set_active_power!(gen, 1)` raise bare `MethodError`s that list signatures but never
+# say what a valid unit is.
 const STRUCT_TEMPLATE = """
 #=
 This file is auto-generated. Do not edit.
@@ -82,6 +89,8 @@ end
 {{accessor}}(value::{{struct_name}}, units) = InfrastructureSystems._strip_units(get_value(value, Val(:{{name}}), Val({{conversion_unit}}), units))
 {{#create_docstring}}\"\"\"Get [`{{struct_name}}`](@ref) `{{name}}` as a unit-bearing quantity in the requested `units` (e.g. `SU`, `DU`, `MW`). For a bare number see [`{{accessor}}`](@ref).\"\"\"{{/create_docstring}}
 {{accessor}}_unitful(value::{{struct_name}}, units) = get_value(value, Val(:{{name}}), Val({{conversion_unit}}), units)
+{{accessor}}(value::{{struct_name}}) = _units_arg_required({{accessor}}, value, :{{name}}, Val({{conversion_unit}}))
+{{accessor}}_unitful(value::{{struct_name}}) = _units_arg_required({{accessor}}_unitful, value, :{{name}}, Val({{conversion_unit}}))
 InfrastructureSystems.display_units_arg(::typeof({{accessor}}), ::{{units_type_sig}}){{#units_bound}} where {T <: {{units_bound}}}{{/units_bound}} = InfrastructureSystems.{{display_units}}
 InfrastructureSystems.display_units_arg(::typeof({{accessor}}_unitful), ::{{units_type_sig}}){{#units_bound}} where {T <: {{units_bound}}}{{/units_bound}} = InfrastructureSystems.{{display_units}}
 {{/needs_conversion}}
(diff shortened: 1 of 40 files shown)


src/generate_structs.jl
```diff
@@ -95,6 +104,8 @@ InfrastructureSystems.display_units_arg(::typeof({{accessor}}_unitful), ::{{unit
 {{#needs_conversion}}
 {{#create_docstring}}\"\"\"Set [`{{struct_name}}`](@ref) `{{name}}`.\"\"\"{{/create_docstring}}
 {{setter}}(value::{{struct_name}}, val) = value.{{name}} = set_value(value, Val(:{{name}}), val, Val({{conversion_unit}}))
+{{setter}}(value::{{struct_name}}, val::Real) = _units_tag_required({{setter}}, value, :{{name}}, Val({{conversion_unit}}), val)
+{{setter}}(value::{{struct_name}}, val::NamedTuple{<:Any, <:Tuple{Vararg{Real}}}) = _units_tag_required({{setter}}, value, :{{name}}, Val({{conversion_unit}}), val)
```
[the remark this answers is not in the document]
luke-kiernan: Yeah this seems worth fixing

src/models/components.jl
```diff
@@ -354,6 +366,93 @@ _unit_category(::Val{:mva}) = APPARENT_POWER
...
+#
+# Every convertible field's accessors are generated in pairs: the working
+# `(component, units)` getter and setter, plus a fallback method that lands here
+# when the units are missing. The fallbacks exist purely so the failure names the
+# accessor, the field and the units that would have worked — Julia's own
+# `MethodError` lists signatures but never says what a valid unit argument is.
+#######################################################
+
+# Natural unit to suggest for each conversion-unit token.
+_natural_unit_example(::Val{:mw}) = "MW"
```
kdayday, line 381: This looks good for current logic. If we go in a pure unitful direction, these would need to be changed to unitful strings instead of the exported type names, correct?

luke-kiernan: yeah good point. I'll have AI add a test that the getters/setters actually accept the suggested syntax.

kdayday (approved):
Looks good, thanks! If we go full unitful direction, error message format will need to change away from `val * MW`, but that could be part of the larger change

PR #1783, opened 2026-09-02
thermal status enums

### LINKED ISSUE #1765: Update the status field in ThermalGeneration
We need to define a clear enum for ThermalGeneration Status since just the boolean won't allow to reflect the unit's availability to be commited for RUC or Services.

jd-lara:
This PR increases the vocabulary to handle other cases of commitment status in generators. This will allow to distinguish the cases where generators can be available to provide services while not commited. 

closes #1765

This is the companion to https://github.com/Sienna-Platform/SiennaSchemas/pull/38


src/definitions.jl
```diff
@@ -102,6 +102,65 @@ This classification is essential for WECC dynamic studies as it determines how l
...
+" OperationalStates
+
+function OperationalStatesModule.OperationalStates(value::Bool)
+    return throw(
+        ArgumentError(
+            "status is an OperationalStates value (OperationalStates.ONLINE or .OFFLINE), not a Bool; got $value",
+        ),
+    )
+end
+Base.convert(::Type{OperationalStates}, value::Bool) = OperationalStates(value)
```
rodrigomha, line 133: I think we should flip the OperationalStates to be OFFLINE = 0, and ONLINE = 1, STARTUP = 2, SHUTDOWN = 3. WIth that, a convert(OperationalStates, true) will result in ONLINE rather than STARTUP

src/models/generated/ThermalStandard.jl
```diff
@@ -47,11 +47,11 @@ This is a standard representation with options to include a minimum up time, min
 - `operation_cost::OperationalCost`: [`OperationalCost`](@ref) of generation
 - `base_power::Float64`: Base power of the unit (MVA) for [per unitization](@ref per_unit), validation range: `(0.0001, nothing)`
 - `time_limits::Union{Nothing, UpDown}`: (default: `nothing`) Minimum up and Minimum down time limits in minutes, validation range: `(0, nothing)`
-- `must_run::Bool`: (default: `false`) Set to `true` if the unit is must run
+- `commitment_mode::CommitmentModes`: (default: `CommitmentModes.COMMITTED`) Commitment mode of the unit. Options are listed [here](@ref commit_list)
```
rodrigomha, line 50: We are replacing the must_run with commitment state, and now the default is that every unit is Committed, rather than not a must run. I'm just wondering if we should leave must_run as a Status or not.

examples/create_system_example.jl
rodrigomha: Shouldn't we change this for completeness to be status = ONLINE. Same issue that before, the status true is STARTUP not ONLINE.

src/descriptors/power_system_structs.json
rodrigomha, line 1: Are we okay than non-thermal structs like HydroDispatch and Hybrids still use status Bool?

rodrigomha: It seems that HydroTurbine and HydroPumpTurbine carry the commitment_mode because they were allowed to be must_run, but HydroDispatch and Hybrids don't. 

The docs mention that the commitment_mode is only for Thermals so maybe we should remove entirely the must_run part of the Hydro units.

rodrigomha (changes requested):

docs/src/api/enumerated_types.md
```diff
@@ -101,6 +101,36 @@ EIA Annual Energy Review. `ThermalFuels` has the options:
 | `OTHER`                                                                                                                            | OTH           | Other type of fuel                                                                                                                  |
 | *Asterisk denotes fuel codes not directly from the current EIA 923 form but kept for compatibility with older versions of the form |               |                                                                                                                                     |
 
+## [Operational States](@id opstate_list)
+
+Each committable [`ThermalGen`](@ref) generator, as well as [`HydroDispatch`](@ref) and
+[`HybridSystem`](@ref), has a field for `status::OperationalStates`, the running on/off
```
jarry7: What about `HydroTurbine` and `HydroPumpTurbine`?

docs/src/api/enumerated_types.md
```diff
@@ -101,6 +101,36 @@ EIA Annual Energy Review. `ThermalFuels` has the options:
...
+`OperationalStates`. `CommitmentModes` has the options:
+
+| Name             | Description                                                                                        |
+|:---------------- |:-------------------------------------------------------------------------------------------------- |
+| `UNCOMMITTED`    | Not committed; the unit is offline but available                                                   |
+| `COMMITTED`      | Committed by the scheduling process (a planning model's commitment decision or a cleared schedule) |
+| `SELF_SCHEDULED` | Scheduled by its owner rather than by the scheduling process                                       |
+| `RELIABILITY`    | Committed by the system operator for reliability rather than by the scheduling process             |
+| `MUST_RUN`       | Required to run by contract or operating constraint                                                |
+
```
jarry7, line 139: This will break previous implementations where users depend on `must_run` (they use it intentionally).

How should this be managed going forward?

src/PowerSystems.jl
```diff
@@ -410,6 +410,8 @@ export DiscreteControlledBranchStatus
 export DiscreteControlledBranchType
 export PrimeMovers
 export ThermalFuels
+export OperationalStates
+export CommitmentModes
 export StorageTech
```
jarry7, line 416: Why do we need both `CommimentModes` and `OperationalStates`? (vs one or the other)

jarry7 (commented):
Left a few small comments and clarifications (do not have much more context than this specific PR).

PR #1784, opened 2026-09-03
Add CurveMultiHour on market bid costs and an identifier on outages

rodrigomha:
Two additive data-model changes for day-ahead market bids, each in its own commit, plus two small preparatory fixes and a wording pass.

## CurveMultiHour and curve_multihour on market bid costs (`59e7f3c9b`)
Day-ahead market bids carry two independent switches: the curve style is the quantity structure of a bid and the multi-hour block indicator is its time structure, and they compose. `CurveStyles` already covers the first; this adds the second.
- New `IS.@scoped_enum` `CurveMultiHour`: `SINGLE_HOUR = 0` (default) and `MULTI_HOUR = 1`.
- `MarketBidCost` and `MarketBidTimeSeriesCost` gain `curve_multihour` with getters, setters, and a kwarg default in every constructor.
- OpenAPI wire form is a plain integer `0/1`, the same convention as `curve_style`; the import converters reject any other value.
- No cross-field validation. CURVE + MULTI_HOUR is schema-permitted but not described by the protocol and stays allowed until checked against NP4-450. POM decides what it supports, as it does for `curve_style`.
- Docs: the enumerated-types page now lists `CurveStyles` and `CurveMultiHour`; `MarketBidTimeSeriesCost` is documented on the market bid cost page.

## Outage identifier (`43e735035`)
The three concrete outage attributes gain `identifier::Union{Nothing, String}`, default `nothing`, so a user can attach a name or a block id. `get_identifier` lives on the abstract `Outage` interface. The structs stay immutable, so the field is set at construction. Carried through the OpenAPI export and import.

## Preparatory fixes
- `e8a0cc536`: the time-series cost test helpers move to `test/common.jl`; `test_market_components.jl` borrowed them from `test_cost_functions.jl` and could not run on its own.
- `b1430c3d8`: `docs/Project.toml` still depended on the retired `PowerTimeSeriesOpenAPIModels`, so the docs environment could not resolve. It now mirrors PSY's Infrastructure Core and TimeSeries OpenAPI dependencies.

## Market-generic docs (`f5c962b21`)
The `CurveStyles` and `CurveMultiHour` docstrings and the enumerated-types page no longer name a specific market operator's bid fields. Sienna is shared across institutions, so docs stay generic.

## Companion PRs and merge order
Merge SiennaSchemas first, then PowerOpenAPIModels, then this PR. PSY's `[sources]` pin the OpenAPI model packages to `main`, so the new converter fields compile only once the PowerOpenAPIModels PR is on `main`.
- SiennaSchemas: https://github.com/Sienna-Platform/SiennaSchemas/pull/40
- PowerOpenAPIModels: https://github.com/Sienna-Platform/PowerOpenAPIModels/pull/13

## Verification
| Check | Result |
|---|---|
| Full test suite | 15505 passed, 0 failed, 46 files |
| Targeted files (market components, outages, OpenAPI document and export) | green, and `test_market_components` now passes in isolation |
| SiennaSchemas parity check | unchanged at the 124 pre-existing `power_units` drifts |
| Formatter | run |

The docs build still fails on psy6 for pre-existing reasons (11 tutorial and how-to code blocks, 455 missing docstrings, 41 unresolved references). This PR adds none of them and removes 22 missing docstrings and 4 unresolved references by documenting `MarketBidTimeSeriesCost`.

## Not included
- Python binding regeneration and the SiennaGridDB column mapping.
- PSB's `data/serialized_system` caches were cleared locally; downstream checkouts need the same after this merges.

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -6,7 +6,9 @@ Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 DocumenterInterLinks = "d12716ef-a0f6-4df4-a9f1-a5a34e75c656"
 DocumenterMermaid = "a078cd44-4d9c-4618-b545-3ab9d77f9177"
+InfrastructureCoreOpenAPIModels = "1f5e1c8d-e0cc-4dbf-8c6d-f2a3d2ae70a8"
 InfrastructureSystems = "2cd47ed4-ca9b-11e9-27f2-ab636a7671f1"
+InfrastructureTimeSeriesOpenAPIModels = "37a216c8-a490-47cf-89d7-db2cb9618199"
 Ipopt = "b6b21f68-93f8-5de0-b562-5493be1d77c9"
 JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
 JuMP = "4076af6c-e467-56ae-b986-b466b2749572"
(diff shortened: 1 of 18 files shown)


jd-lara (approved):
This looks good, we should wait until https://github.com/Sienna-Platform/SiennaSchemas/pull/40 is merged

Issue #946, opened 2023-01-04
Serialize all system files to one zip file


daniel-thom:
The `to_json(system)` serialization process produces 4 files. This can make it difficult to move around. It would be nice to include an option to package everything into one zip file. We would need to change the deserialization process to read from the zip file.

jd-lara:
The proposal will be to support an interface like this: 

`to_file(sys, format = :json, units = CU or NU)` that makes the classical JSON + HDF5 side car and it is the less performant way and an interface `to_file(sys, format = :sienna)` that stores the json for components, the sql tables and the hsd5 in a compressed file with the `*.sn` extension and by default writes to CU and it is the most performant way.

daniel-thom:
~~Could we make a new method for the zip format?~~

~~function save(sys; filename = "sys.zip", units = DEFAULT_UNITS, overwrite = false)~~

Update:
It seems that `save` may be a bad idea due to collisions with FileIO and existing Sienna conventions.

I generally find `overwrite` more clear than `force`, but we have used the latter for a long time.
```
to_file(sys, "case.zip"; units = CU, force = false)   # single-file archive
to_file(sys, "case"; units = CU, force = false)   # directory bundle
to_file(sys, "case.json"; units = CU, force = false)   # JSON file + HDF5 sidecar
```
This proposal drops the format kwarg in favor of auto-detecting the file extension. I would use the .zip file extension so that people know what it is and can extract themselves. Perhaps there is an argument to make it opaque. I would prefer making it obvious.

daniel-thom:
```
to_file(sys, "case.sn"; force = false)   # Create single-file archive
to_file(sys, "case"; units = CU, force = false)   # The directory `case`, must exist, create files in that directory.
to_file(sys, "case.json"; units = CU, force = false)   # Create JSON file + HDF5 sidecar at the given file path.
```

PR #1785, opened 2026-09-03
System file I/O, rebuilt around one format pair

### LINKED ISSUE #1768: `psy6`: add hint for failure to parse
We should add a hint such that if someone tries to do `sys = System("foo.raw")` or similar (matpower), they are told how to properly parse such a filetype. Definitely before release, but even now, such an addition would be greatly appreciated.

jd-lara:
closes #1768 and #946

`to_file` / `from_file` are the only way a `System` reaches or leaves disk. Adds a compressed single-file `.sn` archive beside the bundle directory, and deletes the legacy native JSON serializer rather than bridging it.

## The shape of the change

Two signatures changed and one whole path disappeared.

**Before**
```julia
# bundle directory, power-only keyword name
to_file(sys, dir; power_units = :component_base, force = false, pretty = false)
from_file(System, dir; system_kwargs...)

# plus a second, native path:
IS.to_json(sys, "sys.json")
System("sys.json")
```

**After**
```julia
# one path, two formats
to_file(sys, path; format = :json,               # or :sienna
                   unit_system = :component_base,
                   force = false, pretty = false)
from_file(path; system_kwargs...)

# the native path is gone; both entry points now throw DataFormatError
```

- `power_units` → `unit_system`: it never only governed power — the same conversion carries impedance (`:ohm`) and admittance (`:siemens`). `to_openapi` keeps its own `power_units`; different function.
- `from_file` drops the `::Type{System}` positional because the format is inferred from the path.

## Two formats, one serializer

`:sienna` is not a second writer. It builds the same bundle in a temp directory and archives it, so there is one serialization path and no way for the two to drift.

```
format = :json — a directory        format = :sienna — one file
                                    (entries at archive root; Tar.create
case/                                archives contents, not the directory)
├── system.json      the document
└── time_series.h5   the arrays     case.sn                    tar + gzip
                                    ├── system.json
                                    ├── sienna_extras.json     subsystems
                                    ├── time_series.h5
                                    └── time_series.h5.sqlite  the catalog
```

The extra members are the difference, and it is about **where the association tables live**:

| | associations | consequence |
|---|---|---|
| `:sienna` | InfraStore's `.sqlite`, authoritative on read | lossless; Sienna-only |
| `:json` | the document's `time_series_associations` | two files any non-Julia reader can consume |

`to_openapi(sys; write_catalog)` is the knob. Sidecars appear only when `sys` carries time series.

## How a write travels

```mermaid
flowchart LR
  subgraph W1["write · :json"]
    A1[to_file<br/>warn on loss] --> B1[_to_file_json] --> C1[to_openapi<br/>unit_system] --> D1[PD.write_document<br/>+ IS.serialize_arrays]
  end
  subgraph W2["write · :sienna"]
    A2[to_file<br/>refuse non-default units] --> B2[_to_file_sienna<br/>mktempdir] --> C2[_to_file_json<br/>write_catalog = true] --> D2[+ sienna_extras.json<br/>Tar + gzip]
  end
  subgraph R["read · either"]
    A3[from_file<br/>infer from path] --> B3[_from_file_sienna<br/>extract, if .sn] --> C3[_from_file_json<br/>resolve sidecar] --> D3[from_openapi<br/>replay or validate]
  end
```

`from_openapi` needs no format flag: it dispatches on whether the adopted store brought rows. Empty catalog → replay the document's rows; populated → validate against it, the path a `.sn`, an older bundle, and a PSB cache all take.

`:sienna` only ever writes `:component_base`; asking for `:natural_units` throws rather than paying the conversion the archive format exists to avoid.

---

**Depends on** Sienna-Platform/InfrastructureSystems.jl#632 → NatLabRockies/infrastore#71 → Sienna-Platform/SiennaSchemas (`jd/nonsequential_full_json`).

**Known break, out of scope:** PowerSystemCaseBuilder and PowerOperationsModels call the old signatures. Both fixes are the two mechanical edits this branch already applied 16 times inside PSY.


jd-lara:
The tests will break on PSB. I don't want to update across the board until we settle this and some of the other schemas changes proposed

jd-lara (commented):
This is still not performance optimized, needs a review on performance once we settle on the behavior

src/openapi/file_io.jl
```diff
@@ -20,12 +67,82 @@ rather than an empty HDF5 file that would imply the data went missing.
...
+        name => sort!(collect(get_component_ids(sys, name))) for
+        name in get_subsystems(sys)
+    )
+    open(joinpath(dir, SIENNA_EXTRAS_FILE), "w") do io
+        _print_extras(io, Dict("subsystems" => subsystems), pretty)
+    end
+    return nothing
+end
+
+"""Honour `to_file`'s `pretty` for this member too, so a bundle is not half indented."""
```
luke-kiernan: > Honour

Claude went british for a moment there

src/openapi/file_io.jl
```diff
@@ -46,43 +163,90 @@ end
...
+components need no such handling either way — they are re-masked on read when their owning
+`StaticInjectionSubsystem` is added.
 """
 function to_file(
     sys::System,
-    dir::AbstractString;
-    power_units::Symbol = :component_base,
+    path::AbstractString;
+    format::Symbol = :json,
+    unit_system::Symbol = :component_base,
```
luke-kiernan: > unit_system::Symbol

is there a reason why you went with a symbol, not an `AbstractUnitSystem` instance or subtype?

src/openapi/file_io.jl
```diff
@@ -103,17 +292,38 @@ _sidecar_path_for_write(::Val{true}, dir::AbstractString) = joinpath(dir, TIME_S
...
 
 The sidecar is located by the document's own `time_series_storage_file`, resolved relative to
-`dir` — so a bundle stays readable after being moved or renamed. A document that names a
-sidecar which is not present errors rather than yielding a system silently missing its time
-series.
+the bundle directory — so a bundle stays readable after being moved or renamed. A document that
+names a sidecar which is not present errors rather than yielding a system silently missing its
+time series.
+
+Of `System`'s keywords, `time_series_read_only` and `time_series_directory` are the two that
```
luke-kiernan, line 412: Is `time_series_read_only` mode actually "read only," or is it more like "open in-place?"

src/base.jl
```diff
@@ -249,82 +246,38 @@ function System(
...
+                "bundle directory or a $SIENNA_ARCHIVE_EXTENSION archive written by to_file — " *
+                "a single .json path is neither. A system serialized in the old format must " *
+                "be rebuilt from source and re-serialized with the current to_file.",
+            ),
+        )
+    elseif ext in (".raw", ".m")
+        throw(
+            DataFormatError(
+                "PowerSystems.jl no longer parses $ext files. Use PowerFlowFileParser.jl to " *
+                "parse this file into a System.",
```
luke-kiernan: We should provide documentation links or example syntax for how to turn a raw file into a `System` here. I recognize that those pieces are now outside of the scope of this package, but "look at PFFP" just doesn't cut it, for two reasons. First, it doesn't say where in PFFP's docs to look. Second, PFFP only provides raw-to-`PowerModelsData`. The user would still need to figure out how to turn a `PowerModelsData` dict into a `System`.

jd-lara: Well I think this should be an entry in the PFFP documentation but it doesn't exist yet.

src/openapi/file_io.jl
```diff
@@ -93,6 +257,31 @@ function to_file(
...
 
+function _to_file_sienna(
+    sys::System,
+    path::AbstractString;
+    force::Bool,
+    pretty::Bool,
+)
+    # `IS.create_sienna_archive` owns the container — the extension rule, the guards, and the
+    # compression. What is PSY's is only what goes inside it.
+    IS.create_sienna_archive(path; force = force) do bundle
```
daniel-thom, line 374: This creates a gzipped tar file. I think we should change it to create a zip file instead. That allows different compression settings per file. That is important because we don't want to fight with HDF5 compression - compress JSON, do not compress HDF5.

If you agree, we should make the change before people start using this feature.

docs/src/how_to/serialize_data.md
```diff
@@ -4,12 +4,18 @@
...
 significantly faster than reconstructing the `System` from raw data files.
 
-There are two formats:
+There are three forms, chosen by the extension of the path you pass to `to_file`:
 
-  - `format = :json` (default) — a directory holding `system.json` (an OpenAPI document) plus,
-    when the system has time series, `time_series.h5` and its `time_series.h5.sqlite` catalog.
-  - `format = :sienna` — the same three members, tar+gzip'd into one file (conventionally named
-    `*.sn`).
+  - **a directory** (no extension) — holds `system.json`, an OpenAPI document, plus
```
luke-kiernan, line 9: Default name and file extension for the OpenAPI document?

src/openapi/export_document.jl
```diff
@@ -165,22 +165,15 @@ function warn_unexportable_components(sys::System)
...
-        return NU
-    else
-        error(
-            "to_openapi(sys; power_units=$power_units): unmapped — expected " *
-            ":component_base or :natural_units",
-        )
-    end
+# ── export unit system ───────────────────────────────────────────────────────────
+
+"""Reject a unit system with no wire representation before any blob is written.
```
luke-kiernan, line 170: "wire representation?" "blob?"

luke-kiernan: I'm inferring that "wire" refers to the schema/gridDB side unit enums, which only allow for natural units and device base. If so, where did the "wire" term come from? I'd just call it "schema's unit enums" or similar.

src/openapi/file_io.jl
```diff
@@ -160,103 +182,187 @@ function _prepare_bundle_dir(dir::AbstractString, force::Bool)
...
+
+Both are complete: any `System` exports either way, whatever it was built from. `SU` is refused,
+having no representation in the wire enum. **An archive only ever writes `DU`** — that is the
+representation PSY stores natively, so it costs no conversion pass over every component, and
+the archive form is chosen specifically to be cheap to produce. Passing any other unit system
+with a `$(IS.SIENNA_ARCHIVE_EXTENSION)` path throws rather than silently ignoring the keyword
+or paying the conversion cost the form exists to avoid.
+
+Note the asymmetry between writing and reading. **A write is uniform**: PSY does not track a
+per-component basis, so the choice made here stamps every power-bearing blob in the document
```
luke-kiernan, line 228: Basis? Oh, AI just means "[choice of] unit system." But if that's what you mean, why not just say that...

luke-kiernan: I would also suggest tightening up this docstring. This is 80% explanation/justification/elaboration on alternatives, which imo does not belong in a docstring. E.g. if I were to cut aggressively:

> ~Both are complete: any `System` exports either way, whatever it was built from.~ `SU` is refused,
> ~having no representation in the wire enum.~ **An archive only ever writes `DU`** — ~that is the
> representation PSY stores natively, so it costs no conversion pass over every component, and
> the archive form is chosen specifically to be cheap to produce.~ Passing any other unit system
> with a `$(IS.SIENNA_ARCHIVE_EXTENSION)` path throws ~rather than silently ignoring the keyword
> or paying the conversion cost the form exists to avoid.~

luke-kiernan (commented):
Did a quick pass on "changes since last review"

src/openapi/file_io.jl
jd-lara: The branch in IS is now merged

.claude/CLAUDE.md
```diff
@@ -4,6 +4,147 @@ The Sienna power-system **data model**: the `System` container plus ~210 compone
 
 **This branch has NO parsers.** `src/parsers/` was removed in the psy6 line; all Matpower/PSSE/table parsing lives in PowerFlowFileParser.jl (and PSB's parser wrappers). Do not re-add parsing here.
 
+**There is also NO native JSON serializer.** `to_file`/`from_file` are the only way a `System` reaches or leaves disk — see "System file I/O" below before touching anything serialization-shaped.
+
+## System file I/O — `to_file` / `from_file` only
```
daniel-thom, line 9: This feels like an excessive amount of data to load into context every time.

docs/src/how_to/serialize_data.md
```diff
@@ -1,112 +1,124 @@
...
-dataset from
-[`PowerSystemCaseBuilder.jl`](https://github.com/Sienna-Platform/PowerSystemCaseBuilder.jl),
-simply to illustrate the process.
+The two document forms are readable by any client that can read the OpenAPI schema; the archive
+is Sienna-only, because reading it means reading InfraStore's catalog.
 
-First, load the dependencies and a `System` from `PowerSystemCaseBuilder`:
+!!! warning
+
+    There is no migration path from the old single-file native JSON format (written by the
```
daniel-thom, line 22: Do we need this?

docs/src/how_to/serialize_data.md
```diff
@@ -1,112 +1,124 @@
...
+    reactive_power_limits = (min = -1.0, max = 1.0),
+    ramp_limits = nothing, operation_cost = ThermalGenerationCost(nothing),
+    base_power = 100.0,
+)
+add_component!(sys, gen)
 ```
 
-Set up your target path, for example in a "mysystems" subfolder:
+`to_file` picks the form from the extension of the path you give it — there is no `format`
+keyword. Write it as a directory:
```
daniel-thom, line 59:
```suggestion
`to_file` picks the form from the extension of the path you give it. Write it as a directory:
```

src/openapi/file_io.jl
```diff
@@ -1,17 +1,78 @@
...
+#   - `case.json`    the same two members, named after the document and sitting beside it
+#
+#                        case.json          the OpenAPI document
+#                        case.h5            the InfraStore arrays
+#
+#                    The sidecar takes the document's stem so several systems can share one
+#                    directory; the document records only its basename, so the pair moves
+#                    together.
+#
+#   - `case.sn`      four members, tar+gzip'd into one file
```
daniel-thom, line 19:
```suggestion
#   - `case.sn`      four members, zip'd into one file
```

src/openapi/file_io.jl
```diff
@@ -1,17 +1,78 @@
...
+#
+#                        case.sn
+#                          system.json
+#                          time_series.h5
+#                          time_series.h5.sqlite   InfraStore's own catalog
+#                          sienna_extras.json
+#
+#                    Entries sit at the archive root rather than under a `case/` prefix,
+#                    because `Tar.create` archives a directory's contents, not the
```
daniel-thom, line 28: It's now Zip instead of Tar + Gzip.

src/data_format_conversions.jl
```diff
@@ -1,238 +0,0 @@
-# This file contains code to convert serialized data from old formats to work with newer
```
daniel-thom, line 1: What is our plan for handling future migrations? I'm assuming that the OpenAPI JSON documents will include versions, so we should be able to handle them.

test/Project.toml
```diff
@@ -38,11 +38,8 @@ InfraStore = "6aee0376-896a-4a59-96ff-12f221c99746"
 
 [sources]
 PowerSystems = {path = ".."}
-# IS4 carries the association-id time series keys; InfraStore 0.11 is registered, so it
-# needs no source pin.
-# LOCAL PATH CO-DEV PIN (temporary): see ../Project.toml — IS4's local checkout already
-# migrated off PowerCoreOpenAPIModels/PowerTimeSeriesOpenAPIModels, uncommitted; the pushed
-# IS4 branch fails to resolve. Revert to the git rev once IS4 is pushed.
+# IS4 carries the association-id time series keys and the arrays-only sidecar (IS #632,
+# merged); InfraStore 0.12 is registered, so it needs no source pin.
```
daniel-thom, line 42: Delete

daniel-thom (changes requested):

daniel-thom:
@jd-lara Is there some mechanism of recording the source version in the OpenAPI JSON file or zip, like in a manifest file? I did not see one. At minimum, we should be able to detect a major version upgrade and report warnings.

jd-lara:
> @jd-lara Is there some mechanism of recording the source version in the OpenAPI JSON file or zip, like in a manifest file? I did not see one. At minimum, we should be able to detect a major version upgrade and report warnings.

This is a good point. I'll how to incorporate that. I think that I'll finish this work once the schema catch up and the update to OpenAPI.jl 1.x is complete. Maybe we need to add something to the document about the schema version.

docs/src/how_to/serialize_data.md
jd-lara: ELiminated

.claude/CLAUDE.md
jd-lara: I removed some already

jd-lara:
This PR requires https://github.com/Sienna-Platform/PowerSystemCaseBuilder.jl/pull/224

.claude/CLAUDE.md
```diff
@@ -12,6 +77,25 @@ PNM, PF, POM, and PSB all consume PSY; SiennaSchemas mirrors PSY component field
 2. **clear PSB's `data/serialized_system/` cache** — it has no version-aware invalidation, and stale cached systems produce confusing deserialization failures downstream.
 3. if the change touched a component field also present in SiennaSchemas, flag the schema counterpart (example of real drift: `head_to_volume_factor` moved to `FunctionData` in PSY commit `ed30a682` while `SiennaSchemas/Operations/StaticInjection/HydroReservoir.json` still `$ref`s `ValueCurve`).
 
+**Open cross-repo break (as of the `to_file`/`from_file` cutover):** two packages still call the
```
daniel-thom: Does this belong? Perhaps we need a review of this file.

daniel-thom (commented):
@jd-lara Can you describe how versioning works? I don't see where a System schema version is written to the JSON. Did I miss it? Secondly, what is the plan for handling upgrades? Suppose there is a 6.1 or 7.0 that needs to modify something during de-serialization. Are we saying that we will need to keep PowerSystemsUpdater.jl in sync in perpetuity?

jd-lara:
> @jd-lara Can you describe how versioning works? I don't see where a System schema version is written to the JSON. Did I miss it? Secondly, what is the plan for handling upgrades? Suppose there is a 6.1 or 7.0 that needs to modify something during de-serialization. Are we saying that we will need to keep PowerSystemsUpdater.jl in sync in perpetuity?

I opened an issue to modify the schemas to add the versioning https://github.com/Sienna-Platform/SiennaSchemas/issues/61 . This is a follow up PR since I need to design this capability, for now I want to be able to merge the machinery to be able to restore the results processing in PSI with the system data . 

I don't have a plan now for the potential update from 6 to 7. I guess that in the same vein as with the semantic versioning we need to maintain backwards compatibility for all PSY 6 versions and the updater needs to support going from 5 to 6 for all 6 versions.

.claude/CLAUDE.md
jd-lara: I'll remove it

daniel-thom (approved):

Issue #1788, opened 2026-09-08
SwitchedAdmittance: remove or rename `Y` (follow-up to #1774)


luke-kiernan:
Follow-up to #1774, which covered items 1 and 2 (add `solved_admittance`, rename `initial_status` → `number_engaged`). Item 3 — removing `Y` — was deferred there because the issue itself flagged it as needing investigation, and the field survived that change untouched.

From #1774:

> **Remove Y**. It's basically a fixed admittance sitting inside the switched admittance. This one requires a bit more investigation: are there test cases and data models that need it? [PSS(E)'s data model does not.] At the very least, I'd be in favor of giving the field a more descriptive name.

## What needs answering first

- Which test cases and serialized systems actually carry a non-zero `Y` on a `SwitchedAdmittance`? If the answer is "none," removal is clean; if some do, they need a migration story (a co-located `FixedAdmittance`, most likely).
- PSS/E has no equivalent, so nothing on the parser side should depend on it — worth confirming against PowerFlowFileParser.jl.
- Consumers that compute total admittance as `Y` + `number_engaged` ⋅ `Y_increase` need updating either way.

## If removal turns out to be too disruptive

The fallback position from #1774 is renaming it to something descriptive (`base_admittance` / `fixed_admittance`) rather than the bare `Y`.

## Blast radius

Same cross-repo path as #1774, which is a worked example of the propagation: PSY descriptor + regenerate + hand-written OpenAPI converters → SiennaSchemas (`Operations/StaticInjection/SwitchedAdmittance.json`, verified with `scripts/check_psy_parity.py --psy-path ../PowerSystems.jl`) → PowerOpenAPIModels regeneration → SiennaGridDB (`y_g`/`y_b` columns in `switched_admittance` plus their `column_conventions.json` rows, which exist only because of this field).

Note that GridDB decomposes `Y` into real `y_g`/`y_b` columns, so removing it there is a schema change, not just a registry edit.

Issue #1787, opened 2026-09-08
Regenerating structs breaks precompilation: codegen's DU from_openapi fallback collides with the hand-written NU fallback for market components


luke-kiernan:
Regenerating structs currently produces a package that **fails to precompile**, for reasons unrelated to whatever field the regeneration was actually for. Anyone who follows the documented generated-code workflow hits this.

## Symptom

```
WARNING: Method definition from_openapi(PowerOperationsOpenAPIModels.TradingHub, PowerSystems.OpenAPIRefs)
  in module PowerSystems at src/models/generated/TradingHub.jl:78
  overwritten at src/openapi/import_handwritten.jl:55.
ERROR: Method overwriting is not permitted during Module precompilation.
```

## Cause

`src/generate_structs.jl:143-147` emits a two-argument `from_openapi` fallback for **every** struct that has no `power_units` field:

```julia
{{^has_power_units}}
function from_openapi(po::{{{openapi_po_type}}}, refs::OpenAPIRefs)
    return from_openapi(po, refs, DU)
end
{{/has_power_units}}
```

But the three market components are deliberately hand-declared with the **opposite** default at `src/openapi/import_handwritten.jl:54-56`:

```julia
for T in (:TradingHub, :VirtualParticipant, :PointToPointBid)
    @eval from_openapi(po::PO.$T, refs::OpenAPIRefs) = from_openapi(po, refs, NU)
end
```

So codegen emits a `DU` fallback, the hand-written layer emits an `NU` fallback for the same signature, and the two collide.

This is distinct from the ten types in the `DU` loop at `import_handwritten.jl:50` (`Arc`, `TwoWindingTransformer`, …) — those have fully hand-written converters, so codegen emits nothing for them and there is no clash. The market components are the only types that get *generated* three-arg converters plus a *hand-written* two-arg dispatcher that disagrees with the generated one.

## Why it is currently invisible

`src/models/generated/TradingHub.jl`, `VirtualParticipant.jl`, and `PointToPointBid.jl` are committed **without** the fallback the generator now emits, so `main` builds fine. The staleness is load-bearing: the first regeneration reintroduces the methods and breaks the build. Nothing checks descriptor↔generated consistency, so CI does not catch it either.

## Repro

```sh
julia --project=test -e "using PowerSystems; PowerSystems.StructGeneration.generate_structs(\"./src/descriptors/power_system_structs.json\", \"./src/models/generated\")"
[...]

Issue #1774, opened 2026-08-25
`SwitchedAdmittance` struct changes


luke-kiernan:
There's been a couple rounds of attempts to get the switched shunt behavior and parsing correct, but they're still causing massive discrepancies in power flow cases.  From discussion with @m-bossart, here's the changes we're suggesting to the struct:

1. **New field: `solved_admittance::Union{Float64, Nothing}`**. This is the equivalent of PSS(E)'s `BINIT`. Per the manual: "`BINIT` needs to be set to its actual solved case value only when the network...is to be considered solved as read in, or when the device is to be treated as locked [`MODSW = 0`]." If this is non-nothing, then we use this value as the switched shunt's effective admittance when running power flows, instead of the usual `Y + dot(initial_status, y_increase)` expression. After solving a power flow, we write the solved-for admittance to this field.
2. **Rename field: `initial_status` -> `number_engaged`**. That's how we use it in PowerFlows.jl: we write the solved-for step count to that field. Change the docstring appropriately. We continue parse PSSE's `Si` as this field. 
3. **Remove Y**. It's basically a fixed admittance sitting inside the switched admittance. This one requires a bit more investigation: are there test cases and data models that need it? [PSS(E)'s data model does not.] At the very least, I'd be in favor of giving the field a more descriptive name.

Context: PFFP [#48](https://github.com/Sienna-Platform/PowerFlowFileParser.jl/issues/48).There's also a couple past PRs/issues like #1679 that papered over the issue without actually fixing it.

jd-lara:
I am supportive of these changes we should call it base_admittance and make single field between the Y and B_init. I am sure PSSE has some weird default  when BInit = 0 using the min/max values

PR #1789, opened 2026-09-08
Rename initial_status to number_engaged, add solved_admittance

luke-kiernan:
Closes part of #1774 — items 1 and 2. Item 3 (removing `Y`) is deferred to #1788.

**Depends on Sienna-Platform/SiennaSchemas#44 and a PowerOpenAPIModels regeneration.** `PO.SwitchedAdmittance` still has `initial_status` and no `solved_admittance`, so OpenAPI import/export of a `SwitchedAdmittance` will fail at runtime until those land. PSY still *compiles* — the PO field access is dynamic — and no current test exercises that round trip, which is why CI here stays green either way. Merge order: schemas → PowerOpenAPIModels → this.

## `initial_status` → `number_engaged`

The name was wrong. PowerFlows.jl writes the solved-for step count into this field, so it holds a per-block count of engaged steps, not the binary in/out-of-service flag its docstring claimed. Renamed, with a docstring saying what it actually holds and that it is bounded by `number_of_steps`. Still parses from PSS/E `Si`.

## New: `solved_admittance::Union{Nothing, Float64}`

PSS/E `BINIT`. When non-`nothing` it is the shunt's effective admittance, used in place of `Y` + `number_engaged` ⋅ `Y_increase`, and power flow writes the solved-for admittance back to it. Per the PSS/E manual it should be set only when the case is to be treated as solved as read in, or when the device is locked.

It is the same COMPONENT_MVAR-on-system-base quantity as `Y`, so the hand-written converters in `import_handwritten.jl` / `export_handwritten.jl` give it the same `base_power` division/multiplication; `nothing` passes through unscaled.

## Two things worth a reviewer's attention

**The generated diff is deliberately narrow.** Regeneration also wanted to add a two-argument `from_openapi` fallback to `TradingHub.jl`, `VirtualParticipant.jl`, and `PointToPointBid.jl`. That fallback dispatches to `DU`, but those three are hand-declared with `NU` at `import_handwritten.jl:54-56`, so the two collide and **precompilation fails outright**. I reverted those three files; the generator bug is filed separately as #1787. Their committed staleness is currently load-bearing.

**Descriptor edits were surgical.** Rewriting `power_system_structs.json` through a JSON round trip reformats the whole file, so the edits are text replacements instead — the descriptor diff is 10 lines.

## Verification

- Full suite: **15474/15474**, zero failures or errors.
- `docs/make.jl`: clean.
- Formatter: clean, no changes.
- Cross-repo: `check_psy_parity.py --psy-path .` from the paired schemas branch reports **zero `SwitchedAdmittance` drift`.
- New coverage in `test_devices.jl` for both fields — defaults, construction, and the write-back setters power flow relies on.

## Not covered here

SiennaGridDB still references `initial_status` in `schema/coverage_decisions.json`, and `solved_admittance` may warrant a column plus a `column_conventions.json` registry row. Follow-up.

https://claude.ai/code/session_01LP6eB1zx4eyE3hd7tpvSue


src/descriptors/power_system_structs.json
```diff
@@ -2969,7 +2969,7 @@
         },
         {
             "struct_name": "SwitchedAdmittance",
-            "docstring": "A switched admittance, with discrete steps to adjust the admittance.\n\nMost often used in power flow studies, iterating over the steps to see impacts of admittance on the results. Total admittance is calculated as: `Y` + `number_of_steps` * `Y_increase`",
+            "docstring": "A switched admittance, with discrete steps to adjust the admittance.\n\nMost often used in power flow studies, iterating over the steps to see impacts of admittance on the results. Total admittance is calculated as: `Y` + `number_engaged` * `Y_increase`, unless `solved_admittance` is set, in which case that value is the effective admittance",
```
m-bossart: Didn't we discuss removing `Y` field. It seems to me that with the addition of `solved_admittance` we don't need `Y`. If there is actually an admittance that doesn't correspond to the switched blocks, it can be a separate fixed admittance

luke-kiernan: We did. AI made me question if that might break something. But even if it did, I guess it'd still be an open question of whether that indicates a bug in our code or bad input data. (PF's "adjust discrete control devices" stuff isn't _that_ well-tested...)

m-bossart: I think we should try to replace Y with solved_admittance and see if anything breaks.

jd-lara (approved):

PR #1793, opened 2026-09-11
Restore the test suite on the regenerated OpenAPI packages

jd-lara:
Restores the test suite on the regenerated OpenAPI packages (OpenAPI.jl 1.1, SiennaSchemas `b49f2b8`).

## What changed

- **Generated converters** (`src/generate_structs.jl`, regenerated `src/models/generated/`): an omitted wire value is `ABSENT`, distinct from an explicit `null`; every optional-by-omission field is guarded and falls back to the descriptor default (or `nothing` for a nullable field). Cost fields typed as a `oneOf` (`OperationalCost`, a `Union` of cost structs) unwrap the generated wrapper; a field typed as one concrete cost struct does not.
- **Export**: `vom_cost` is required on `CostCurve`/`FuelCurve` and always emitted (only `startup_fuel_offtake` stays optional); enum fields go out as their wrapper types; `power_units` as `IC.UnitSystem`; `LossCurve` replaces `TwoTerminalLoss` on the HVDC converters; `ChargeDischarge` replaces the per-site storage start-up object; `AverageRateCurve` wraps its function data in its own wrapper (a copy-paste of the incremental one, found by the round-trip tests).
- **Import**: `Absent` handled alongside `nothing` by dispatch; enum wrappers read through `.value`; the service-offer fill-in rebuilds immutable cost structs instead of mutating them.
- **Environments**: `test/Project.toml` and `docs/Project.toml` resolve PowerSystemCaseBuilder, PowerTableDataParser and PowerFlowFileParser from the sibling checkouts (the previous PSB pin pointed at a branch still shipping the dev8 test data), until the OpenAPI packages register. CI on remote pins is expected to fail until then.
- **Tests**: constructions of generated types carry the now-required fields and wrapper types; assertions compare `.value` on enum wrappers and `ABSENT` on omitted fields; three tests that asserted "a required field can be omitted and the converter errors" now assert the wire type's constructor itself refuses the omission, since the schema no longer allows the absence.

## Known

`docs/make.jl` has zero `missing_docs` but still stops on content outside this change: a tutorial whose fixture is a Matpower `.m` file this branch has no parser for, one tutorial written against an older InfrastructureSystems API, and cross-references into InfrastructureSystems docstrings that do not exist yet.

## Test plan

- [x] `julia --project=test test/runtests.jl` — 15561 passed, 0 failed, 0 errored

https://claude.ai/code/session_01QDkZStcFzebhh9XFpPKRSo


.github/workflows/cross-package-test.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        julia-version: [1]
+        julia-version: ['1.12']
```
jd-lara, line 15: @daniel-thom We need to pin this because of the issue in infrastore with Julia 1.13

jd-lara (commented):
Claude created a template with a lot of `isa` calls. This is not acceptable, this PR works but need type stability review

jd-lara:
This PR is ready to restore all the functionaility to psy6

PR #1796, opened 2026-09-14
Treat ramp limits as a power-per-time quantity

luke-kiernan:
Builds on `jd/openapi-1x`, per the instruction to base struct-generation changes there.

## Why

`ramp_limits` is documented everywhere as MW/min, but all six component types that carry it declared `conversion_unit: ":mw"` — plain active power. The stored arithmetic was right (the per-minute basis is a pass-through constant that `base_power` scaling never touches), so no number on the wire changes. Only the labels were wrong, in both directions:

- `get_ramp_limits(g, u"MW")` returned a per-minute rate labeled MW.
- `get_ramp_limits(g, u"MW/minute")` — the one correct spelling — threw a `DimensionError`.

The schemas and the GridDB column registry already declared this field as `ActivePowerChangeRate` in MW/min, and `Core/units.json` already carried the quantity. **PSY was the layer lagging behind**; this brings it into line.

## What

Two commits: the first unifies unit categories by their base exponents, the second builds the rate category on top.

- **Rate category.** Since `UnitCategory` describes a quantity by its base exponents, a rate needs no new machinery: its natural unit carries the time dimension, and its per-unit bases are the wrapped category's, because time has no base. `storage_time` names the unit the stored number is denominated in. Unitful does the power and time conversion together, so `u"kW/s"` works without enumerating it.
- **A relative marker needs a time.** `CU`/`SU` alone are rejected on a rate field; `CU/u"minute"`, `SU/u"hr"` and natural units like `u"MW/minute"` take their place. The rejection names the field, and an unset field still returns `nothing` like every other getter.
- **`RateUnit`** is the target-unit argument. A tagged rate *value* needs no new type — `RelativeQuantity <: Number` composes with Unitful's own `/(::Number, ::Units)`.
- Storage stays per-minute on the component base, so there is no data migration and no fixture renumbering.

## Breaking

Consumers must name the time: `get_ramp_limits(g, CU/u"minute")` rather than `get_ramp_limits(g, CU)`. Every relative-marker call site fails loudly with a message naming the valid spellings.

Known downstream call sites: PowerSystemCaseBuilder (1, PR to follow), PowerOperationsModels (5), PowerSimulations (1). All mechanical.

## Verification

- Full suite 15721/15721, exit 0.
- Regeneration idempotent on this base; the generated diff is confined to the six ramp fields.
- Formatter clean.
- Docs build: both ramp `@setup` failures cleared against a patched PSB. The remaining docs failures are pre-existing and unrelated (the removed `.m` parser in `parse_ts_from_csvs.md`, an `ignore_scaling_factors` MethodError in the time-series tutorial, and IS cross-reference resolution).

## Not included

`scripts/check_psy_parity.py` in SiennaSchemas is what lumped `ActivePowerChangeRate` under `":mva"` and hid this drift. Its family table is stale well beyond this field — it keys only `:mva`/`:ohm`/`:siemens` while the descriptor uses `:mw` (75), `:mvar` (60), `:mva` (28) — and it doesn't understand the discriminated `x-units` form at all, reporting ~127 drifts on a clean tree. That wants a rewrite, not a patch, so it is left for a separate issue.

https://claude.ai/code/session_01MJK3vJgGKX7vG1ZsiSgzSF


src/units/conversions.jl
```diff
@@ -249,6 +335,124 @@ end
...
+# A value expressed per `FROM` becomes `value * _time_factor(FROM, TO)` per `TO`
+# (1/minute → 1/hr multiplies by hr/minute = 60). Both units are singleton types, so
+# this folds to a literal.
+_time_factor(::FROM, ::TO) where {FROM <: Unitful.Units, TO <: Unitful.Units} =
+    Unitful.ustrip(Unitful.uconvert(Unitful.NoUnits, 1.0 * (TO() / FROM())))
+
+# Re-base a bare relative value between CU and SU. Written out rather than delegated to
+# `convert_units`, because for a rate category that would land on the bare-marker
+# rejection below: the rejection is about an incomplete *target*, not about the power
+# arithmetic, which is the same as for any other category.
```
luke-kiernan: I don't follow the reasoning here

src/units/conversions.jl
```diff
@@ -64,6 +99,62 @@ const IMPEDANCE = ImpedanceCategory()
...
+"""
+    storage_time(category) → Unitful.Units
+
+The time unit a rate category's stored per-unit value is denominated in: a stored
+`ramp_limits` of 0.1 means 0.1 pu **per minute**.
+
+Declared per rate category rather than recovered from `natural_unit`, which would mean
+picking the time factor out of a compound `FreeUnits`' internals.
+"""
+storage_time(::ActivePowerChangeRateCategory) = u"minute"
```
luke-kiernan: Poor name choice: "storage" overlaps with "battery storage." 

It also looks kind of strange to have a function with a single definition...but I plan to do some similar stuff for cost curves and loss curves, at which point we'll have other rates, `$/MW hr`.

src/units/conversions.jl
```diff
@@ -64,6 +99,62 @@ const IMPEDANCE = ImpedanceCategory()
...
+
+# Categories compose, so a derived quantity's base and natural unit follow from its
+# parts rather than being declared: `VOLTAGE^Val(2) / POWER` *is* `IMPEDANCE`. The
+# exponents add and Unitful composes the natural unit, both at the type level.
+Base.:*(
+    ::UnitCategory{N1, P1, V1},
+    ::UnitCategory{N2, P2, V2},
+) where {N1, P1, V1, N2, P2, V2} =
+    UnitCategory{typeof(N1() * N2()), P1 + P2, V1 + V2}()
```
luke-kiernan, line 124: Could be`@generated`--check if that makes compilation difference

src/units/conversions.jl
```diff
@@ -74,16 +165,20 @@ const CURRENT = CurrentCategory()
...
+# errors when the base voltage is unset, and a quantity with `V == 0` must not care.
+# One method per exponent that actually occurs, so `^` (a libm `pow` call) never
+# appears in the generated code -- the same reason `_convert_curve_axes` in
+# InfrastructureSystems is written one method per `Val`.
+@inline _base_factor(::Val{0}, getter, c) = 1.0
+@inline _base_factor(::Val{1}, getter, c) = getter(c)
+@inline _base_factor(::Val{-1}, getter, c) = inv(getter(c))
+@inline _base_factor(::Val{2}, getter, c) = (v = getter(c); v * v)
+@inline _base_factor(::Val{-2}, getter, c) = (v = getter(c); inv(v * v))
+@inline _base_factor(::Val{N}, getter, c) where {N} = getter(c)^N
```
luke-kiernan, line 176: Check if there's an advantage to these special cases--the `Val{N}` case might handle the `Val{1}` and `Val{2}` cases just fine.

src/units/conversions.jl
```diff
@@ -249,6 +335,124 @@ end
...
+    v = _relative_rebase(c, value, cat, from, base)
+    return RelativeQuantity(v * _time_factor(storage_time(cat), time_unit(to)), base) /
+           time_unit(to)
+end
+
+# A tagged rate value → anywhere: drop to the category's storage time, then re-enter the
+# engine as an ordinary relative value. Mirrors the `RelativeQuantity` unwrap guard below.
+# A `RelativeRate`'s Unitful parameter is the *inverse* time unit (`minute^-1`), since
+# the quantity is a per-time one; the time unit itself is its reciprocal.
+_rate_time(::RelativeRate{T, U, D, TU}) where {T, U, D, TU} = inv(TU())
```
luke-kiernan: Confusing name choice--does this return something with _time_ dimensions (`hr`) or with _rate_ dimensions(`1/hr`)?

src/units/conversions.jl
```diff
@@ -249,6 +335,124 @@ end
...
+
+# `from` is matched as `AbstractRelativeUnit` and the tag checked in the body rather
+# than tied to `U` in the signature: tying it leaves these ambiguous against the
+# "Unitful value carries a relative `from`" guard below, which they must beat.
```
luke-kiernan, line 377: The `_time_factor`, `_rate_time`, and `storage_time` triple seems over-engineered.

jd-lara (commented):
Do we need a schema PR too?

jd-lara (approved):

PR #1798, opened 2026-09-16
Follow IS's write-through time series adds in the OpenAPI tests

luke-kiernan:
`IS.make_add_batch` / `IS.commit_batch!` were deleted in IS; update the OpenAPI test helpers to use `add_time_series!(store, owner_id, owner_type, owner_category, ts)` instead. All three sites are single adds, so none needs a transaction.

The `assoc_id` / `ts_key` lookups are left as they are. `add_time_series!` hands back the key and could shortcut both, but the catalog round-trip is what the second site is testing.

CI here will fail with errors in PSB until NLR-Sienna/PowerTableDataParser.jl#10 is merged

Verified: `test_openapi_document.jl` and `test_openapi_cost_conversion.jl` against current IS4, 601 pass / 0 fail.


luke-kiernan:
This is straightforward, only a dozen lines of changes (all in the tests), so I'll just merge it

Issue #1730, opened 2026-07-22
Start-up/Shut-down ramp and time limits


jd-lara:
More broadly, there is a question on how to represent the start up and shut down time parameters for thermal units. 

Open questions: Should we have this in all thermals or put the ThermalMultistart only?

PR #1801, opened 2026-09-18
Add switching_times to the thermal generators

annacasavant:
ThermalStandard and ThermalMultiStart gain an optional switching_times field, an UpDown in minutes: `up` is the time the unit takes to reach ONLINE after a start is initiated, `down` the time to reach OFFLINE after a shut-down. It defaults to nothing on both types, so existing callers and serialized systems load unchanged, and a value is validated non-negative.

The field sits after dynamic_injector on both structs, among the defaulted trailing fields, so the positional constructors keep their existing argument order for every caller that stops at dynamic_injector.

Mirrored in SiennaSchemas as an UpDown $ref with x-unit min on both thermal schemas, and stored in the relational mirror as a nullable JSON {up, down} column on thermal_generators with an OperationalDuration/min unit convention. check_psy_parity.py reports no drift.


rodrigomha:
Btw this closes #1730

PR #1802, opened 2026-09-18
Add currency units and component-aware curve rebasing

luke-kiernan:
Targets `psy6`. Builds on the `UnitCategory{NU, P, V}` engine already on that branch.

## Currency and cost categories

`@dimension 𝐂` + `@refunit USD` — currency is a base dimension of its own, so unlike `MVAr`/`MVA` there is no existing Unitful unit to scale from. `CostPerTimeCategory` has both base exponents zero: currency has no base and there is no time base, so a cost rate per-unitizes nothing and its `base_value` is 1.0 for every component.

Curve coefficient categories are **derived, not declared**. For `y = sum a_n x^n` the category of `a_n` is `Y / X^Val(n)`, which the existing category operators already compose:

| coefficient | natural unit | base |
|---|---|---|
| a₀ | `USD hr⁻¹` | 1 |
| a₁ | `USD hr⁻¹ MW⁻¹` | 1/S |
| a₂ | `USD hr⁻¹ MW⁻²` | 1/S² |

So a cost curve of any degree needs no new constant. A `LossCurve` slope comes out `ACTIVE_POWER / ACTIVE_POWER^Val(1)` — dimensionless, base-invariant.

## Component-aware curve rebasing

`IS.convert_power_units(curve, to, ratio)` takes the x-axis base ratio as a plain number because InfrastructureSystems has no component to derive it from. Its docstring names PowerSystems as the layer that resolves it "per component and physical category". That layer was never written, so the function had **no callers anywhere** — IS's own tests spell the ratio out by hand with a local `_x_base` helper.

This adds it:

```julia
function IS.convert_power_units(c::Component, curve::IS.ValueCurveWithUnits, to::IS.AbstractUnitSystem)
    from = IS.get_power_units(curve)
    from === to && return curve
    ratio = _x_axis_base_value(c, to) / _x_axis_base_value(c, from)
    return IS.convert_power_units(curve, to, ratio)
end
```

plus `get_<field>(component, to)` over all sixteen curve-bearing fields, generated by two `@eval` loops over explicit, commented name lists:

- **Through `operation_cost` (8):** `variable_operation_cost`, `curtailment_cost`, `charge_variable_cost`, `discharge_variable_cost`, `incremental_offer_curves`, `decremental_offer_curves`, `import_offer_curves`, `export_offer_curves`
- **Fields of the component itself (4 names, 4 types):** `loss_function` (`InterconnectingConverter`), `loss` (`TwoTerminalGenericHVDCLine`, `TwoTerminalLCCLine`), `converter_loss_from`/`converter_loss_to` (`TwoTerminalVSCLine`)

The no-units getters are untouched.

### Identity conversion

`from === to && return curve` is not just an optimization. Without it the identity case routes through `scale_x`, which rejects time-series-backed curves outright — and `MarketBidTimeSeriesCost` and `ImportExportTimeSeriesCost` are TS-backed by construction, so asking for the basis a curve already had was a reachable throw. A genuine change of basis on a TS curve still fails, correctly, from IS's guard.

### No setters

A curve carries its power basis in its own `U` type parameter. Unlike a scalar field, which is stored canonically in CU and so has a basis to convert *to* on write, a curve is self-describing and a unitful setter would have nothing to do.

## Testing

Four new testsets in `test/test_units.jl`, 27 assertions, covering the coefficient table against a real 50 MVA `ThermalStandard`, CostCurve NU↔CU↔SU round trips, a `LossCurve` on a converter (the y-axis rebasing path cost curves never exercise), storage's two curves, and the identity case. `Aqua.test_ambiguities` and `test_undefined_exports` pass — ambiguity was the real risk in adding generic `::Component` methods alongside the per-type generated accessors.

Those runs were made while the branch was based on `lk/enumx`, where the units files and `src/models/components.jl` are byte-identical to `psy6`. After rebasing onto `psy6` the commit applied with no conflicts; all twelve wired field accessors and the `_sys_with_thermal` helper were confirmed present on `psy6`, but **the suite has not been re-run there** — the local InfrastructureSystems checkout is on `lk/enumx` while `psy6` pins IS to `IS4`, so a local run would mean switching branches in another working tree. CI on this PR is the real check.

Separately, on `lk/enumx` the `[sources]` git-`rev` pin of InfrastructureSystems breaks `Pkg.dependencies()`, which `IS/src/serialization.jl:352` calls, so every `to_json` on a System throws `could not find source path for package InfrastructureSystems`. `psy6` pins IS by `rev` too (`IS4`), so it may well be reachable here as well — untested, and unrelated to this change either way.

## Not included

Cost **scalars** (`fixed`, `start_up`, `shut_down`, and the shortage/surplus costs — ~20 `Float64` fields across 10 cost structs) are untouched. None is per-unitized, so their accessors would do pure `uconvert` plus a dimension check on write; assigning them quantity kinds should come from the SiennaSchemas vocabulary rather than being guessed.


luke-kiernan:
> Getters that return units-decorated values: could be added, but we'd be limited to natural units. The relative unit stuff isn't flexible enough to express $/(hr * CU) yet.

This will need some revisions once we decide on how we want to register SU/CU with Unitful.jl (#1817 or #1818). That'll also unblock the units-decorated getters, too.

PR #1804, opened 2026-09-18
Follow InfrastructureSystems' move to EnumX

luke-kiernan:
Companion to **InfrastructureSystems [#643](https://github.com/Sienna-Platform/InfrastructureSystems.jl/pull/643)**, which replaces the hand-rolled `@scoped_enum` with `EnumX`. Merge #643 first — this branch pins `[sources]` at IS `lk/enumx` and that pin reverts to `IS4` once it lands.

`IS.@scoped_enum` now wraps `EnumX.@enumx`, so `Name` is a module and the enum type is `Name.Value`. Definition sites and every `Name.MEMBER` value use are unchanged; what moves is the type positions.

## The original change

- `data_type` in the struct descriptor gains `.Value`, and the generated structs are regenerated from it. No hand edits under `src/models/generated/`.
- The generator accepts `Name.Value` as an enum type name and drops the suffix when deriving the OpenAPI wrapper's name — the wrapper is named for the enum, not its type.
- `ENUMS` in `enums_conversion.jl`, the `convert`/constructor call sites in the OpenAPI import path, and the `Bool`-rejecting guards in `definitions.jl` take `.Value`.
- The guards no longer reach into `OperationalStatesModule` / `CommitmentModesModule`, which EnumX does not create.
- An enum value has no `.value` field, so the wire-format tests read the integer through `Integer(x)`, and a missing member on a module is an `UndefVarError` rather than an `ErrorException`.

## Brought up to psy6

Merged psy6 in (34 commits). Both conflicts were the same one — the `[sources]` IS pin — resolved in favour of `lk/enumx` with a note in both Project.tomls that it reverts to `IS4`.

Running the merged tree turned up three conversions the original sweep missed. All three predate the merge:

| miss | symptom |
|---|---|
| `states_types`' `internal_default` still read `Vector{StateTypes}()` (the sweep updated `data_type` on the adjacent descriptor line) | `AVRFixed(; Vf = 1.0)` threw `TypeError: in GenericMemory, in element type, expected Type, got a value of type Module` |
| `_curve_style_to_wire` / `_curve_multistep_to_wire` took `.value` off an enum with no fields — annotations had been converted, bodies had not | `FieldError: CurveStyles.Value has no field value` on every MarketBidCost export |
| four `get_enum_value(ACBusTypes, …)` call sites passed the module while `ENUM_MAPPINGS` is keyed by `.Value` | `KeyError: key PowerSystems.ACBusTypes not found` from every string-valued `bustype` constructor |

The descriptor fix was made in the descriptor and regenerated — only `AVRFixed.jl` changes, and `test_generated_structs` agrees.

## Testing

**22 of 23 PSCB-independent test files pass: 2903 assertions, 0 failures, 0 errors.**

The full suite is still blocked downstream, and not by anything in this PR: `PowerTableDataParser/src/enums.jl:29` runs `for value in instances(enum)` in a top-level loop over an `ENUMS` tuple that includes `IS.UnitSystem`. Under EnumX that is a module, `instances(::Module)` has no method, and PTDP therefore fails to *precompile* — taking PowerSystemCaseBuilder with it. PTDP needs the same `.Value` sweep in its own repo. That also blocks the one test file left out above, `test_openapi_ptdp_end_to_end.jl`.

Expect IS #643's own `Julia v1 - PowerSystems` check to stay red regardless: that job clones PowerSystems' *default* branch (`main`, which is 370 commits from psy6) with no `ref:`, so it exercises the pre-EnumX tree. It is `continue-on-error: true` and non-blocking.


jd-lara (approved):
Revert the pin to merge once the tests pass please

jd-lara:
@luke-kiernan the tests are still failing

luke-kiernan:
> the tests are still failing

Unrelated: also fails on `psy6`, before the `EnumX` work. Per Claude:

> The actual mechanism (independent of enums): `_load_time_series_associations!` only validates when `_catalog_is_authoritative(store)`. The test writes with the directory document form, which ships no catalog. No catalog, no validation, no throw due to discrepancies -> test fails.
> 
> The test asserts adopt-and-validate behavior that only the `.sns` form (or a sidecar carrying its own catalog) produces. It needs either a `.sns` round-trip or a different expectation.

